PR #1, opened 2017-03-31
MMT-855: Adds ability to choose 'No Access to Collections' for Collection Permissions

charleshuang80:
Adds ability to choose 'No Access to Collections' for Collection Permissions


spec/features/permissions/viewing_permissions_spec.rb
```diff
@@ -210,8 +231,7 @@
     # PUMP can also add group permissions where the group has empty permissions so we should test that as well
     context 'when the collection permission has create, update, and delete permissions as well as groups with empty permissions' do
       let(:perm_concept) { 'ACL3333-CMR' }
```
abbottry: Minor and picky but this abbreviation is misleading, suggest using full word.

app/controllers/permissions_controller.rb
```diff
@@ -339,21 +342,23 @@ def construct_request_object(provider)
...
-      elsif val == 'true'
-        @collection_access_value[key] = true
-      else
-        @collection_access_value[key] = val.to_f
+    if collection_applicable
+      @collection_access_value = params[:collection_access_value] || {}
+      @collection_access_value.each do |key, val|
+        if val.blank?
+          @collection_access_value.delete(key)
+        elsif val == 'true'
```
abbottry: Is this always a string or should we be wrapping this in a boolean check?

charleshuang80: I am running into this issue: http://stackoverflow.com/questions/1711547/how-to-check-if-a-param-is-true-or-false. Is there a better way to handle it?

PR #5, opened 2017-04-04
MMT-900: Enables CMR Management in production

abbottry:


app/controllers/application_controller.rb
```diff
@@ -7,6 +7,7 @@ class ApplicationController < ActionController::Base
   before_action :is_logged_in
   before_action :setup_query
   before_action :refresh_urs_if_needed, except: [:logout, :refresh_token]
+  before_filter :groups_enabled?
```
charleshuang80: Won't this redirect all the collections and drafts as well? Maybe this is better in the ManageCmrController? (also, if you move it, you may want to change what the ProviderHoldingsController inherits from)

abbottry: I think you're right. I'll move the before_filter to the ManageCmrController but I need to leave the method so that it can be used in the navigation. 

Why do we want to change the provider holdings parent? My guess is because users should be able to access it, but ultimately they can't get to the tab at all in the navigation so they would need a direct link to the provider holdings, so I would just leave it. But ultimately this PR enables this link for all environments so it COULD be deleted entirely.

charleshuang80: I was mistaken about the provider holdings parent - I thought that the logged out state also used the ProviderHoldingsController, but actually it is using WelcomeController.

PR #6, opened 2017-04-04
MMT-637: Prevent writing the return_to path to the session for ajax endpoints

abbottry:
On certain endpoints the paths were pretty long, this was being written to the session for the purposes of redirecting after login, but a user will never visit these pages which means we can skip the filter that sets it. The filter also ensures that the user is logged in but the ajax request wont fire because the page itself WILL run through that before filter.

A user could hit the endpoint directly but will get nothing back because it requires a token that is retrieved and stored for logged in user, so there is no risk here.

### DIFF
--- a/app/controllers/manage_cmr_controller.rb
+++ b/app/controllers/manage_cmr_controller.rb
@@ -7,6 +7,13 @@ class ManageCmrController < PagesController
   before_filter :check_if_system_acl_administrator, only: :show
   before_filter :check_if_current_provider_acl_administrator, only: :show
 
+  # These are json respones for ajax calls that user wouldnt get to without being logged in.
+  skip_before_filter :is_logged_in, only: [
+    :provider_collections,
+    :service_implementations_with_datasets,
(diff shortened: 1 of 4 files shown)


app/controllers/manage_cmr_controller.rb
```diff
@@ -7,6 +7,13 @@ class ManageCmrController < PagesController
   before_filter :check_if_system_acl_administrator, only: :show
   before_filter :check_if_current_provider_acl_administrator, only: :show
 
+  # These are json respones for ajax calls that user wouldnt get to without being logged in.
+  skip_before_filter :is_logged_in, only: [
```
charleshuang80: should you add this for the `get_all_collections` endpoint in PermissionsController also?

abbottry: The PermissionsController needs to be updated to use these endpoints as soon as CMR supports concept-id on the permissions endpoints. As it stands this call does something entirely different and no issues have been reported.

PR #10, opened 2017-04-08
MMT-909: Support guest orders

abbottry:


app/views/orders/show.html.erb
```diff
@@ -6,64 +6,64 @@
       <tbody>
         <tr>
           <td>Order Guid:</td>
-          <td><%= @order['guid'] %></td>
+          <td><%= params['id'] %></td>
         </tr>
         <tr>
           <td>State:</td>
-          <td><%= @order['state'] %></td>
+          <td><%= @order['State'] %></td>
```
macrouch: all these capital letters look weird in a view

abbottry: I don't disagree -- but its so much extra work to convert the case and creates more room for error when doing so, also creates a lot more code. At the end of the day its what the endpoint supplies us..

PR #11, opened 2017-04-10
MMT 890: UMM-C from 1.8 to 1.9 (except RelatedUrl changes)

macrouch:
MMT-890 update UMM-C from 1.8 to 1.9 (except RelatedUrl changes)


macrouch:
@abbottry @charleshuang80 the change log of 'Added GetServiceType, GetDataType, and GetDataTypeFormatEnum' will be part of MMT-891, it isn't needed in this ticket

macrouch:
note to @macrouch to check the preview gem for any changes

app/assets/javascripts/drafts.coffee
```diff
@@ -76,6 +76,11 @@ $(document).ready ->
     e.stopImmediatePropagation()
 
   incrementElementIndex = (newDiv, multipleIndex, simple, type) ->
```
charleshuang80: I think there are some elements in the Platforms fields (on the Acquisition Information form) for which the labels are incrementing correctly but the divs for the fields and subfields are not.

For instance, in a new draft on the Acquisition Information form, click on 'Add another Platform' and then for that platform click on 'Add another Instrument'. In the html of the page, you can see a div with `id="draft_platforms_1_instruments"`, and there are divs nested under that, with ids `draft_platforms_1_instruments_0` and `draft_platforms_1_instruments_1`. If we look at the div with `id="draft_platforms_1_instruments_1"`, most of the fields nested there seem to have incremented correctly. But there are two that were not:

1. For Characteristics, there is a label `for="draft_platforms_1_instruments_1_characteristics"`, but the div has `id="draft_platforms_1_instruments_0_characteristics"`. The first child of that div has `id="draft_platforms_1_instruments_0_characteristics_0"`
2. For ComposedOf, the label is `for="draft_platforms_1_instruments_1_composed_of"`, but the div has `id="draft_platforms_1_instruments_0_composed_of"` and the first child also has `id="draft_platforms_1_instruments_0_composed_of_0"`

macrouch: The good news is, it is also doing it in master, so I didn't break it! At least not in the branch haha

I'm looking into it

macrouch: This should be fixed, please test it again

charleshuang80: It looks like point 2 above is incrementing properly now, but I think point 1 ('instrument_n_characteristics') is still not incrementing correctly:
screen shot 2017-04-12 at 4 56 18 pm
and
screen shot 2017-04-12 at 4 59 43 pm

charleshuang80: This was my mistake - I hadn't pulled down the latest changes. doh.

PR #14, opened 2017-04-11
MMT-867: Remove provider holdings from the homepage

abbottry:


charleshuang80:
I'm ok with this, but since there are commented out tests/routes, I'm wondering if there's a ticket to rework it when the CMR changes come through, and if we should add that ticket number to comments indicating that they'll be updated with the ticket.

abbottry:
@charleshuang80 no because there is no plan yet as to what we are going to do with provider holdings going forward.

macrouch:
I think adding a comment in some of the files for why we are just commenting out a lot of stuff might be appreciated by our future selves

PR #15, opened 2017-04-11
MMT-860: Compile and compress CoffeeScript

abbottry:


package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "mmt",
+  "version": "1.0.0",
```
charleshuang80: Is this just for our benefit, or do we need to keep this updated somehow (i.e. the version in Bamboo)?

abbottry: https://docs.npmjs.com/files/package.json#version

PR #17, opened 2017-04-14
MMT-731: for collection permissions, prevent selecting same group for…

charleshuang80:
… search AND search & order by disabling selected group in other select input. assume 'search' permissions for groups that only have 'order' permissions. add/edit tests


charleshuang80:
Are you ok with adding the coffeelint.json file to .gitignore until we decide something on a coffeescript linter?

abbottry:
Sure, you can add anything you want to the .gitignore :)

PR #21, opened 2017-04-17
MMT-922: Updates all calls to collection search to use POST

abbottry:


app/controllers/order_option_assignments_controller.rb
```diff
@@ -43,24 +44,19 @@ def edit
       end
 
       empty_assignment_cnt = 0
-      @collections_to_list.each do |collection|
-        if collection['option-def'].nil?
-          empty_assignment_cnt += 1
-        end
+      @collections_to_list.each do |coll|
```
charleshuang80: minor, but I thought you wanted to avoid abbreviated variable/parameter names that can be ambiguous.

abbottry: You're correct, I made this change because 'collection' was clashing with the variable just outside the scope of this loop.

PR #23, opened 2017-04-17
MMT-745: remove initial management group from groups form and control…

charleshuang80:
…ler methods. remove 'Group Management for...' in provider and system identity permissions tables. remove tests, coffeescript, helper, etc references to management groups or single instance acls

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -2,8 +2,6 @@ class GroupsController < ManageCmrController
   include GroupsHelper
   include PermissionManagement
 
-  before_filter :set_system_and_provider_groups, except: [:index, :show, :destroy]
-
   add_breadcrumb 'Groups', :groups_path
 
   RESULTS_PER_PAGE = 25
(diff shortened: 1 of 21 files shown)


app/controllers/groups_controller.rb
```diff
@@ -68,53 +63,13 @@ def new
...
-          Rails.logger.error("Group Deletion Error: #{delete_group_response.inspect}")
-          flash[:error] = "Group [#{@group['name']}] was created but there were issues with the Initial Management Group permissions. Please delete this group and try again."
-
-          redirect_to group_path(concept_id) and return
-        end
-      end
-      flash[:success] = 'Group was successfully created.'
-
-      redirect_to group_path(concept_id)
+      redirect_to group_path(group_creation_response.body.fetch('concept_id', nil)), flash: { success: 'Group was successfully created.' }
```
abbottry: `group_creation_response.body.fetch('concept_id', nil)` is the same thing that you had before `group_creation_response.body['concept_id']`

app/controllers/system_identity_permissions_controller.rb
```diff
@@ -60,12 +57,9 @@ def edit
...
     @group_id = params[:id]
     permissions_params = params[:system_permissions]
-    group_management_params = params[:group_management]
-
-    redirect_to system_identity_permissions_path and return if permissions_params.nil? && group_management_params.nil?
+    redirect_to system_identity_permissions_path and return if permissions_params.nil?
 
-    permissions_params ||= {} # permissions_params might be nil even if group_management_params are not
-    permissions_params.each { |_target, perms| perms.delete('') } unless permissions_params.nil?
+    permissions_params.each { |_target, perms| perms.delete('') }
```
abbottry: An alternative to this would be to use `.reject`

PR #25, opened 2017-04-19
MMT-921: change rake task to output only umm version number to .umm-version file

charleshuang80:


charleshuang80:
Pending any comments that Leo has on the PR in the gem repo.
Also, as before, will want to wait until the gem PR is merged, then bundle to get the updated version number/hash.

PR #29, opened 2017-04-25
MMT-874: Implements ability to search for collections by a single field

abbottry:


app/views/bulk_updates_searches/new.html.erb
```diff
@@ -0,0 +1,65 @@
...
+            <th>Collection</th>
+            <th>Short Name</th>
+            <th>Version</th>
+          </tr>
+        </thead>
+        <tbody>
+          <% if @collections.any? %>
+          <% @collections.each do |collection| %>
+          <tr>
+            <td><%= link_to collection.fetch('umm', {}).fetch('entry-title'), collection_path(collection.fetch('meta', {}).fetch('concept-id')), target: '_blank' %></td>
```
charleshuang80: Could these second `fetch`es throw a nil error?

app/views/pages/manage_metadata.html.erb
```diff
@@ -1,182 +0,0 @@
...
-    <section class="eui-callout-box col-left">
-      <h3 class="eui-callout-box__title">Upload XML Files</h3>
-      <div class="eui-callout-box__list">
-        <a class="eui-btn eui-btn--sm disabled" href="javascript:void(0);">Select files</a><span class="coming-soon">Coming Soon</span>
-      </div>
-    </section>
-
-    <!-- <section class="callout-box col-left">
-      <div class="callout-content">
-        <h3>Collection Templates</h3>
```
charleshuang80: We may be working on Templates soon-ish right? Would it be helpful to keep this section?

spec/features/bulk_updates/bulk_updates_search_spec.rb
```diff
@@ -0,0 +1,77 @@
+require 'rails_helper'
+
+describe 'Searching collections to bulk update', reset_provider: true do
+  before(:all) do
+    # Create a few collections with unique attributes that we can search for
+    2.times { |i| publish_draft(version: "nasa.001#{i}") }
+    3.times { |i| publish_draft(short_name: "nasa.002#{i}") }
```
charleshuang80: I think setting up the test this way is fine, but an alternative could have been to switch providers to LARC or SEDAC and running a search against those collections, since they have been ingested.

app/views/pages/manage_metadata.html.erb
abbottry: No, commented code isn't useful and this is nothing but a placeholder.

spec/features/bulk_updates/bulk_updates_search_spec.rb
abbottry: Noted

app/views/bulk_updates_searches/new.html.erb
abbottry: COULD, it'd be very very unlikely but I'll update, its a good practice to ensure you catch even unlikely scenarios

spec/features/bulk_updates/bulk_updates_search_spec.rb
charleshuang80: I just meant that you could have used some searches similar to what is done in `search_collections_result_spec`. But I think the methods that we have to publish drafts and stipulate certain attributes is much improved since then.

PR #30, opened 2017-04-26
MMT-886: add ability to query endpoints and view bulk updates index and show pages

charleshuang80:
with current dummy data.

### DIFF
--- a/app/controllers/bulk_updates_controller.rb
+++ b/app/controllers/bulk_updates_controller.rb
@@ -5,20 +5,9 @@ class BulkUpdatesController < ManageMetadataController
   add_breadcrumb 'Bulk Updates', :bulk_updates_path
 
   def index
-    @provider_id = current_user.provider_id
-
     @tasks = []
 
-    # We cannot implement pagination yet, because the cmr dummy responses do not include it
-    # when the response has `hits` and accepts page information, we should implement it in accordance with other resources (i.e. groups)
-    filters = {}
-
-    # filters[:page_size] = RESULTS_PER_PAGE
-    # default page to 1
-    # page = params.fetch('page', 1)
-    # @filters[:page_num] = page.to_i
-
-    bulk_updates_list_response = cmr_client.get_bulk_updates(@provider_id, filters, token)
+    bulk_updates_list_response = cmr_client.get_bulk_updates(current_user.provider_id, token)
 
     if bulk_updates_list_response.success?
       @tasks = bulk_updates_list_response.body.fetch('tasks', [])
@@ -29,11 +18,9 @@ class BulkUpdatesController < ManageMetadataController
 
   def show
     @task_id = params[:id]
-    @provider_id = current_user.provider_id
-
     @task = {}
 
-    bulk_update_status_response = cmr_client.get_bulk_update(@provider_id, @task_id, token)
+    bulk_update_status_response = cmr_client.get_bulk_update(current_user.provider_id, @task_id, token)
     if bulk_update_status_response.success?
       @task = bulk_update_status_response.body
 
--- a/app/controllers/bulk_updates_controller.rb
+++ b/app/controllers/bulk_updates_controller.rb
@@ -1,5 +1,6 @@
 # :nodoc:
 class BulkUpdatesController < ManageMetadataController
+  before_filter :bulk_updates_enabled?
 
   add_breadcrumb 'Bulk Updates', :bulk_updates_path
 
@@ -17,13 +18,12 @@ class BulkUpdatesController < ManageMetadataController
     # page = params.fetch('page', 1)
     # @filters[:page_num] = page.to_i
 
-    bulk_updates_list_response = cmr_client.get_bulk_updates_list(@provider_id, filters, token)
+    bulk_updates_list_response = cmr_client.get_bulk_updates(@provider_id, filters, token)
 
     if bulk_updates_list_response.success?
       @tasks = bulk_updates_list_response.body.fetch('tasks', [])
     else
       Rails.logger.error("Error retrieving Bulk Updates Jobs List: #{bulk_updates_list_response.inspect}")
-      flash[:error] = Array.wrap(bulk_updates_list_response.body['errors'])[0]
     end
   end
 
@@ -33,14 +33,13 @@ class BulkUpdatesController < ManageMetadataController
 
     @task = {}
 
-    bulk_update_status_response = cmr_client.get_bulk_update_status(@provider_id, @task_id, token)
+    bulk_update_status_response = cmr_client.get_bulk_update(@provider_id, @task_id, token)
     if bulk_update_status_response.success?
       @task = bulk_update_status_response.body
 
       add_breadcrumb @task_id, bulk_update_path(@task_id)
     else
       Rails.logger.error("Error retrieving Bulk Update Task: #{bulk_update_status_response.inspect}")
-      flash[:error] = Array.wrap(bulk_update_status_response.body['errors'])[0]
     end
   end
 end
--- a/app/controllers/bulk_updates_controller.rb
+++ b/app/controllers/bulk_updates_controller.rb
@@ -1,0 +1,49 @@
+# nodoc
+class BulkUpdatesController < ApplicationController
+
+  add_breadcrumb 'Bulk Updates', :bulk_updates_path
+
+  # RESULTS_PER_PAGE = 25
+
+  def index
+    @provider_id = current_user.provider_id
+
+    @tasks = []
+    # pagination? Response does not have `hits`
+    # @filters = { page_size: RESULTS_PER_PAGE }
+    # # default page to 1
+    # page = params.fetch('page', 1)
+    # @filters[:page_num] = page.to_i
+
+    bulk_updates_jobs_response = cmr_client.get_bulk_updates_tasks_list(@provider_id, token) # TODO add filters for pagination?
+
+    if bulk_updates_jobs_response.success?
+      @tasks = bulk_updates_jobs_response.body.fetch('tasks', [])
+      @http_status = bulk_updates_jobs_response.body.fetch('status', nil)
+    else
+      Rails.logger.error("Error retrieving Bulk Updates Jobs List: #{bulk_updates_jobs_response.inspect}")
+      flash[:error] = Array.wrap(bulk_updates_jobs_response.body['errors'])[0]
+    end
+  end
+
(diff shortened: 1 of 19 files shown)


app/controllers/bulk_updates_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    # page = params.fetch('page', 1)
+    # @filters[:page_num] = page.to_i
+
+    bulk_updates_list_response = cmr_client.get_bulk_updates_list(@provider_id, filters, token)
+
+    if bulk_updates_list_response.success?
+      @tasks = bulk_updates_list_response.body.fetch('tasks', [])
+    else
+      Rails.logger.error("Error retrieving Bulk Updates Jobs List: #{bulk_updates_list_response.inspect}")
+      flash[:error] = Array.wrap(bulk_updates_list_response.body['errors'])[0]
```
abbottry: No flash errors on pages that users have not taken an action on. These should only be used when the user has submitted a form or some other action where they COULD resolve the issue by modifying the data they are submitting

app/controllers/bulk_updates_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    @task = {}
+
+    bulk_update_status_response = cmr_client.get_bulk_update_status(@provider_id, @task_id, token)
+    if bulk_update_status_response.success?
+      @task = bulk_update_status_response.body
+
+      add_breadcrumb @task_id, bulk_update_path(@task_id)
+    else
+      Rails.logger.error("Error retrieving Bulk Update Task: #{bulk_update_status_response.inspect}")
+      flash[:error] = Array.wrap(bulk_update_status_response.body['errors'])[0]
```
abbottry: Again, no flash errors here.

app/views/bulk_updates/index.html.erb
```diff
@@ -0,0 +1,28 @@
+<div class="row row-content">
+  <section>
+    <h2>Bulk Updates for <%= @provider_id %></h2>
```
abbottry: Our headers are of the format `<%= @provider_id %> Bulk Updates`

app/views/bulk_updates/index.html.erb
```diff
@@ -0,0 +1,28 @@
...
+      </thead>
+      <tbody>
+        <% if @tasks.blank? %>
+          <tr>
+            <td colspan="2">No tasks found</td>
+          </tr>
+        <% else %>
+          <% @tasks.each do |task| %>
+            <tr>
+              <td><%= link_to task.fetch('task-id', nil), bulk_update_path(task.fetch('task-id', 'not found')) %></td>
```
abbottry: Prefer `task['task-id']` over using fetch.

app/views/bulk_updates/index.html.erb
```diff
@@ -0,0 +1,28 @@
...
+              <td><%= task.fetch('status', nil) %></td>
```
abbottry: Prefer `task['status]` over using fetch.

app/views/bulk_updates/show.html.erb
```diff
@@ -0,0 +1,38 @@
+<div class="row row-content">
+  <section>
+    <h2>Bulk Update <%= @task_id %></h2>
+
+    <h3>Task Status: <%= @task.fetch('task-status', 'Unknown') %></h3>
+    <h4>Status Message: <%= @task.fetch('status_message', 'Unknown') %></h4>
+
+  </section>
+  <section>
+    <% if @task.fetch('collection-statuses', nil).blank? %>
```
abbottry: No need to display this, if there are no errors, just don't show this.

app/views/bulk_updates/show.html.erb
```diff
@@ -0,0 +1,38 @@
...
+      <h4>No Bulk Update Errors for this Task</h4>
+      <!-- TODO put this in table or keep out of table? -->
+    <% else %>
+
+      <h4>Bulk Update Errors</h4>
+      <table class="space-top space-bot" id="bulk-update-errors-table">
+        <thead>
+          <tr>
+            <th>Collection Concept ID</th> <!-- TODO should we be requesting and displaying other info? i.e. entry title, entry id... -->
```
abbottry: I would suggest matching the other tables across the app in display the `entry title`, `short name`, and `version`.

app/views/bulk_updates/show.html.erb
```diff
@@ -0,0 +1,38 @@
...
+            <th>Status</th>
+            <th>Status Message</th>
+          </tr>
+        </thead>
+
+        <tbody>
+          <% @task['collection-statuses'].each do |task_error| %>
+            <tr>
+              <td><%= link_to @task_id, collection_path(@task_id) %></td>
+              <td><%= task_error.fetch('status', 'Unknown') %></td>
```
abbottry: Remove this guy per our conversation (and the endpoint update)

app/views/bulk_updates/show.html.erb
```diff
@@ -0,0 +1,38 @@
+<div class="row row-content">
+  <section>
+    <h2>Bulk Update <%= @task_id %></h2>
+
+    <h3>Task Status: <%= @task.fetch('task-status', 'Unknown') %></h3>
```
abbottry: Remove this guy per our conversation (and the endpoint update)

lib/cmr/cmr_client.rb
```diff
@@ -136,30 +167,36 @@ def delete_collection(provider_id, native_id, token)
       delete(url, {}, nil, headers.merge(token_header(token)))
     end
 
-    # This method will be replaced by the work from CMR-2053, including granule counts in umm-json searches
-    def get_granule_count(collection_id, token)
+    ### CMR Bulk Updates, via CMR Ingest
+
+    def get_bulk_updates_list(provider_id, filters, token)
```
abbottry: Suggest changing to `get_bulk_updates`

spec/controllers/bulk_updates_controller_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'rails_helper'
+
+describe BulkUpdatesController, reset_provider: true do
+  # TODO add contexts 'when bulk updates enabled' and 'when bulk updates disabled' ?
```
abbottry: Yes, please. Make sure that users cannot access this controller if bulk edits are not enabled. Can easily be copied from bulk updates search controller spec.

app/views/bulk_updates/index.html.erb
charleshuang80: For both? For the second one, in the unlikely case that there was no `task-id` I was trying to avoid the No Route error if the value passed to the route helper was `nil`

abbottry: You could use `link_to_if` here, if there is no task-id then you're currently displaying a link with no text and supplying 'not found' as the ID. I'd prefer this not be a link at all, *or better yet, if no task-id exists, this row shouldn't be displayed because its a useless piece of data*

app/views/bulk_updates/show.html.erb
```diff
@@ -0,0 +1,41 @@
...
+          <tr>
+            <th>Entry Title</th>
+            <th>Short Name</th>
+            <th>Version</th>
+            <th>Status Message</th>
+          </tr>
+        </thead>
+
+        <tbody>
+          <% @task['collection-statuses'].each do |task_error| %>
```
abbottry: In most spots of this PR you've got a bit far on ensuring we have values to display but on this one you're at risk of throwing a `nil has no method each` due to a lack of default on the 'collection-statuses' key 😃 

When iterating over an array, we should ensure we actually have an array by using `fetch` (most commonly used throughout the app) or even `||`

app/controllers/bulk_updates_controller.rb
```diff
@@ -0,0 +1,45 @@
+# :nodoc:
+class BulkUpdatesController < ManageMetadataController
+  before_filter :bulk_updates_enabled?
+
+  add_breadcrumb 'Bulk Updates', :bulk_updates_path
+
+  def index
+    @provider_id = current_user.provider_id
```
abbottry: Unnecessary instance variable (see full comment below)

app/views/bulk_updates/index.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      </thead>
+      <tbody>
+        <% if @tasks.blank? %>
+          <tr>
+            <td colspan="2">No tasks found</td>
+          </tr>
+        <% else %>
+          <% @tasks.each do |task| %>
+            <tr>
+              <% next unless task['task-id'] %>
```
charleshuang80: No, I'm not that concerned with it. I would have taken it out but in your PR I made a comment and you said 'its a good practice to ensure you catch even unlikely scenarios'. I will leave it out, I just sometimes go back and forth figuring out where that line should be.

app/views/bulk_updates/show.html.erb
```diff
@@ -0,0 +1,41 @@
...
+    <h2>
+      Bulk Update <%= @task_id %>
+      <span class="bulk-update-status eui-badge--sm"><%= @task.fetch('task-status') %></span>
+    </h2>
+    <p class="bulk-update-status-message"><%= @task.fetch('status-message', 'Status not retrieved') %></p>
+
+  </section>
+
+  <section>
+    <% unless @task.fetch('collection-statuses', nil).blank? %>
```
charleshuang80: Thanks. I had it like that before, I think I changed it b/c nil seems a more explicit check but what you're saying here makes more sense.

app/views/bulk_updates/show.html.erb
charleshuang80: haha :). I think it seemed ok because of the check above to get into the table, but yes there could still be some issue with trying to iterate over the response.

app/views/bulk_updates/show.html.erb
```diff
@@ -0,0 +1,42 @@
+<div class="row row-content">
+  <section>
+    <h2>Bulk Update <%= @task_id %></h2>
+
+    <div class="row">
+      <label class="bulk-update-status"><%= @task.fetch('task-status') %></label>
```
abbottry: KeyNotFound Error. Since we're here and need to make an update, both of these should probably just use `['']` given that they are defaulting to nil. `status-message` is also used below in this file, and could/should also use the `['']` because it also defaults to nil (if the key doesnt exist when using `['']` it just returns nil, so its the same result with less work)

PR #35, opened 2017-05-08
MMT-930: Updates bulk update tests

abbottry:
Partial implementation of MMT-930 for the purposes of getting test
suite to pass for all other functionality.


spec/features/bulk_updates/bulk_updates_list_spec.rb
```diff
@@ -5,24 +5,17 @@
...
 
-  context 'when there are bulk updates' do
+  context 'when there are no bulk updates' do
     before do
-      # when the endpoint accepts and returns real data we will need to create some bulk updates
-      
       visit bulk_updates_path
     end
 
     it 'displays the page with a list of bulk updates' do
```
charleshuang80: maybe the `it` statement should be changed to reflect the message?

spec/features/manage_metadata/bulk_updates_list_spec.rb
```diff
@@ -9,33 +9,9 @@
     visit manage_metadata_path
   end
 
-  context 'when there are bulk updates' do
+  context 'when there are no bulk updates' do
     it 'displays the bulk updates' do
```
charleshuang80: this `it` statement also should be updated

PR #36, opened 2017-05-11
MMT-926: prevent destructive inadvertent removal of collections from …

charleshuang80:
…collection permissions upon update, when user does not have access to see the collection.

- keep collections that are not retrieved by search, and make sure they are passed to update call
- add notification when there are collections not shown for collection permissions
- add collections to ingest for provider to use in tests
- add ACLs in cmr setup to create scenario to test when user may not have access to all collections in a collection permission
- add tests

### DIFF
--- a/app/controllers/permissions_controller.rb
+++ b/app/controllers/permissions_controller.rb
@@ -171,6 +171,7 @@ class PermissionsController < ManageCmrController
 
         @collection_options = params[:collection_options]
         @collection_selections = params[:collection_selections]
+        @hidden_entry_titles = params[:hidden_collections]
         @granule_options = params[:granule_options]
 
         @collection_access_value = params[:collection_access_value] || {}
@@ -354,9 +355,7 @@ class PermissionsController < ManageCmrController
       end
 
       collection_identifier['entry_titles'] = entry_titles
-      if params[:hidden_collections]
-        collection_identifier['entry_titles'] += params[:hidden_collections].split(';;; ')
-      end
+      collection_identifier['entry_titles'] += params.fetch('hidden_collections', [])
     end
 
     if collection_applicable
--- a/app/controllers/permissions_controller.rb
+++ b/app/controllers/permissions_controller.rb
@@ -88,7 +88,6 @@ class PermissionsController < ManageCmrController
 
   def create
     request_object = construct_request_object(current_user.provider_id)
-
     response = cmr_client.add_group_permissions(request_object, token)
 
     if response.success?
@@ -257,12 +256,6 @@ class PermissionsController < ManageCmrController
 
     collections_response = cmr_client.get_collections_by_post(query, token)
     parse_get_collections_response(collections_response.body)
-
-    # we previously used the GET method for searching collections by entry title,
-    # but that may have been too large for Jetty. we are using the POST method
-    # until/unless CMR changes collection permissions to utilize concept ids or entry ids
-    # collections_response = cmr_client.get_collections(query, token).body
-    # parse_get_collections_response(collections_response)
   end
 
   def parse_get_collections_response(response)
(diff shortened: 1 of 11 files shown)


app/views/permissions/_form.html.erb
```diff
@@ -11,7 +11,11 @@
   <%= label_tag 'collection_options', 'Collections', class: 'required eui-required-o' %>
   <%= select_tag 'collection_options', options_for_select(PermissionsHelper::CollectionsOptions, @collection_options), class: 'full-width', prompt: '-- Select an Option --' %>
   <%= hidden_field_tag 'collection_selections', @collection_selections %>
-  <div id="chooser-widget" class="is_hidden"></div>
+  <div id="chooser-widget" class="is-hidden"></div>
+
+  <% unless @hidden_entry_titles.blank? %>
+    <%= hidden_field_tag('hidden_collections', @hidden_entry_titles.join(';;; ')) %>
```
abbottry: Let's talk about a different way of doing this, delimiters like this are bad news bears.

app/controllers/permissions_controller.rb
```diff
@@ -361,6 +354,9 @@ def construct_request_object(provider)
       end
 
       collection_identifier['entry_titles'] = entry_titles
+      if params[:hidden_collections]
+        collection_identifier['entry_titles'] += params[:hidden_collections].split(';;; ')
```
abbottry: Let's talk about a different way of doing this, delimiters like this are bad news bears.

app/views/permissions/_hidden_collections_notification.html.erb
```diff
@@ -0,0 +1,7 @@
+<% unless @hidden_entry_titles.blank? %>
```
abbottry: We'll need to run this through @mreese84 and/or @jsiarto -- I don't think we want this type of messaging.

app/views/permissions/show.html.erb
```diff
@@ -8,7 +10,7 @@
     <p>
       <b>Collections |
         <% if @collection_options %>
-          <%= @collection_entry_ids.blank? ? 'All Collections' : "#{@collection_entry_ids.count} Selected Collections" %>
+          <%= @collection_entry_ids.blank? && @hidden_entry_titles.blank? ? 'All Collections' : "#{@collection_entry_ids.count + @hidden_entry_titles.count} Selected Collections" %>
```
abbottry: If I understand this correctly, we're increasing the collection count, and displaying the message to the user, but not listing the collections anywhere, is that right?

app/views/permissions/_form.html.erb
charleshuang80: picked this of the top of my head, and totally open to changing it. the array does become a string for the hidden field though, so what kinds of delimiters are appropriate?

app/views/permissions/_hidden_collections_notification.html.erb
charleshuang80: Agreed. I had already added this when you said that @mreese84 wouldn't want a message to the user, so I left it in so we could discuss.

app/views/permissions/show.html.erb
charleshuang80: Yes, for 2 reasons:
1. The way the logic was working, if the user couldn't see any of the collections (they turned out all be in the hidden field), the show page would say it was a permission for 'All Collections'. We can obviously change it how this works, but it could be weird to have it say '0 Selected Collections'
2. If we have some sort of notification about there being collections the user can't see, then the discrepancy would make sense to them. They might be able to also infer how many collections they can't see, but if that is something we want them to know we should be more direct about it.

spec/features/permissions/updating_permission_with_restricted_collections_spec.rb
```diff
@@ -0,0 +1,285 @@
...
+        }
+      }
+    }
+
+    @collection_permission_all_restricted = cmr_client.add_group_permissions(collection_permission_all_restricted, 'access_token_admin').body
+
+    wait_for_cmr
+  end
+
+  # need to delete?
```
abbottry: What might we need to delete?

charleshuang80: oops. I'll remove this. it was a placeholder for me to decide if I needed to delete the collection permissions after the tests ran.

PR #38, opened 2017-05-17
MMT-871: integrate Earthdata Status App into MMT

charleshuang80:
- add Earthdata Status banner and JS tag
- remove notifications and calls to ECHO calendar (which will be deprecated)


app/views/layouts/application.html.erb
```diff
@@ -15,14 +15,19 @@
   <%= render partial: 'shared/assets' %>
 </head>
 
-<body class="<%= (controller.action_name == 'manage_metadata' || controller.action_name == 'manage_cmr' ? 'dashboard' : '') %>">
```
macrouch: was this not being used?

charleshuang80: no - we have the Manage Cmr and Manage Metadata Controllers now, with their own layouts, and the controller actions are `show`

macrouch:
Where is the documentation for the status app?

And is there anyway to write a test for it?

charleshuang80:
The documentation is at: https://wiki.earthdata.nasa.gov/display/ET/How+to+integrate+the+Earthdata+Status+App

When I spoke with @abbottry, he thought it would be nice to use the widget since it was maintained by someone else and we shouldn't have to write tests for it. I think we may be able to record one on vcr if that is something we need. What do you think?

PR #42, opened 2017-05-19
MMT-874: update bulk update search by single field

charleshuang80:


app/controllers/bulk_updates_searches_controller.rb
```diff
@@ -18,7 +18,7 @@ def new
     collection_results = []
 
     if search_params.key?(:field) && search_params.key?(:query)
-      cmr_params[search_params[:field]] = search_params[:query].dup
+      cmr_params[search_params[:field].to_sym] = search_params[:query].dup
```
abbottry: I'm interested to know what this is for

spec/features/bulk_updates/bulk_updates_search_spec.rb
```diff
@@ -4,7 +4,9 @@
   before(:all) do
     # Create a few collections with unique attributes that we can search for
     2.times { |i| publish_draft(version: "nasa.001#{i}") }
-    3.times { |i| publish_draft(short_name: "nasa.002#{i}") }
+    3.times { |i| publish_draft(short_name: "nasa.002", version: "#{i}") }
+
+    wait_for_cmr
```
abbottry: What part of the draft publishing is asynchronous, that would require this?

charleshuang80: I added this while I was working on fixing the tests. I think that there is some indexing that happens after publishing (at least in the real CMR) that might be async and sometimes slow. We have a comment about it in line 127 in the `get_revisions` method in `CollectionsController`.

I think the tests work fine (locally) without this though, do you think I should remove it?

app/controllers/bulk_updates_searches_controller.rb
charleshuang80: So because this is using the value of the param, it was getting stored as a string, and when it was trying to match against the `wildcard_keys` in `hydrate_params`, they were not matching. This addition allowed the keys to match. Is there a better way this should be fixed?

app/helpers/bulk_updates_helper.rb
```diff
@@ -2,33 +2,128 @@ module BulkUpdatesHelper
...
+        format: 'date',
+        description: 'Accepted Date Format: yyyy-mm-ddThh:mm:ss.SSSZ'
+      }
+    },
+    'science_keywords': {
+      title: 'Science Keywords',
+      data_attributes: {
+        supports_wildcard: false,
+        format: 'text',
+        description: 'Accepted Values: a value at any level of the Science Keyword hierarchy'
```
abbottry: 'Searches for provided search term at all levels of the science keyword hierarchy.'

app/helpers/bulk_updates_helper.rb
```diff
@@ -2,33 +2,128 @@ module BulkUpdatesHelper
...
+        supports_wildcard: true,
+        format: 'text'
+      }
+    },
+    'revision_date': {
+      title: 'Revision Date',
+      data_attributes: {
+        supports_wildcard: false,
+        format: 'date',
+        description: 'Accepted Date Format: yyyy-mm-ddThh:mm:ss.SSSZ'
```
abbottry: 'Accepts dates in `yyyy-mm-ddThh:mm:ss.SSSZ` format.

app/helpers/bulk_updates_helper.rb
```diff
@@ -2,33 +2,128 @@ module BulkUpdatesHelper
...
+        supports_wildcard: false,
+        format: 'text'
+      }
+    },
+    'updated_since': {
+      title: 'Updated Since',
+      data_attributes: {
+        supports_wildcard: false,
+        format: 'date',
+        description: 'Accepted Date Format: yyyy-mm-ddThh:mm:ss.SSSZ'
```
abbottry: 'Accepts dates in `yyyy-mm-ddThh:mm:ss.SSSZ` format.

app/assets/javascripts/bulk_updates.coffee
```diff
@@ -1,13 +1,39 @@
...
+  if($selectedFieldData.hasOwnProperty('description'))
+    $queryFieldDescription.text($selectedFieldData['description'])
+  else
+    $queryFieldDescription.text('')
+
+  searchFieldElement.next('.form-description').remove()
+
+  if($selectedFieldData.hasOwnProperty('supports_wildcard') && $selectedFieldData['supports_wildcard'] == true)
+    $('<p>').addClass('form-description')
+      .text('This field supports wildcard searching. Add `*` to your search to return more results.')
```
abbottry: Based on a quick search and read of other applications' same feature the following seems to align with the feature. 

'This field supports wildcard searches. Use an asterisk (*) to find collections that match zero or more characters at the beginning, middle or end of your term.'

PR #45, opened 2017-05-22
MMT-884: Adds ability for user to select collections for bulk update

abbottry:


app/views/bulk_updates_searches/new.html.erb
```diff
@@ -31,36 +31,55 @@
...
+            <tr>
+              <th class="align-c"><input type="checkbox" name="checkall" class="checkall" data-group="bulk_update_collections[]" /></th>
+              <th>Collection</th>
+              <th>Short Name</th>
+              <th>Version</th>
+            </tr>
+          </thead>
+          <tbody>
+            <% if @collections.any? %>
+            <% @collections.each do |collection| %>
```
charleshuang80: indentation within the `if` erb tags. or we can just make the edit while doing 863.

charleshuang80:
So there aren't any tests. But the tests would be just looking at the params on the next page, so should we just add the tests into 863?

app/views/bulk_updates_searches/new.html.erb
abbottry: Let's take care of it in 863.

abbottry:
@charleshuang80 yea the tests would ultimately just be checking for checkboxes and/or that sorting is correct. I think it might make more sense to add a lot of tests towards the end of the PI when the functionality is more fully baked and thought through.

PR #51, opened 2017-05-24
MMT-941: Handles ACLs with no permissions at all

abbottry:


app/views/permissions/show.html.erb
```diff
@@ -11,7 +11,7 @@
         </tr>
       </thead>
       <tbody>
-        <% if @permission.fetch('group_permissions', []).blank? %>
+        <% if @permission.fetch('group_permissions', []).blank? || @permission.fetch('group_permissions', []).all? { |group| group.fetch('permissions', []).blank? } %>
```
charleshuang80: Unfortunately, I think you need to handle legacy data cases where the permissions might have `create`, `update`, or `delete` permissions, and still have it display 'No groups listed'.

abbottry: That is already handled, `show` calls `hydrate_groups` which, if the group is valid calls `hydrate_permissions` which deletes any permissions that are not **read** or **order**

PR #57, opened 2017-06-05
MMT-911: Replaces group membership form with autocomplete

abbottry:


app/controllers/groups_controller.rb
```diff
@@ -173,7 +171,24 @@ def invite
...
+    @added = recipient_uid && @invite.accept_invite(cmr_client, recipient_uid, token)
+  end
+
+  def urs_search
+    render json: render_users_from_urs(
+      search_urs(params[:search])
+    )
+  end
+
+  def provided_urs_users
```
charleshuang80: I looked for where this endpoint was being used but didn't find anything. Did I miss it or is this not being used?

lib/cmr/urs_client.rb
```diff
@@ -16,26 +16,33 @@ def get_profile(endpoint, token)
...
+    def search_urs_users(query)
+      client_token = get_client_token
+      Cmr::Response.new(connection.get('/api/users', { search: query }, 'Authorization' => "Bearer #{client_token}"))
     end
 
     protected
 
     def get_client_token
       # URS API says that the client token expires in 3600 (1 hr)
       # so cache token for one hour, and if needed will run request again
```
charleshuang80: If we're not caching anymore, maybe remove the comments that say we are?

spec/factories/user_invites.rb
```diff
@@ -15,9 +15,9 @@
   factory :user_invite_with_group, class: UserInvite do
     manager_name 'Manager Name'
     manager_email 'manager@example.com'
-    user_first_name 'First Name'
-    user_last_name 'Last Name'
-    user_email 'user@example.com'
+    user_first_name 'Execktamwrwcqs'
+    user_last_name '02Wvhznnzjtrunff'
+    user_email 'pvblvweo@hqdybllrn.sghjz'
```
charleshuang80: somewhere there's a bad email validator ;)

app/assets/javascripts/groups.coffee
```diff
@@ -1,4 +1,65 @@
...
+
+setCurrentGroupMembers = ->
+  $.each $('.urs_autocomplete'), ->
+    selectedUsers = $.map $(this).find('option'), (option) -> 
+      $(option).val()
+
+    $(this).val(selectedUsers).trigger('change')
+
 $(document).ready ->
+  $('.urs_autocomplete').select2
```
charleshuang80: Since most of the group members JavaScript is now in this file, `groups_dual_select.coffee` is not being used and can be removed now, right?

app/controllers/groups_controller.rb
abbottry: I like your style.

PR #58, opened 2017-06-05
MMT-865: create preview page displaying all values of a bulk update operation

charleshuang80:


app/controllers/bulk_updates_controller.rb
```diff
@@ -30,5 +30,17 @@ def new
     @science_keywords = cmr_client.get_controlled_keywords('science_keywords')
   end
 
-  def preview; end
+  def preview
+    redirect_to new_bulk_updates_search_path if request.get?
+
+    add_breadcrumb 'Preview', bulk_update_preview_path
+
+    collections_response = cmr_client.get_collections_by_post({ concept_id: params[:concept_ids], page_size: params[:concept_ids].count }, token)
```
abbottry: I think I've discovered that if `params[:concept_ids]` is empty, this will return **ALL** collections. We have this issue across the application, but in this case, its a real possibility because we don't require collections.

app/helpers/bulk_updates_helper.rb
```diff
@@ -172,4 +182,48 @@ def update_type_select
...
+      description = data_attributes[:new_description]
+    when 'find'
+      title = data_attributes[:find_title]
+      description = data_attributes[:find_description]
+    end
+
+    "<h4>#{title}</h4><p>#{description}</p>".html_safe
+  end
+
+  def display_science_keyword(keyword)
```
abbottry: Could you please add some comments to this method, I'm confused by the `reached_value` variable specifically.

spec/controllers/bulk_updates_controller_spec.rb
```diff
@@ -114,7 +114,63 @@
...
+      end
+    end
+  end
+
+  describe 'GET #preview' do
+    context 'when bulk updates are enabled' do
+      before do
+        sign_in
+
+        get :preview, concept_ids: ['1', '2']
```
abbottry: Shouldn't GET requests get sent to `new_bulk_updates_search_path`?

spec/controllers/bulk_updates_controller_spec.rb
```diff
@@ -114,7 +114,63 @@
       it 'redirects the user to the manage metadata page' do
         sign_in
 
-        post :new, search_field: 'short_name', search_query: 'dunno'
+        post :new
+
+        expect(response).to redirect_to(manage_metadata_path)
```
abbottry: Maybe this should go to `new_bulk_updates_search_path` like the others? I don't think we want to use Manage Metadata as an entry point generally.

spec/controllers/bulk_updates_controller_spec.rb
```diff
@@ -114,7 +114,63 @@
...
+      before do
+        allow(Mmt::Application.config).to receive(:bulk_updates_enabled).and_return(false)
+      end
+
+      it 'redirects the user to the manage metadata page' do
+        sign_in
+
+        get :preview
+
+        expect(response).to redirect_to(manage_metadata_path)
```
abbottry: Shouldn't GET requests get sent to `new_bulk_updates_search_path`?

spec/controllers/bulk_updates_controller_spec.rb
charleshuang80: this `context` is `'when bulk updates are disabled'`

spec/controllers/bulk_updates_controller_spec.rb
charleshuang80: I believe that is what the test in the `it` block below is verifying

spec/controllers/bulk_updates_controller_spec.rb
charleshuang80: This is also `'when bulk updates are disabled'`

app/views/bulk_updates/preview.html.erb
```diff
@@ -1,7 +1,77 @@
...
+  </section>
+
+  <section id="bulk-update-preview-submit">
+    <!-- Form submitting data to create the bulk update -->
+    <%= form_tag(bulk_updates_path, method: :post, id: 'bulk-update-preview-submit-form') do %>
+      <% params[:concept_ids].each do |collection| %>
+        <%= hidden_field_tag 'concept-ids[]', collection %>
+      <% end %>
+
+      <%= hidden_field_tag 'update-field', params['update-field'] %>
```
macrouch: Shouldn't field names be underscored and not hyphenated? `update_field`, etc.

charleshuang80: Normally, yes, but I'm trying to make it as close to what we will pass to CMR as possible for minimal modifications, and the keys in the JSON they are expecting are `"concept-ids", "update-type", "update-field", "find-value",` and `"update-value"`

abbottry: Yea I hate this too, I have no idea why CMR chose hyphens or underscored, I've never seen this in practice anywhere else. But, its what we got.

macrouch: I believe Clojure best practice is hyphens, where Ruby is underscores. Most of the time (search/ingest/etc.) they accept underscored parameters. Do they not for this?

charleshuang80: I brought up the topic with Lauren a while ago. I referenced [CMR-2579](https://bugs.earthdata.nasa.gov/browse/CMR-2579) and [MMT-525](https://bugs.earthdata.nasa.gov/browse/MMT-525), as something that both teams had to do with regards to the groups endpoints. I didn't push hard for it though, so she said she was going to keep it with hypens. Should we ask them to change it?

abbottry: I'd prefer that they make these changes rather than us massage responses for the sake of changes hyphens to underscores. I think it's ugly, and it bothers me, but its the API decision, not ours.

PR #61, opened 2017-06-07
MMT-893 Added GES DISC specific validations

macrouch:


charleshuang80:
Should there be any tests for this?

PR #62, opened 2017-06-08
MMT-930: Updates display of bulk updates using newly provided data from CMR

abbottry:


app/assets/javascripts/bulk_updates.coffee
```diff
@@ -386,3 +385,14 @@ $(document).ready ->
       messages:
         'bulk_update_collections[]':
           required: 'You must select at least 1 collection.'
+
+  if $('#bulk-update-errors-table').length > 0
```
macrouch: If this is more of a status table, shouldn't we change the name?

app/views/shared/bulk_updates/display/_science_keywords.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  <%= display_value_title(bulk_update['update-type'], 'find') %>
+
+  <ul class="arrow-tag-group-list">
+    <% display_science_keyword(bulk_update['find-value']).each do |field, value| %>
+      <li itemprop="keyword" class="arrow-tag-group-item"><%= value %></li>
+    <% end %>
+  </ul>
+<% end %>
+
+<% if bulk_update['update-type'] == 'ADD_TO_EXISTING' || bulk_update['update-type'] == 'CLEAR_ALL_AND_REPLACE' %>
```
charleshuang80: the wiki says "If NOT type FIND_AND_REMOVE, New value required", (and FIND_AND_REPLACE should have the new aka update value), so that might be a better test here

app/concerns/bulk_updates.rb
```diff
@@ -12,6 +12,36 @@ def retrieve_bulk_updates
...
+      # request object and insert them into the root
+      task['request-json-body'].each do |key, value|
+        task[key] = value
+      end
+    end
+  end
+
+  # The response from CMR only includes the concept-id
+  # but for display purposes we need detailed information
+  def hydrate_collections(bulk_update)
```
charleshuang80: very minor thing, but you're passing `@task` into both the hydrate methods, but in `hydrate_task` you call the variable `task` and for `hydrate_collections` you call it `bulk_update`. I know we're using them interchangeably but seems odd since the two methods are so close together (both in the concern and in usage).

app/views/shared/bulk_updates/display/_science_keywords.html.erb
charleshuang80: I'm going to approve this, but I think this line needs to be changed. I would suggest `<% unless bulk_update['update_type'] == 'FIND_AND_REMOVE' %>`.
The other item was just a thought, I'll leave it to you if you feel like changing anything there.

abbottry: That's consistent with our check elsewhere as well.

PR #64, opened 2017-06-12
MMT-458: add Version field to Metadata Association on Data Identification form

charleshuang80:


Gemfile
```diff
@@ -102,4 +102,5 @@ gem 'builder'
 
 # run this command to work from a local copy of the gem's repo
 # bundle config local.cmr_metadata_preview /path/to/local/git/repository
-gem 'cmr_metadata_preview', git: 'https://git.earthdata.nasa.gov/scm/cmr/cmr_metadata_preview.git', branch: 'master'
+# gem 'cmr_metadata_preview', git: 'https://git.earthdata.nasa.gov/scm/cmr/cmr_metadata_preview.git', branch: 'master'
+gem 'cmr_metadata_preview', git: 'https://git.earthdata.nasa.gov/scm/cmr/cmr_metadata_preview.git', branch: 'MMT-458'
```
macrouch: Be sure to change this back before merging. This file should not change, but Gemfile.lock should have a new revision.

charleshuang80: Yeah, that's why I thought it was ok to leave the commented line in there as I would change it back before merging.
It's a funny process, but essentially there's no way to prevent running the tests after the preview gem PR gets merged.

spec/factories/valid_draft_factories.rb
```diff
@@ -815,7 +819,7 @@
           'Eon' => 'test 3 Eon text 1'
         }]
       }],
-      'ISOTopicCategories' => %w(farming climatologyMeteorologyAtmosphere health)
+      'ISOTopicCategories' => %w[farming climatologyMeteorologyAtmosphere health]
```
macrouch: Rubocop says this should be `%w()` instead of `%w[]`

charleshuang80: LOL, I only changed this because my Rubocop said it should be `%w[]`

macrouch: Thats not good. What version of Rubocop do you have? I have always seen it prefer `()`

I have 0.47.1

charleshuang80: I think they updated the style guide to use `[]`, https://github.com/bbatsov/ruby-style-guide#percent-w.

charleshuang80: I have 0.48.1

macrouch: Yeah, looks like it was changed a couple of months ago. I updated everything and now I don't see the warning

abbottry: %w[]!? really.. thats odd. Now we'll have a bunch of warning for the previous version hah

PR #68, opened 2017-06-12
MMT-945: Removes the BETA tag

abbottry:


charleshuang80:
also approving (though it didn't ask me)

PR #74, opened 2017-06-30
MMT-1038: Prepares the codebase for UMM-V

abbottry:


db/schema.rb
```diff
@@ -11,7 +11,17 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170411191927) do
+ActiveRecord::Schema.define(version: 20170630141235) do
+
+  create_table "bulk_update_searches", force: :cascade do |t|
```
charleshuang80: seems odd that this change made it into the PR. also, are we planning on storing these at some point?

app/views/collection_drafts/new.html.erb
```diff
@@ -44,7 +44,7 @@
         <% end %>
       </div>
 
-      <%= render :partial => "drafts/forms/#{@draft_form}", :locals => { draft: @draft } %>
+      <%= render partial: "collection_drafts/forms/#{@draft_form}", :locals => { draft: get_resource } %>
```
charleshuang80: since you changed it from `:partial =>` to `partial:`, maybe you want to change `:locals =>` to `locals:` also?

app/views/collection_drafts/index.html.erb
```diff
@@ -0,0 +1,45 @@
+<div class="row row-content">
+  <section>
+    <h2><%= current_user.provider_id %> Drafts</h2>
+
+    <% if get_resources.any? %>
+      <p><%= page_entries_info get_resources, entry_name: 'Draft' %></p>
```
charleshuang80: Should this be 'Collection Draft'?

app/views/drafts/index.html.erb
```diff
@@ -2,8 +2,8 @@
   <section>
     <h2><%= current_user.provider_id %> Drafts</h2>
 
-    <% if @drafts.any? %>
-      <p><%= page_entries_info @drafts, entry_name: 'Draft' %></p>
+    <% if get_resources.any? %>
+      <p><%= page_entries_info get_resources, entry_name: 'Draft' %></p>
```
charleshuang80: Should we change this and the title (on line 3) to use something like `resource_name.titleize`?

spec/features/drafts/draft_creation_spec.rb
```diff
@@ -28,9 +28,9 @@
 
     context 'when saving data into the draft' do
       before do
-        within '.metadata' do
-          click_on 'Collection Information'
-        end
+        # within '.metadata' do
```
charleshuang80: should this just be removed?

app/views/drafts/index.html.erb
abbottry: This isnt used in this branch, however it IS used in MMT-998 so I'll make this change there.

PR #76, opened 2017-07-14
MMT-1049 handle orders that do not have a contact name

macrouch:


app/helpers/orders_helper.rb
```diff
@@ -45,6 +45,6 @@ def contact_info(contact_address)
   end
 
   def contact_name(contact)
-    [contact['FirstName'], contact['LastName']].reject(&:empty?).join(' ')
+    [contact['FirstName'], contact['LastName']].reject(&:empty?).join(' ') unless contact.nil?
```
abbottry: What is the default return value for a method? That would be returned here right? IF it's `nil` will that cause problems with whatever is calling this method? If it's just `<%= contact_name(@stuff) %>` then we'll be fine.

macrouch: I can't remember what it was, but I tested it with this and it worked

```
contact = nil
[contact['FirstName'], contact['LastName']].reject(&:empty?).join(' ') unless contact.nil?
```

abbottry: Within `app/views/orders/show.html.erb` the method `contact_info` passes a local variable to the 'contact_information' partial. There are a number of calls in there that use nested hash calls, so it seems like those would throw an error. Do we have a test that tests for this? I presume our tests have a contact person so we'll not see this in our tests, we should add one.

macrouch: Added tests, which did reveal more problems

app/helpers/orders_helper.rb
```diff
@@ -38,13 +38,17 @@ def order_owner(owner_guid)
...
+
     contact_address['Name'] = contact_name(contact_address)
     contact_address['Phone'] = contact_address.fetch('Phones', {}).fetch('Item', {})
 
     contact_address
   end
 
   def contact_name(contact)
-    [contact['FirstName'], contact['LastName']].reject(&:empty?).join(' ') unless contact.nil?
+    contact = {} if contact.nil?
```
abbottry: The desired result here is a string 'FirstName LastName' but if nothing is supplied we're returning an empty {} -- We should match the return types. Ultimately I'm not sure the nil check is necessary?

macrouch: If nothing is supplied, this method returns an empty string. I'm not sure what you mean

abbottry:
If it's nil it returns {}, unless I'm seeing an odd diff on git.

macrouch:
@abbottry It might be git looking weird, here is the code:

```
contact = {} if contact.nil?

[contact.fetch('FirstName', ''), contact.fetch('LastName', '')].reject(&:empty?).join(' ')
```

abbottry:
👌🏻

PR #78, opened 2017-07-18
MMT-998: Adds tests for creating, saving, and editing UMM-Vars in MMT

charleshuang80:


abbottry:
One of the rebases was off -- there is a fair amount of stuff here that shouldn't be. I'll have to work on fixing that this morning.

PR #79, opened 2017-07-19
MMT-1001 delete variable drafts

macrouch:


app/controllers/base_drafts_controller.rb
```diff
@@ -102,7 +102,7 @@ def destroy
     Rails.logger.info("Audit Log: Draft #{get_resource.entry_title} was destroyed by #{current_user.urs_uid} in provider: #{current_user.provider_id}")
 
     respond_to do |format|
-      format.html { redirect_to manage_metadata_path, flash: { success: 'Draft was successfully deleted.' } }
+      format.html { redirect_to send("#{plural_resource_name}_path"), flash: { success: 'Draft was successfully deleted.' } }
```
abbottry: Looking through the design comp discussion there have been many requests to specify the type of draft. So rather than saying "draft has been deleted" we would want this to say "variable draft has been deleted"

We're currently using en.yml for this copy within the bar drafts controller and I think it'd be a good thing to continue using going forward.

app/views/drafts/show.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="row row-content">
+  <section class="action">
+    <div class="cta">
+      <%#= link_to 'Publish Draft', publish_collection_draft_path(get_resource), method: :post, class: 'eui-btn--blue spinner' %>
+      <%= link_to 'Delete Draft', "#delete-draft-modal", class: 'display-modal' %>
+      <div id="delete-draft-modal" class="eui-modal-content">
+        <a href="#" class="modal-close float-r"><i class="fa fa-times"></i><span class="is-invisible">Close</span></a>
+        <p>Are you sure you want to delete this draft?</p>
```
abbottry: Same minor detail here, we need to specify the type of draft.  We can use `resource_name` in these cases of shared views.

spec/features/variable_drafts/variable_draft_deletion_spec.rb
```diff
@@ -0,0 +1,41 @@
+# MMT-58
```
abbottry: Wrong ticket number (copy and paste I'm sure)

app/controllers/collections_controller.rb
```diff
@@ -16,7 +16,7 @@ def show
   def edit
     draft = CollectionDraft.create_from_collection(@collection, current_user, @native_id)
     Rails.logger.info("Audit Log: Draft for #{draft.entry_title} was created by #{current_user.urs_uid} in provider #{current_user.provider_id}")
-    flash[:success] = 'Draft was successfully created'
+    flash[:success] = 'Collection Draft Created Successfully!'
```
abbottry: It looks like you added all of these translations to en.yml but you're not utilizing it here.

app/views/drafts/show.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="row row-content">
+  <section class="action">
+    <div class="cta">
+      <%#= link_to 'Publish Draft', publish_collection_draft_path(get_resource), method: :post, class: 'eui-btn--blue spinner' %>
+      <%= link_to 'Delete Draft', "#delete-draft-modal", class: 'display-modal' %>
```
abbottry: Should probably be `Delete Variable Draft` -- consider https://stackoverflow.com/questions/13547144/i18n-for-model-specific-rails-submit-button, IF it would work. I know this is a modal so it would likely need to be specified but for forms and related buttons we could/should use the method described in that answer.

PR #81, opened 2017-07-20
Mmt 1033 - Change Manage Metadata to Manage Collections page

rdcody:
- Changed menu option from Manage Metadata to Manage Collections
- Changed all paths, layouts, controller names and variables from Manage Metadata to Manage Collections
- Moved button and changed name to "Create New Record"
- Changed the "Or use the search..." text to more relevant wording and added link that when clicked changes focus to search box.
- Added display:none to radio button indicating "Create New Collection Record" selection, as we may want to implement radio button functionality in the future with "Update".


macrouch:
When viewing a collection draft, the spacing looks off of the orange highlight line

screen shot 2017-07-21 at 3 21 45 pm

rdcody:
> hlight line

rdcody:
> When viewing a collection draft, the spacing looks off of the orange highlight line

Looks like this currently exists within production; In addition, this is seen on the collection drafts page rather than the Manage Collections page. I will look into a fix for this, but it may need to be a minor bug fix or on another ticket.

spec/features/collection_drafts/progress_circles_spec.rb
```diff
@@ -7,10 +7,9 @@
     before do
       login
 
-      visit '/manage_metadata'
-      choose 'New Collection Record'
-      click_on 'Create Record'
-      
+      visit '/manage_collections'
```
charleshuang80: I think we should be using `visit manage_collections_path` instead of `visit '/manage_collections'` if possible

app/views/manage_collections/show.html.erb
```diff
@@ -4,46 +4,28 @@
...
-              <%= radio_button_tag(:type, "from_template", nil, disabled: true, id: 'from-template') %>
-              <label class="disabled" for="from-template">From Template</label><span class="coming-soon">Coming Soon</span>
-              <!-- <select name="select-template" class="single-select">
-                <option value="Select Template">Select Template</option>
-              </select> -->
-            </div>
-            <div class="row">
-              <%= radio_button_tag(:type, "blank_record", nil, disabled: true, id: 'blank-record') %>
-              <label class="disabled" for="blank-record">From Blank Record</label><span class="coming-soon">Coming Soon</span>
+              <p><strong>OR</strong> use the <a id="search-focus" href="#">search</a> in the top right corner to find a published collection to clone or edit.</p>
```
charleshuang80: should the link be linking to something?

app/controllers/drafts_controller.rb
```diff
@@ -1,6 +1,6 @@
 # :nodoc:
-class DraftsController < ManageMetadataController
-  include ManageMetadataHelper
-  
+class DraftsController < ManageCollectionsController
```
charleshuang80: @rdcody, @abbottry, it seems a little weird to me to have the DraftsController inherit from a `ManageCollectionsController`. This would mean that VariableDrafts and eventually ServiceDrafts would also be inheriting from that controller. I know we need to change the UI to represent 'Manage Collections' instead of 'Manage Metadata' and to create space for a 'Manage Variables'. But maybe we should still have a `ManageMetadataController` that sits above both of those controllers as well as the `DraftsController`?

app/controllers/manage_collections_controller.rb
```diff
@@ -1,12 +1,12 @@
 # :nodoc:
-class ManageMetadataController < PagesController
+class ManageCollectionsController < PagesController
```
charleshuang80: If there are any changes from my comments to `drafts_controller.rb`, this may require a modification as well.

app/views/manage_collections/show.html.erb
rdcody: Clicking this triggers a jQuery function to change focus to the search box in the top right. So there the # is just a placeholder for that.

spec/features/collection_drafts/progress_circles_spec.rb
rdcody: Changed this and removed a commented out versions of this.

PR #82, opened 2017-07-25
MMT-1003 publishing variable drafts

macrouch:


app/controllers/variable_drafts_controller.rb
```diff
@@ -13,6 +14,45 @@ def new
     set_current_form
   end
 
+  def show
+    super
+
+    validate_metadata
+  end
+
+  def publish
```
abbottry: In an effort to use rails conventions as often as possible, it seems as though we could use `create` instead of `publish` which would allow us to use the standard 7 actions within rails.

app/models/draft.rb
```diff
@@ -15,4 +16,9 @@ def display_short_name
   def default_values
     self.draft ||= {}
   end
+
+  def set_native_id
+    self.native_id ||= "mmt_collection_#{id}"
```
abbottry: Moving this makes sense but I don't think we want variables having a native id with `collection` in it. Maybe this method should use the draft_type or class to determine this portion of the string?

app/views/drafts/show.html.erb
```diff
@@ -0,0 +1,26 @@
...
+      <% if @errors && @errors.size > 0 %>
+        <%= link_to 'Publish Draft', '#invalid-draft-modal', class: 'eui-btn--blue display-modal' %>
+        <div id="invalid-draft-modal" class="eui-modal-content">
+          <p>This variable draft is not ready to be published. Please use the progress indicators on the draft preview page to address incomplete or invalid fields.</p>
+          <p>
+            <a href="javascript:void(0)" class="eui-btn modal-close">Ok</a>
+          </p>
+        </div>
+      <% else %>
+        <%= link_to 'Publish Draft', send("publish_#{resource_name}_path", get_resource), method: :post, class: 'eui-btn--blue spinner' %>
```
abbottry: `Publish Variable Draft` or use the resource_name/i18n like you are below.

lib/cmr/cmr_client.rb
```diff
@@ -173,6 +173,25 @@ def delete_collection(provider_id, native_id, token)
...
+      # if native_id is not url friendly or encoded, it will throw an error so we check and prevent that
+      if Rails.env.development? || Rails.env.test?
+        url = "http://localhost:3002/providers/#{provider_id}/variables/#{encode_if_needed(native_id)}"
+      else
+        url = "/ingest/providers/#{provider_id}/variables/#{encode_if_needed(native_id)}"
+      end
+
+      headers = {
+        'Accept' => 'application/json',
+        'Content-Type' => "application/#{Rails.configuration.umm_version}; charset=utf-8"
```
abbottry: Do we want the umm_version here? Based on my conversations with CMR we are likely to have a new version and we'll start splitting these out into `umm_c_version` and `umm_v_version` -- either way, var isn't versioned yet.

lib/cmr/cmr_client.rb
```diff
@@ -173,6 +173,25 @@ def delete_collection(provider_id, native_id, token)
...
+      }
+
+      # content_type is passed if we are reverting to a revision with a different format
```
abbottry: I think we only have one format of Var, right? This can likely be removed.

lib/json_schema_form/umm_json_form.rb
```diff
@@ -152,7 +152,10 @@ def convert_key_to_type(value, type)
     return (value.casecmp('true') >= 0 ? true : false) if type == 'boolean'
 
     # Numbers
-    return UmmUtilities.convert_to_integer(value) if type == 'number'
+    return UmmUtilities.convert_to_number(value) if type == 'number'
+
+    # Integers
+    return UmmUtilities.convert_to_integer(value) if type == 'integer'
```
abbottry: We weren't able to come up with a use case for this, is there something in the schema that is defined as an integer vs a number?

app/controllers/variable_drafts_controller.rb
```diff
@@ -13,6 +14,45 @@ def new
...
+    draft = get_resource.draft
+
+    ingested = cmr_client.ingest_variable(draft.to_json, get_resource.provider_id, get_resource.native_id, token)
+
+    if ingested.success?
+      # get information for publication email notification before draft is deleted
+      Rails.logger.info("Audit Log: Draft #{get_resource.entry_title} was published by #{current_user.urs_uid} in provider: #{current_user.provider_id}")
+      user_info = get_user_info
+      short_name = get_resource.draft['ShortName']
```
abbottry: Variable doesn't have 'ShortName', might make more sense to use `get_resource.short_name` which is on the model and both collections and variables write to.

app/controllers/variable_drafts_controller.rb
```diff
@@ -13,6 +14,45 @@ def new
...
+      version = get_resource.draft['Version']
```
abbottry: I'm not sure we set 'Version' anywhere do we? Unless you did it as part of this ticket.

lib/json_schema_form/umm_json_form.rb
macrouch: I don't think there are any integer fields right now, but previously this was converting to an integer if the type was number, which was causing problems. Seemed like a good idea to cover all the bases

abbottry: Ok, cool.

PR #84, opened 2017-07-26
Mmt 1040

rdcody:
-Moved all items in header to the menu.
-Added drop down menu for user info.
-Added transition states to menu links
-Changed some styling and added new classes and IDs
-Added hover states for menu links in header and nav
-Added caret that rotates when clicked to user-info


app/views/shared/_search_form.html.erb
```diff
@@ -1,3 +1,30 @@
+
```
abbottry: The advanced search just used the markup from the upper right, so the new user dropdown is duplicated and I don't think it should be. 

screen shot 2017-07-27 at 8 56 34 am

abbottry: The alignment of the `Find` button is no longer aligned properly
screen shot 2017-07-27 at 9 01 19 am

app/assets/javascripts/search.coffee
```diff
@@ -42,3 +42,9 @@ $(document).ready ->
 # Handle presence of Javascript by turning off or on visibility of JS sensitive objects
 $('.js-disabled-object').css 'visibility': 'hidden'
 $('.js-enabled-object').css 'visibility': 'visible'
+
+# Change focus of cursor when search link is clicked on Manage Collections, Manage Variables, or Manage Services pages.
+$ ->
```
rdcody: Just to clarify: you mean giving focus to the search box will close the dropdown menu?

abbottry: Yea -- As I type in the search field it's covered by the drop down

app/views/shared/_search_form.html.erb
rdcody: Fixed.

rdcody: Fixed.

app/assets/javascripts/search.coffee
rdcody: Fixed.

app/assets/javascripts/search.coffee
```diff
@@ -42,3 +42,17 @@ $(document).ready ->
...
+ $('#search-focus').on 'click', ->
+    document.getElementById('keyword').focus()
```
macrouch: @abbottry Do you think this should use jQuery? It probably works just fine, but the rest of the app uses jQuery for this kind of thing

rdcody: I can use either; Although, I haven't see jQuery in the app. I've only seen coffee. I may need you to point me in the right direction.

rdcody: I've changed this over to jQuery syntax to match the other files.

abbottry: Yes, jQuery -- good call @rdcody on just making the change.

mreese84:
Looks pretty good. I'd change the provider context badge to fire the switch provider modal. But everything else looks good.

app/assets/javascripts/search.coffee
charleshuang80: I believe what @macrouch may have meant (correct me if I'm wrong) was that you can use `$('#keyword')` here instead of `document.getElementById('keyword')`

charleshuang80: I think what @macrouch may have meant (correct me if I'm wrong) was that you can use `$('#keyword')` here instead of `document.getElementById('keyword')`

spec/features/provider_context/provider_context_spec.rb
```diff
@@ -72,7 +73,7 @@
         select 'MMT_1', from: 'select_provider'
 
         wait_for_ajax
-
+        click_on 'profile-link'
```
charleshuang80: Since we now have 2 links that will prompt a user to change their provider context (the badge and the link in the profile link dropdown) should we test both ways to make sure they're working?

app/assets/javascripts/search.coffee
rdcody: Changed. Thanks for the heads-up Charles. I missed some of the JS.

spec/features/provider_context/provider_context_spec.rb
rdcody: Yes, I will add a second test to test this.

PR #85, opened 2017-07-26
MMT-1048: Add Map as subtype for Visualization URL

macrouch:


abbottry:
Pretty decent schema change in this as well, is this something we missed before or new schema changes that we need to ensure are accounted for?

macrouch:
This was the first time I had seen the schema changes as well, I thought schema changes meant a new schema version. 

Most of them are just descriptions, so we wouldn't need to do anything. The `SpatialCoverageTypeEnum` didn't make sense to me, so I didn't add those values to the dropdown

PR #89, opened 2017-08-01
MMT-1033: Layout changes to Manage Collections and Path change to prepare for manage variables

rdcody:
-Change Manage Metadata paths so they are Manage Collections.
-Wrote if statement to give ability for the same layout to manage variables and collections.
-Changed text on search area.
-Added ability to click search link and change focus to search box.


config/routes.rb
```diff
@@ -85,7 +85,9 @@
 
   get 'search' => 'search#index', as: 'search'
 
-  resource :manage_metadata, only: :show, controller: 'manage_metadata'
+  varialbe = ['asdf']
```
charleshuang80: typo/something left in?

nohup.out
```diff
@@ -0,0 +1 @@
+Error: Could not find or load main class cmr.dev_system.runner
```
charleshuang80: not sure how this got included

config/routes.rb
abbottry: ha! woops. My bad @rdcody

rdcody: Ah, yes. Ryan was showing me something and I forgot to remove this. Removed.

nohup.out
rdcody: I had to manually add nohup.out to the gitignore file. I will remove this.

app/controllers/manage_metadata_controller.rb
```diff
@@ -1,20 +1,7 @@
 # :nodoc:
 class ManageMetadataController < PagesController
-  include BulkUpdates
   include ManageMetadataHelper
 
-  layout 'manage_metadata'
+  # layout 'manage_metadata'
```
charleshuang80: Does this mean we're not using that layout?

rdcody: The layout is being used but not within this controller. Two new controllers have been created for collections and variables where it will be used. This controller is just a container for collections and variables now.

app/views/layouts/manage_metadata.html.erb
```diff
@@ -33,7 +33,10 @@
 
               <nav>
                 <div class="content">
-                  <h2 class="current"><%= link_to 'Manage Metadata', manage_metadata_path %></h2>
+                  <%= content_tag(:h2, class: ('current' if controller_name == 'manage_collections' || 'collection_drafts')) do %>
```
macrouch: I think we should use HTML here instead of a ruby helper for the h2

```
<h2 class="<%= 'current' if controller_name == 'manage_collections' || controller_name == 'collection_drafts' %>">
....
</h2>
```

rdcody: Changed this; Thanks for the tip!

app/controllers/manage_metadata_controller.rb
rdcody: I've removed the commented layout.

abbottry:
I'm seeing an issue with the 'header', notice the gap at the bottom of the grey stripe on the ride side.
screen shot 2017-08-02 at 10 35 34 pm

rdcody:
@abbottry It seems the issue with the grey stripe is currently existing in SIT. This is not an issue created by this ticket. I will look into it and see if this is something I can fix easily.

PR #92, opened 2017-08-02
MMT-1034: Added Variables and Services

rdcody:
-Added Variables and Services controllers, paths, and views.
-Left Services commented out.
-Reorganization and styling will need to be done before Services is added because there isn't enough room in the header. 
-Fixed a problem with the modal window not being rounded because of Overflow: scroll.


spec/features/collection_drafts/collection_draft_deletion_spec.rb
```diff
@@ -33,8 +33,13 @@
     #   end
     # end
 
+<<<<<<< HEAD
```
charleshuang80: rebase merge conflict stuff left in...

app/controllers/manage_collections_controller.rb
```diff
@@ -1,12 +1,11 @@
 # :nodoc:
 class ManageCollectionsController < PagesController
   include BulkUpdates
-  include ManageCollectionsHelper
 
-  layout 'manage_collections'
+  layout 'manage_metadata'
```
charleshuang80: I thought we had discussed keeping a bare `ManageMetadataController` that the controllers for ManageCollections, ManageVariables, and ManageServices could all inherit from. Then we wouldn't need to have to declare this layout in the three different controllers. Had that decision been changed, or did those edits get lost somehow? I recall seeing that and discussing it on the PR review for 1033 the other day.

nohup.out
```diff
@@ -0,0 +1 @@
+Error: Could not find or load main class cmr.dev_system.runner
```
charleshuang80: sneaked in again!

app/controllers/manage_collections_controller.rb
rdcody: I think the plan was to keep the Manage Metadata controller empty, but also having a manage collections, variables and services controller. Although, we would only have one layout, the manage metadata layout, to accommodate all of their views. May need @macrouch or @abbottry to verify that, but that is what is in MMT-1033.

rdcody: Added back in the Manage Metadata Controller.

PR #93, opened 2017-08-02
MMT-1086 updated UMM-C v1.9 schema files

macrouch:


charleshuang80:
@abbottry, following your call with them, do we need to worry about any functionality for the "HORIZONTAL_VERTICAL" and "ORBITAL_VERTICAL" additions to the enum?

app/views/collection_drafts/forms/fields/_geometry_fields.html.erb
```diff
@@ -48,74 +48,58 @@
   <% g_polygons_checked = !geometry['GPolygons'].nil? %>
```
charleshuang80: trying to add a comment to line 30:
- the required icon is not showing up for Coordinate System for me, so I was showing an error on the preview circles (for Spatial Extent), and when I clicked on it, no validation prompt was telling me what was missing on the form
- When I click on the `clear` link under Coordinate System, any geometry information I have typed into the text box inputs (points, bounding rectangles, etc) gets cleared out as well (but not the checkboxes). Not sure if that is the desired behavior

spec/features/collection_drafts/forms/spatial_information_form_spec.rb
```diff
@@ -545,9 +535,8 @@
...
       # Spatial Extent
       within '.spatial-extent' do
-        expect(page).to have_no_checked_field('Horizontal')
-        expect(page).to have_no_checked_field('Vertical')
-        expect(page).to have_checked_field('Orbit')
+        expect(page).to have_field('Spatial Coverage Type', with: 'ORBITAL')
+
         expect(page).to have_field('Swath Width', with: '1.0')
         expect(page).to have_field('Period', with: '2.0')
         expect(page).to have_field('Inclination Angle', with: '3.0')
```
charleshuang80: I know this might be tedious (sorry), but I feel like we should add tests for the new options for Spatial Coverage Type (that they said folks likely wouldn't use) - `Horizontal and Vertical` and `Orbital and Vertical`

lib/assets/schemas/umm-cmn-json-schema.json
```diff
@@ -1498,7 +1500,7 @@
     },
     "SpatialCoverageTypeEnum": {
       "type": "string",
-      "enum": ["HORIZONTAL", "VERTICAL", "ORBITAL"]
+      "enum": ["HORIZONTAL", "VERTICAL", "ORBITAL", "HORIZONTAL_VERTICAL", "ORBITAL_VERTICAL"]
     },
     "CoordinateSystemEnum": {
       "type": "string",
```
charleshuang80: I just downloaded the schema and compared what we have in this PR.
comment for line 1541: we have `THREADS` (for the first two instances) and the version online has `THREDDS` (for all three instances)
comment for line 1553: we have ...`google-earch`... and the online version has `google-earth`

what we have in this file is also in the merged one.

charleshuang80:
I think the cmr_metadata_preview also needs to be updated to handle the new spatial coverage types. I had input values for `Orbital and Vertical`, and somehow the Orbital values were displayed (it seems like they maybe shouldn't have been based on how the code is set up) but the Vertical values were not.

PR #94, opened 2017-08-02
MMT-1075 integrate MMT with TopHat2

macrouch:


app/views/shared/_header.html.erb
```diff
@@ -1,4 +1,4 @@
-<header role="eui-banner">
+<header class="mmt-header" role="eui-banner">
```
abbottry: We have a `header` tag in manage_metadata and manage_cmr layouts as well -- do we need this change there as well?

macrouch: It looks like almost all of `header.mmt-header` is overwritten by `.internal header`, which explains why I couldn't find any visual difference 
screen shot 2017-08-02 at 5 40 40 pm

PR #96, opened 2017-08-03
MMT-1000: adds ability to edit a draft created from a published Variable

charleshuang80:
* add create_from_variable method to VariableDraft
* add controller methods for editing a draft from published variable
* add methods to search and retrieve variables from cmr
* add routes and link for edit
* add model, controller, and integration tests
* add tests helper method to publish variable, modified name of existing method for collections

### DIFF
--- a/app/controllers/variables_controller.rb
+++ b/app/controllers/variables_controller.rb
@@ -1,6 +1,6 @@
 # :nodoc:
 class VariablesController < ManageMetadataController
-  before_action :set_variable, only: [:show, :edit]]
+  before_action :set_variable, only: [:show, :edit]
   before_action :set_schema, only: [:show]
   before_action :set_form, only: [:show]
 
--- a/app/controllers/variables_controller.rb
+++ b/app/controllers/variables_controller.rb
@@ -1,19 +1,24 @@
 # :nodoc:
-class VariablesController < ManageMetadataController
+class VariablesController < ApplicationController
   before_action :set_variable, only: [:show, :edit]
 
   def show
   end
 
   def edit
-    draft = VariableDraft.create_from_variable(@variable, current_user, @native_id)
-    Rails.logger.info("Audit Log: Variable Draft for #{draft.entry_title} was created by #{current_user.urs_uid} in provider #{current_user.provider_id}")
-    flash[:success] = I18n.t('controllers.draft.variable_drafts.create.flash.success')
-    redirect_to variable_draft_path(draft)
+    if @native_id
+      draft = VariableDraft.create_from_variable(@variable, current_user, @native_id)
+      Rails.logger.info("Audit Log: Variable Draft for #{draft.entry_title} was created by #{current_user.urs_uid} in provider #{current_user.provider_id}")
+      flash[:success] = I18n.t('controllers.draft.variable_drafts.create.flash.success')
+      redirect_to variable_draft_path(draft)
+    else
+      # if we cannot locate the native_id for the Variable, we should discontinue editing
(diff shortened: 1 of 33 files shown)


app/models/variable_draft.rb
```diff
@@ -8,6 +8,28 @@ class << self
...
+
+    def create_from_variable(variable, user, native_id)
+      new_long_name = (variable['LongName'].blank? ? nil : variable['LongName'])
+
+      if native_id
+        # Editing a record
+        draft = VariableDraft.find_or_initialize_by(native_id: native_id)
+        draft.entry_title = new_long_name
+        draft.short_name = (variable['Name'].blank? ? nil : variable['Name'])
+      else
```
macrouch: I don't see a ticket for the clone functionality

charleshuang80: I wasn't sure what to do with the case where a native_id didn't come through properly, so I left in what was happening with collections and cloning. Should I just remove this, make it save a new native_id, or something else?

app/controllers/variables_controller.rb
```diff
@@ -1,9 +1,19 @@
...
+  def edit
+    draft = VariableDraft.create_from_variable(@variable, current_user, @native_id)
+    Rails.logger.info("Audit Log: Variable Draft for #{draft.entry_title} was created by #{current_user.urs_uid} in provider #{current_user.provider_id}")
+    flash[:success] = I18n.t('controllers.draft.variable_drafts.create.flash.success')
+    redirect_to variable_draft_path(draft)
+  end
+
   def create
-    variable_draft = VariableDraft.find(params[:id])
+    @variable_draft = VariableDraft.find(params[:id])
```
macrouch: this `@` breaks publishing

app/controllers/variables_controller.rb
```diff
@@ -31,4 +41,35 @@ def create
...
   end
+
+  private
+
+
+  def set_variable
+    @concept_id = params[:id]
+    @revision_id = params[:revision_id]
+
+    # search for variable by concept id to get the native_id
```
charleshuang80: add loop to make sure cmr has time to index

app/models/variable_draft.rb
```diff
@@ -8,6 +8,17 @@ class << self
     def forms
       []
     end
+
+    def create_from_variable(variable, user, native_id)
+      draft = VariableDraft.find_or_initialize_by(native_id: native_id)
+      draft.entry_title = (variable['LongName'].blank? ? nil : variable['LongName'])
```
abbottry: if `'LongName'` doesn't exist, the `[]` operator will return `nil`, this ternary operator is unnecessary. (same thing goes for below)

abbottry: If your concern is that it's being provided, but is an empty string.. then this can remain.

lib/cmr/cmr_client.rb
```diff
@@ -42,6 +42,16 @@ def get_collections_by_post(query, token = nil)
       post(url, query.to_query, headers.merge(token_header(token)))
     end
 
+    def get_variables(options = {}, token = nil)
```
abbottry: If this is the same as the method below (`search_collections`) we should keep the naming consistent.

app/models/variable_draft.rb
charleshuang80: I had just kept it from the method for collections. I'll just remove the unnecessary parts

lib/cmr/cmr_client.rb
charleshuang80: actually I was trying to mirror `get_collections` above, though they are similar and it could be either. I had tested trying to make a `get_variables_by_post`, but they only have the GET endpoint. Should I change the name of the method or leave it?

abbottry:
Needs breadcrumbs on edit action

PR #97, opened 2017-08-03
MMT-1007 delete variable record from CMR

macrouch:


macrouch:
Branched off of MMT-1000, you can just view my commit to see my changes

app/views/variables/show.html.erb
```diff
@@ -13,6 +13,19 @@
         <% elsif available_provider?(@provider_id) %>
           <%= link_to 'Edit Record', '#not-current-provider-modal', class: 'display-modal not-current-provider', data: { 'provider': @provider_id, record_action: 'edit-variable' } %> |
         <% end %>
+        <% if current_provider?(@provider_id) %>
+          <%= link_to 'Delete Variable Record', "#delete-record-modal", class: 'display-modal delete-collection' %>
+        <% elsif available_provider?(@provider_id) %>
+          <%= link_to 'Delete Variable Record', '#not-current-provider-modal', class: 'display-modal not-current-provider', data: { 'provider': @provider_id, record_action: 'delete-collection' } %>
```
charleshuang80: would it be better to call this `delete-variable` for the modal and update `change_current_provider.coffee`, or possibly abstract it to `delete-record`?

Also, I think that the class would make the modal say 'Deleting this collection'.

app/controllers/variables_controller.rb
```diff
@@ -42,8 +42,22 @@ def create
...
   end
 
-  private
+  def destroy
+    delete = cmr_client.delete_variable(@provider_id, @native_id, token)
+    if delete.success?
+      flash[:success] = I18n.t('controllers.variables.destroy.flash.success')
+      Rails.logger.info("Audit Log: Variable with native_id #{@native_id} was deleted for #{@provider_id} by #{session[:urs_uid]}")
+
+      redirect_to manage_metadata_path
```
charleshuang80: I think this will need to change to `manage_collections_path`

app/controllers/variables_controller.rb
```diff
@@ -22,6 +22,13 @@ def edit
     end
   end
 
+  def edit
```
charleshuang80: If you expand or view the whole file, it looks like this PR is now inserting a second `edit` method

macrouch: Yeah, the merge from master didn't go well, I am still working on it

spec/factories/valid_variable_draft_factories.rb
```diff
@@ -53,8 +53,8 @@
     draft_type 'VariableDraft'
 
     draft {{
-      'Name': draft_short_name || "#{Faker::Space.galaxy}_#{Faker::Number.number(6)}",
-      'LongName': draft_entry_title || "#{Faker::Space.nebula} #{Faker::Space.star_cluster} #{Faker::Number.number(6)}",
+      'Name': draft_short_name || 'PNs_LIF',
```
abbottry: I think we want to leave the old way of doing this (generating random stuff) given the uniqueness constraint.

PR #98, opened 2017-08-03
MMT-953 Related Urls to Related URLs

macrouch:


spec/features/collection_drafts/form_accordions_spec.rb
```diff
@@ -148,7 +148,7 @@
 
   context 'when collapsing all accordions manually' do
     before do
-      click_on 'Related Urls'
+      click_on 'Related Urls', match: :first
```
charleshuang80: This one didn't need to be changed? (sorry, I'm just confused about which ones are changing or not)

macrouch: I don't know. It shouldn't be "Urls" though

PR #99, opened 2017-08-03
MMT-1048 add missing Map related url subtypes

macrouch:


macrouch:
Tests are failing on this branch because it is picking up the cmr_metadata_preview changes to Related URLs from MMT-953. We have to merge #98 before these tests will pass :(

PR #100, opened 2017-08-03
MMT-1032: Added new search bar to the top right with radio buttons and provider dropdown. 

rdcody:


abbottry:
Looks like there may be a z-index issue with the search box, im unable to click on any of the enabled elements on the search popup @rdcody

rdcody:
@abbottry This should be okay now if you pull the latest.

abbottry:
@rdcody 

1. Is this removing quick/full search and combining into one?
2. Seems to be quite a few tests deleted.. is this due to something in #1?
3. I can't make the drop downs (provider or search) collapse, seems as though when I take certain actions it would, like clicking away from them?

rdcody:
@abbottry 
1. That is correct. The the keyword text area always uses the full search now. As there is always a default scope set (collections, variables, or services).
2. Yes the test removed are from the full search modal window and also draft search that no longer exists. 
3. I will add another click action that collapses the menu if anywhere on the screen is clicked. 
----
The currently failing 6 tests are something I have not been able to figure out with sorting. If you have a moment could you take a look?

PR #109, opened 2017-08-14
MMT-1079: Removes custom pluralization of UMM Var keys

abbottry:


macrouch:
There are some test failures on this branch that might be real

PR #112, opened 2017-08-15
MMT 974 provide select field for platform shortname

macrouch:


app/controllers/collection_drafts_controller.rb
```diff
@@ -475,7 +476,30 @@ def set_projects
...
-    @platform_types = cmr_client.get_controlled_keywords('platforms')['category'].map { |category| category['value'] }.sort
+    @platform_types = cmr_client.get_controlled_keywords('platforms')['category'].map do |category|
+      short_names = category.fetch('short_name', []).map do |short_name|
+        {
+          short_name: short_name['value'],
+          long_name: short_name.fetch('long_name', [{}]).first['value']
+        }
+      end
+
+      series_entities = category.fetch('series_entity', []).map do |se|
```
charleshuang80: sorry to be nitpicky - I thought we wanted slightly more descriptive argument names

app/helpers/form_helper.rb
```diff
@@ -63,28 +63,34 @@ def mmt_select(options)
...
-      values = options[:value] || []
-      values.each do |value|
-        if value && invalid_select_option(select_options, value)
-          # handle invalid options for multi_select
-          select_options.unshift value
-          disabled_options << value
-        end
-      end
+    if options[:grouped]
+      select_options = grouped_options_for_select(select_options, options[:value])
```
macrouch: @abbottry what is your opinion?

The code below this adds existing values as disabled options in the select. This is in case the value previously entered isn't part of the controlled keywords. I'm not sure how we would handle this with the options being grouped by Type.

Right now if the existing value isn't found in the short name's it will just display 'Select Short Name'

spec/features/collection_drafts/forms/invalid_picklists_spec.rb
```diff
@@ -416,34 +417,6 @@
     end
   end
 
-  context 'when viewing the Platform Type field' do
```
charleshuang80: Since you made the functionality for invalid picklist values to work for platforms, do should we keep the tests for it, or is it not necessary?

PR #113, opened 2017-08-16
MMT-1005: adds search for variables by keyword

charleshuang80:
- switching between pages or choosing different record types change Search button text and will initiate different searches
- adds searching for variables, using the `.umm_json` search endpoint extension
- change methods and views to accommodate searching for collections and variables
- add and refactor some tests


charleshuang80:
As of now, still waiting on a fix for when `reset_provider` is hit in our tests, and how that affects the variable search endpoint and for SIT to be back up.

app/views/search/_variable_search_results.html.erb
```diff
@@ -0,0 +1,54 @@
...
+        <td colspan="4">
+          No variables found.
+        </td>
+      </tr>
+    <% else %>
+
+      <% variables.each do |variable| %>
+        <tr class='<%= cycle("alt", "") %>'>
+          <td>
+            <%= link_to variable['umm']['Name'], variable_path(variable['meta']['concept-id']) %>
```
abbottry: I'm guessing this is how collections render this, but nested keys here could result in an error if 'umm' or 'meta' are missing -- `fetch` would be nice here.

app/views/search/_variable_search_results.html.erb
```diff
@@ -0,0 +1,54 @@
...
+          </td>
+          <td>
+            <%= variable['umm']['LongName'] %>
+          </td>
+          <td>
+            <%= variable['meta']['provider-id'] %>
+          </td>
+          <td>
+            <% fragments = variable['meta']['revision-date'].split('T')
+               date = fragments[0].nil? ? 'UNKNOWN' : fragments[0] %>
```
abbottry: Is `UNKNOWN` consistent and/or was this discussed as the value to use when no revision date is provided?

abbottry: Also, bad news setting a variable IN the view. Great spot for a view helper.

spec/support/search_matchers.rb
```diff
@@ -1,16 +1,30 @@
 module SearchUtil
   # Usage
-  # expect(page).to have_search_query(1, "Short Name: #{short_name}")
-  # expect(page).to have_search_query(1, "Short Name: #{short_name}", "Record State: Published Records")
+  # expect(page).to have_collection_search_query(1, "Short Name: #{short_name}")
```
abbottry: No longer used?

app/views/search/index.html.erb
```diff
@@ -1,11 +1,28 @@
 
 <div class="row row-content">
   <section>
-    <div id="collection_search_criteria">
-      <%= render partial: 'collection_search_criteria' %>
-    </div>
-    <div id="collection_search_results">
-      <%= render partial: 'collection_search_results' %>
+    <div id="search-criteria">
+      <%= render partial: 'search_criteria' %>
```
abbottry: When using partials we should pass variables in using `locals` rather than expecting instance variables to exist. (I know this is how collections does it so 1. Update both if its not a massive piece of work **or** 2. improve this and we'll continue to work this type of thing (collections) in the future as long as it pertains to a ticket)

app/assets/javascripts/search.coffee
```diff
@@ -21,47 +21,56 @@ $(document).ready ->
...
-  $('#record_state_published_records').on 'click', ->
-    $("#search-text").text('Search Collections')
 
-  $('#record_state_variables').on 'click', ->
-    $("#search-text").html('<span style="padding-right:11px;">Search Variables</span>')
+  if $('#search-submit-button').length > 0
+    recordType = $('input[name="record_type"]:checked').val()
+    $searchButtonSpan = $('#search-submit-button-text')
+
+    switch recordType
```
abbottry: This can be rewritten to check for variables, and `else` use collections instead of having its own case and being the default

app/controllers/search_controller.rb
```diff
@@ -9,109 +9,48 @@ def index
...
-      # If search came from full search, ignore whatever was in search
-      @query = params.clone
-      @query['keyword'] ||= ''
-      @query.delete('provider_id') if @query['provider_id'].blank?
-    end
+    @record_type = params['record_type']
 
+    # set query
+    @query = {}
+    @query['keyword'] = params['keyword'] || ''
```
abbottry: `@query['keyword'] ||= ''` is a more concise way of writing this (looks like it was removed)

app/controllers/search_controller.rb
```diff
@@ -9,109 +9,48 @@ def index
...
+    @query['provider_id'] = params['provider_id'] unless params['provider_id'].blank?
+    @query['sort_key'] = params['sort_key'] if params['sort_key']
```
abbottry: `provider_id` and `sort_key` seem like the same check but they are done with different code. Would increase readability if they check for these was the same, if possible.

app/views/search/_collection_search_results.html.erb
```diff
@@ -1,40 +1,36 @@
 <div class="pagination-summary">
-  <%= page_entries_info @collections, entry_name: 'record' %>
+  <%= page_entries_info collections, entry_name: 'record' %>
```
abbottry: Seems like we would want this to say 'No collections found' instead of 'No records found'.

app/views/search/_variable_search_results.html.erb
```diff
@@ -0,0 +1,54 @@
+<div class="pagination-summary">
+  <%= page_entries_info variables, entry_name: 'record' %>
```
abbottry: Seems like we would want this to say 'No variables found' instead of 'No records found'.

app/views/search/_search_criteria.html.erb
```diff
@@ -0,0 +1,6 @@
+<p class="search-meta">
```
abbottry: See related comment related to sending instance variables.

app/assets/javascripts/search.coffee
charleshuang80: My thinking was that since we will likely need to add a `services` option, that using `case/switch` might be slightly preferable. Would you rather I changed it to what you suggested?

app/controllers/search_controller.rb
charleshuang80: That worked better when it had  `@query = params.clone` before it. Since this is not cloning params, `@query['keyword']` would always end up as `''` (that's how I had it before I ran into issues)

app/controllers/search_controller.rb
charleshuang80: I'll update these with the same check

app/views/search/_variable_search_results.html.erb
charleshuang80: I kept this as it had been implemented for collections, but I will implement a view helper for these.

app/views/search/index.html.erb
charleshuang80: I was already doing this with the two partials below, but since it was such a small partial I wasn't sure if the same principle should apply. Apparently yes, so I will make the change.

app/assets/javascripts/search.coffee
abbottry: You can still use the case/switch statement, I'm just suggesting putting the default value as the `else` statement (default). When services come it'll ready variables, then services, then else/default. Just removing the redundancy not changing the method of checking.

app/controllers/search_controller.rb
abbottry: Got it 👌

app/assets/javascripts/search.coffee
charleshuang80: Understood. The change has been made.

PR #118, opened 2017-08-18
MMT-1057 Fix breadcrumbs for collection and variable drafts and records

macrouch:


app/helpers/manage_metadata_helper.rb
```diff
@@ -1,12 +1,15 @@
 # :nodoc:
 module ManageMetadataHelper
   def display_entry_id(metadata, type)
-    blank_short_name = type == 'draft' ? '<Blank Short Name>' : 'New Collection'
-    short_name = metadata['ShortName'] || blank_short_name
+    short_name = if type.include? 'collection'
```
abbottry: sneaky -- `include` instead of `replace`. 🎩 (may want to downcase it, unless we're sure its always lower case as `include` is case sensitive)

macrouch: Right now it is either coming from `resource_name` or `'collections'`. I've already changed this method some for MMT-1117, I'll include downcase in that ticket

spec/features/collections/breadcrumbs_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper'
+
+describe 'Collections breadcrumbs' do
+  before do
+    login
+    ingest_response, @concept_response = publish_collection_draft
```
charleshuang80: not necessary to make this change, but would it reduce a bit of test time to visit one of the LARC or SEDAC collections so we don't have to ingest one for this?

PR #119, opened 2017-08-18
MMT-1117 display header title information for variables

macrouch:


app/views/drafts/show.html.erb
```diff
@@ -1,3 +1,8 @@
+<% content_for :header_title do %>
+  <h2><%= display_entry_id(get_resource.draft, resource_name) %></h2>
+  <p class="subtitle"><%= display_variable_long_name(get_resource.draft) %></p>
```
charleshuang80: This feels a little odd because it this is the generic draft show view, but you're using a method specific to variables. I'm not sure if that means we need to change it since collections has its own show view, but just wanted to note that it seems odd.

lib/assets/schemas/umm-var-form.json
```diff
@@ -98,7 +98,7 @@
       ]
     },
     {
-      "id": "fill_value",
+      "id": "fill_values",
```
abbottry: I would expect this to break a fair number of tests.

app/helpers/manage_metadata_helper.rb
```diff
@@ -1,6 +1,6 @@
 # :nodoc:
 module ManageMetadataHelper
-  def display_entry_id(metadata, type)
+  def breadcrumb_name(metadata, type)
```
abbottry: No `downcase` 😢

abbottry: I like the name change, its much clearer what the gem is doing and what MMT is doing.

lib/assets/schemas/umm-var-form.json
macrouch: Yep, it did. I noticed the title on the edit form didn't match once I added the form titles. Sets and Services as well

app/helpers/manage_metadata_helper.rb
macrouch: doh! I forgot

PR #120, opened 2017-08-18
MMT-1095: underlines the proper 'Manage _' header option for collections and variables, including with search and bulk updates

charleshuang80:


app/views/layouts/manage_metadata.html.erb
```diff
@@ -33,13 +33,14 @@
 
               <nav>
```
macrouch: Why don't we take this `<nav>` and make it a partial so the different layouts stop duplicating code.

PR #126, opened 2017-08-25
MMT-1008: Adds the ability to create variable collection associations

abbottry:


lib/tasks/drafts.rake
```diff
@@ -21,6 +21,24 @@ namespace :drafts do
     end
   end
 
+  desc 'Load full draft into database'
+  task load_full_variable: :environment do
```
macrouch: `load_full_variable` is already in this file

abbottry: Hm. Look at that.

abbottry: Fixed that up -- good catch (the one i added was broke anyway 🤦‍♂️ )

app/controllers/collection_associations_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def new
+    add_breadcrumb 'New', new_variable_collection_association_path(params[:variable_id])
+
+    super
+
+    association_response = cmr_client.get_variables({ concept_id: params[:variable_id] }, token)
+
+    @previously_associated_collections = association_response.body.fetch('items', []).first.fetch('associations', {}).fetch('collections', []).map { |collection| collection['concept-id'] }
```
charleshuang80: You're parsing the association concept ids multiple times - is it worth making it into a method?

PR #128, opened 2017-08-25
MMT-976 add select2 field for instrument short name and populate long…

macrouch:
… name

### DIFF
--- a/app/assets/javascripts/select2.coffee
+++ b/app/assets/javascripts/select2.coffee
@@ -148,3 +148,31 @@ $(document).ready ->
         $longNameElement.attr('readonly', true)
       else
         $longNameElement.attr('readonly', false)
+
+  # when selecting short name, populate long name and set to readonly
+  $('.instrument-short-name-select').on 'select2:select', (event) ->
+    $element = $(this)
+    longName = $element.find(':selected').data('longName')
+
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/select2.coffee
```diff
@@ -148,3 +148,31 @@ $(document).ready ->
...
+    longName = $element.find(':selected').data('longName')
+
+    $longNameElement = $element.parent().siblings().find('.instrument-long-name')
+
+    if longName?
+      $longNameElement.val(longName)
+      $longNameElement.attr('readonly', true)
+    else
+      $longNameElement.val('')
+      $longNameElement.attr('readonly', false)
```
charleshuang80: So we have MMT-1116, that a user shouldn't be able to key in a longname for KMS values. Should that apply to instruments as well?

macrouch: Yes it should, but this ticket it written this way. I like the idea of taking care of all of the long name fields at the same time anyway

charleshuang80: Taking care of it all in 1116 seems fine to me, but it looks like there are comments in the ticket that seem contrary to how long name behavior is written in the acceptance criteria, just fyi.

charleshuang80:
Should the Instrument Child fields be covered by this?

PR #132, opened 2017-08-29
MMT-1084 fix problem in Safari with validation running on newly creat…

macrouch:
…ed fields


macrouch:
@charleshuang80  Please test for the problems you were seeing again, I can't remember exactly what they were but I think they have been resolved

PR #135, opened 2017-08-30
MMT-1009: Adds ability to delete variable collection associations

abbottry:


app/controllers/collection_associations_controller.rb
```diff
@@ -57,4 +79,18 @@ def add_high_level_breadcrumbs
 
     add_breadcrumb 'Collection Associations', variable_collection_associations_path(params[:variable_id])
   end
+
+  def log_issues(response)
+    response.body.select { |assocation_response| assocation_response.fetch('warnings', []).any? }.each do |warnings|
+      warnings.each do |warning|
+        Rails.logger.error "Variable Assocation [warning]: #{warning}"
```
charleshuang80: I don't know what the warnings or errors look like, but does this provide sufficient information (i.e. the variable and/or collection concept ids)?

spec/support/draft_helpers.rb
```diff
@@ -101,6 +101,16 @@ def publish_variable_draft(provider_id: 'MMT_2', native_id: nil, name: nil, long
       end
     end
 
+    def add_variable_collection_associations(variable_id, collection_concept_ids)
```
charleshuang80: I added a similar helper for this in MMT-1124 (and it just got merged). I'm not sure if this PR is the right place to consolidate the two or if we should do that in a future PR.

app/controllers/collection_associations_controller.rb
abbottry: Yes, it does.

spec/support/draft_helpers.rb
abbottry: Sure, I'll take a look

PR #136, opened 2017-08-30
MMT-1103: fix sorting variable search results test by provider

charleshuang80:


spec/features/search/variable_results_sorting_spec.rb
```diff
@@ -99,8 +99,10 @@
       end
 
       it 'sorts the result by Name Asc' do
-        within '#search-results tbody tr:nth-child(1)' do
-          expect(page).to have_content('Larc Var')
+        within '#search-results tbody tr:nth-child(1) td:nth-child(3)' do
```
abbottry: But you're sorting by name. I think we need a more reliable way to determine sorting is working.

charleshuang80: oops. No this is sorting by provider (the before block has `click_on 'Sort by Provider Asc'`). I must have copied parts of this test to start it - the `it` description is wrong here. I'll update the text, but do you have another suggestion for determining sorting?

PR #137, opened 2017-08-31
MMT-976 fix bug in parsing short names

macrouch:


macrouch:
You can test locally with SIT data by changing line 110 in `cmr_client.rb` to be `get("/search/keywords/#{type}").body`

PR #138, opened 2017-08-31
MMT-1125: Adds `Expand All` link to Variable Draft forms

abbottry:


macrouch:
By default the link says 'Expand All', but on empty forms, all accordions are already expanded. If they are all expanded it should probably say 'Collapse All'

abbottry:
@macrouch I changed it so that only the last one is expanded to match collections, as part of this PR.

macrouch:
@abbottry That makes sense, but on an empty form or just one accordion, it defaults to Expanded, but the link says 'Expand All'. You have to click it twice to get it to do anything

macrouch:
Nevermind, this behaves the same as the Collection forms

charleshuang80:
If I click on 'Expand All', the link changes to 'Collapse All', but then when I click the accordion headers to close all the accordions on the page, the link does not change to 'Expand All'.

PR #140, opened 2017-09-05
MMT-1008: Fixes paging issues on Variable Collection Associations

abbottry:


app/views/cmr_search/shared/_search_results.html.erb
```diff
@@ -35,6 +35,10 @@
           <% end %>
         </tbody>
       </table>
+
+      <% if (collections.map { |c| c.fetch('meta', {})['concept-id'] } - (ignored_concept_ids rescue [])).empty?  %>
```
macrouch: This feels like a lot of logic in a view, maybe make a helper for it?

PR #146, opened 2017-09-15
MMT-1127: adds ensuring correct provider for variable manage collection associations

charleshuang80:
* add methods to ensure correct provider for variable manage collection associations actions
* refactor ensure correct provider methods for variables and collections
* convert ensuring correct provider for drafts to use Pundit policies

### DIFF
--- a/app/policies/draft_policy.rb
+++ b/app/policies/draft_policy.rb
@@ -2,7 +2,6 @@
 class DraftPolicy < ApplicationPolicy
   def show?
     user.user.provider_id == target.provider_id
-    # user.available_providers.include?(target.provider_id)
   end
 
   def new?
--- a/app/policies/draft_policy.rb
+++ b/app/policies/draft_policy.rb
@@ -1,0 +1,35 @@
+# :nodoc:
+class DraftPolicy < ApplicationPolicy
+  def show?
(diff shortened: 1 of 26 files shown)


app/policies/draft_policy.rb
```diff
@@ -0,0 +1,35 @@
+# :nodoc:
+class DraftPolicy < ApplicationPolicy
+  def show?
+    user.user.provider_id == target.provider_id
```
macrouch: Why do these policies use `user.user.provider_id`?

app/policies/draft_policy.rb
```diff
@@ -0,0 +1,35 @@
...
+    # user.available_providers.include?(target.provider_id)
```
macrouch: commented out code

charleshuang80: I'll remove this.

app/policies/draft_policy.rb
charleshuang80: We add additional user context to pass in the token (for the ACL checks), in `def pundit_user` in `ApplicationController`. That adds another level in the object, so we need to use `user.user` to access the user and `user.token` to access the token.

PR #147, opened 2017-09-19
MMT-1153: add cloning of published Variable to create a Variable Draft

charleshuang80:
* add controller, model, and view changes for functionality to clone a Variable to create a Variable Draft, including tests
* add ensuring correct provider and changing provider functionality for cloning Varibles, including tests
* update action links css
* update action links wording for Collections and Collection Drafts, and tests that use the links

### DIFF
--- a/app/assets/stylesheets/components/_action.scss
+++ b/app/assets/stylesheets/components/_action.scss
@@ -7,7 +7,7 @@
   padding-top: 0.9375em;
   padding-bottom: 0.625em;
 
-  a:first-child {
+  a.eui-btn--blue:first-child {
     margin-right: 0.625em;
   }
 }
--- a/app/assets/stylesheets/components/_action.scss
+++ b/app/assets/stylesheets/components/_action.scss
@@ -7,17 +7,8 @@
   padding-top: 0.9375em;
   padding-bottom: 0.625em;
 
-  a.eui-btn--blue:first-child {
+  a[class^="eui-btn"]:not(.eui-btn--link):first-child {
     margin-right: 0.625em;
-  }
-}
-
-.cta {
-  width: 60%;
-  float: left;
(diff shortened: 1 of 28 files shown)


app/assets/stylesheets/components/_action.scss
```diff
@@ -7,7 +7,7 @@
   padding-top: 0.9375em;
   padding-bottom: 0.625em;
 
-  a:first-child {
+  a.eui-btn--blue:first-child {
```
abbottry: I'd rather be less specific here and make sure we catch all buttons. `a[class^="eui-btn"]:first-child`

mreese84: +1 to this. If you're wondering what the heck that selector does... check this out number 13 at this link: https://code.tutsplus.com/tutorials/the-30-css-selectors-you-must-memorize--net-16048

I don't think you need to memorize all 30 of those selectors, but I reference this list all the time. It's gold.

app/views/collections/show.html.erb
```diff
@@ -57,55 +57,53 @@
       </div>
     <% else %>
       <section class="action">
-        <div class="cta">
```
charleshuang80: these changes seem much bigger than they actually are, it appears that way because I removed the `<div class="cta">` and had to outdent everything within it

app/assets/stylesheets/components/_action.scss
```diff
@@ -7,20 +7,11 @@
   padding-top: 0.9375em;
   padding-bottom: 0.625em;
 
-  a.eui-btn--blue:first-child {
+  a[class^="eui-btn"]:not(.eui-btn--link):first-child {
```
mreese84: Look at that selector!

PR #148, opened 2017-09-19
MMT-1150 view revisions page for variables

macrouch:


app/controllers/manage_metadata_controller.rb
```diff
@@ -24,25 +24,37 @@ def set_variable_information
...
-                        variables_search_response.body['items'].first
+                        variables_search_response.body.fetch('items', [])
                       else
-                        {}
+                        [{}]
                       end
+      variable_data.sort! { |a, b| b['meta']['revision-id'] <=> a['meta']['revision-id'] }
 
-      break if !variable_data.nil? && variable_data.fetch('meta', {})['concept-id'] == @concept_id
+      # break if revisions is wrong
```
charleshuang80: is this a comment left in or something you need to implement?

macrouch: left in and now in a weird spot after stuff got moved around lol

app/controllers/manage_metadata_controller.rb
```diff
@@ -24,25 +24,37 @@ def set_variable_information
...
+
+      break if latest && !@revision_id
+      break if latest && latest['meta']['revision-id'] >= @revision_id.to_i && latest.fetch('meta', {})['concept-id'] == @concept_id
       attempts += 1
       sleep 0.05
     end
 
-    if variable_data.blank?
+    @revisions = variable_data
+    first_variable = variable_data.first
```
charleshuang80: is there a difference between `latest` and `first_variable`? if not, would it be better to just use one of them?

macrouch: no, they are the same, I'll change it

PR #149, opened 2017-09-21
MMT-1174: add ability to download json of variable metadata

charleshuang80:


app/controllers/variables_controller.rb
```diff
@@ -80,6 +80,10 @@ def destroy
     end
   end
 
+  def download_json
+    send_data @variable, type: 'application/json; charset=utf-8', disposition: "attachment; filename=#{@concept_id}.json"
```
abbottry: The data downloaded is not in JSON format, its a Ruby hash object, you think this can be fixed by just adding `to_json` to your `@variable` here.

app/views/variables/show.html.erb
```diff
@@ -43,6 +43,8 @@
           <%= link_to 'Clone Variable Record', '#not-current-provider-modal', class: 'display-modal not-current-provider eui-btn--link', data: { 'provider': @provider_id, record_action: 'clone-variable' } %> |
         <% end %>
 
+        <%= link_to 'Download JSON', download_json_variable_path(@concept_id), class: 'download-json', id: 'download-var-umm-json', target: '_blank' %> |
```
macrouch: Are class and id needed? I don't see them being used

PR #154, opened 2017-09-27
MMT-1151 display banner if open draft exists for variable

macrouch:


spec/features/variables/variable_with_draft_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+      end
+
+      context 'when clicking the link' do
+        before do
+          click_on 'Click here to view it.'
+        end
+
+        it 'displays the draft' do
+          within '.eui-breadcrumbs' do
+            expect(page).to have_content('Drafts')
```
charleshuang80: not totally necessary, but you could make this more specific and check for 'Variable Drafts'

spec/features/variables/variable_with_draft_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+            wait_for_ajax
+          end
+
+          it 'switches the provider context' do
+            expect(User.first.provider_id).to eq('MMT_2')
+          end
+
+          it 'displays the draft' do
+            within '.eui-breadcrumbs' do
+              expect(page).to have_content('Drafts')
```
charleshuang80: same comment as above

app/views/shared/_not_current_provider_modal.html.erb
```diff
@@ -22,8 +22,13 @@
     <% end %>
 
     <% if options[:draft] %>
-      <%= link_to 'Yes', collection_draft_path(options[:draft_id]), class: 'eui-btn--blue spinner is-invisible', id: 'not-current-provider-view-draft-link' %>
-      <%= link_to 'Yes', edit_collection_draft_path(options[:draft_id], form: options[:draft_form]), class: 'eui-btn--blue spinner is-invisible', id: 'not-current-provider-edit-draft-link' %>
+      <% if options[:variable] %>
+        <%= link_to 'Yes', variable_draft_path(options[:draft_id]), class: 'eui-btn--blue spinner is-invisible', id: 'not-current-provider-view-draft-link' %>
```
charleshuang80: The ids for the links to `view` and `edit` the variable drafts are identical to the ids for links for the collection drafts here. They should be unique, right?

macrouch: Only one set of IDs is rendered onto the page, so they are unique in the browser

charleshuang80: ah, ok.

PR #163, opened 2017-10-10
MMT-822: update collection download xml links to download from controller action

charleshuang80:


app/controllers/collections_controller.rb
```diff
@@ -3,6 +3,7 @@ class CollectionsController < ManageCollectionsController
 
   before_action :set_collection
   before_action :ensure_correct_collection_provider, only: [:edit, :clone, :revert, :destroy]
+  before_action :set_collection_download, only: [:download_xml]
```
abbottry: This piece of code will only be used in this one action. I don't think it needs to be a separate method, I would think we could just move the work that is done in this method into the `download_xml` method.

app/controllers/collections_controller.rb
```diff
@@ -65,6 +66,10 @@ def revert
     end
   end
 
+  def download_xml
+    send_data @collection_download, type: 'text/xml', disposition: "attachment; filename=#{@concept_id}.#{@collection_format}"
```
abbottry: I feel like we should be setting the `type` to an appropriate value here, potentially listed in the `DOWNLOAD_XML_OPTIONS` or something. If `.native` is umm-json we'd be telling the file its xml when its really json.

app/controllers/collections_controller.rb
```diff
@@ -138,17 +142,21 @@ def get_revisions(concept_id, revision_id)
...
+
+    collection_response = cmr_client.get_collection_concept_for_download(@concept_id, @collection_format, token, revision_id)
+
+    @collection_download = if collection_response.success?
+                             collection_response.body
+                           else
+                             Rails.logger.error("Error retrieving concept for Collection #{@concept_id}, revision #{revision_id} in #{@collection_format} format for user #{current_user.urs_uid}")
+
+                             error_message = "There was an error retrieving the collection metadata for Collection #{@concept_id}, revision #{revision_id} in #{@collection_format} format that you requested."
+                             "<?xml version='1.0' encoding='UTF-8'?><error>#{error_message}</error>"
```
abbottry: I'm 50/50 on supplying the error in the file. 

1. If the file isn't xml (when the native format is umm-json for instance) this would be odd.
2. Is there a better way to display this? A redirect or something? We may want to punt on trying to show the error given that

    - this wasn't something we did before this change
    - variable json download doesnt do this
    - we have an upcoming ticket to display appropriate error pages
    - it seems unlikely to throw an error given that were looking at the show page when this link is displayed.

lib/cmr/cmr_client.rb
```diff
@@ -115,6 +115,17 @@ def get_granule_count(collection_id, token)
       get(url, options, token_header(token))
     end
 
+    # This method is to retrieve the collection for xml download
+    def get_collection_concept_for_download(concept_id, collection_format, token, revision_id = nil)
```
abbottry: What is different between this and the `get_concept` method? Seems like we can just allow the ability to pass query params (to support pretty=true)

app/controllers/collections_controller.rb
charleshuang80: I will remove the before action. I think the method should be best kept separate (based on keeping methods smaller if possible), but I don't have to.

app/controllers/collections_controller.rb
charleshuang80: this sounds like a good idea. I should have, but overlooked, thinking about if `.native` is umm_json.

lib/cmr/cmr_client.rb
charleshuang80: There is not much difference aside from the extension. At the time I thought it might be helpful to keep it separate, but it does seem unnecessary.

app/controllers/collections_controller.rb
charleshuang80: 1. I agree but maybe this part could be handled 

1. I can certainly not provide the error in the file, but I'm not sure there's a better way to display it:
  - `flash.now` didn't work when I tried so I'm not sure how any flash type message might work
  - A redirect won't work because the link isn't taking the user anywhere it is initiating a download with an empty tab opening at most

I agree that the show page was being viewed so an error is unlikely, so I can remove the error in the file here. But -- if there is an error should we provide an empty file or not allow a download?

PR #172, opened 2017-10-24
MMT-1205 put each location/science keyword on new line

macrouch:


macrouch:
This will have a Gemfile update before being merged

PR #173, opened 2017-10-25
MMT-1200 add provider_id validation to drafts

macrouch:


db/migrate/20171025180642_remove_drafts_with_missing_provider.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveDraftsWithMissingProvider < ActiveRecord::Migration
+  def change
+    Draft.all.each do |draft|
+      draft.destroy if draft.provider_id.nil?
```
abbottry: In the interest of best practices this type of things is best written using where clauses --
 this code loads all the drafts in the database which is easily avoidable using `Drafts.where(provider_id: nil).each`. While minor, I'd suggest making this change and considering these types of efficiencies in the future.

PR #174, opened 2017-10-25
MMT-1088 remove feedback link from footer

macrouch:


app/views/shared/_footer.html.erb
```diff
@@ -5,7 +5,6 @@
       <li><a href="http://www.nasa.gov/FOIA/index.html">FOIA</a></li>
       <li><a href="http://www.nasa.gov/about/highlights/HP_Privacy.html">NASA Privacy Policy</a></li>
       <li><a href="http://www.usa.gov/">USA.gov</a></li>
-      <li><a href="javascript:feedback.showForm();">Feedback</a></li>
     </ul>
 </footer>
```
charleshuang80: I'm asking the Earthdata team, but shouldn't we need to remove any js assets associated with the link? I'm thinking the script tags below in the footer all have to do with what this link is triggering

charleshuang80: James verified that the script tags below are all related to the feedback module, so I think they should be removed.

PR #177, opened 2017-10-26
MMT-1134 make instrument child shortname longname controlled vocabulary

macrouch:


app/views/collection_drafts/forms/fields/_instrument_child_fields.html.erb
```diff
@@ -2,15 +2,17 @@
...
+        title: 'Short Name',
         prefix: name_prefix,
         value: instrument_child['ShortName'],
+        options: @instruments.map { |instrument| [instrument[:short_name], instrument[:short_name], { 'data-long-name' => instrument[:long_name] }] },
         help: 'definitions/PlatformShortNameType',
+        classes: 'select2-select instrument-short-name-select',
         validate: true,
-        required: true
-       ) %>
+        always_required: true
```
charleshuang80: We just tested with this field reverted to `required: true`, and in the instrument fields as well. That seemed closer to the expected behavior, but the required icon for the instrument and instrument child short name does not always go away when expected, sometimes it needed an additional blur off of the long name as well.

charleshuang80:
sorry, still one more validation thing -- I'm still seeing validation errors on the first `instrument child` inputs when I enter things on the second. this is happening (I think) because the `data-level` attribute doesn't match the id for the added fields

screen shot 2017-11-01 at 1 33 20 pm

PR #180, opened 2017-11-07
MMT-1133 redirect users to provider context page if no provider_id is…

macrouch:
… set


db/migrate/20171106140232_remove_select_provider_provider_id.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveSelectProviderProviderId < ActiveRecord::Migration
+  def up
+    User.where(provider_id: 'Select Provider').all.each do |user|
```
charleshuang80: I don't know if this is as big a deal, but it seems possible that 'Select Provider' somehow got saved into a user's `available_providers` as well, should we be removing that too?

PR #190, opened 2017-11-17
MMT-1222 update available providers on login

macrouch:


charleshuang80:
If I am logged in with a provider context (i.e. 'MMT_1'), and I remove that provider from my available providers, and click on the provider context badge and then 'Refresh your available providers' in the modal, the provider is removed from the list (in development I have to change the providers that get set in `set_available_providers`), but I stay logged in with the provider that is no longer in the provider dropdown list. If a user's current provider is removed the user should probably be redirected to the provider context landing page to choose a different provider.

PR #195, opened 2017-11-27
MMT-1053 add anyOf error messages for Geometry

macrouch:


app/assets/javascripts/validation.coffee
```diff
@@ -168,12 +168,22 @@ $(document).ready ->
...
     if error.keyword == 'anyOf'
-      # anyOf errors are showing up in the data contacts form, but only when
-      # there are other validation errors. as the error messages are duplicate
-      # and don't have the specificity of the other error messages (or have
-      # information useful to the user), it seems best to suppress these
-      # more info at https://github.com/epoberezkin/ajv/issues/201#issuecomment-222544956
+      if error.dataPath.indexOf('Geometry') != -1
+        error.message = 'At least one Geometry Type is required'
+      else
+        # anyOf errors are showing up in the data contacts form, but only when
```
charleshuang80: this was my concern, if the data contacts form would be handled properly as well. both seem to work as we want now.

PR #196, opened 2017-11-30
MMT-1224: force refresh cache of providers list when a user logs in or refreshes their available providers

charleshuang80:


spec/features/providers_list_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+        login
+
+        visit groups_path
+      end
+
+      context 'when then adding the provider', js: true do
+        new_provider = "#{Faker::Ancient.primordial.slice(0, 5).upcase}_#{Faker::Number.number(3)}"
+
+        before(:each, create_provider: true) do
+          create_provider(new_provider)
```
macrouch: Look at removing the newly created providers after the test is over

PR #198, opened 2017-12-04
MMT-1221 show new providers when refreshing available providers with …

macrouch:
…only one provider


spec/features/provider_context/provider_context_spec.rb
```diff
@@ -179,19 +179,16 @@
 
     context 'when a user refreshes their available providers' do
```
charleshuang80: I think this context statement is slightly ambiguous, we should add they the user gains a provider to have more than one. I think a few of the context statements above could also be more clear/specific (i.e. on line 53), but we may want to improve them in a future ticket.

PR #206, opened 2017-12-19
MMT-1250: handle html responses for cmr client requests

macrouch:


charleshuang80:
I really like how the `cmr_error_messages` and the `body_is_html?` methods were pushed to the `response.rb` class. I'm not arguing against it, but what was the decision process to not change the response body if it is html?

I think that not changing the body will still allow for some 500 errors to occur. In the places we check a response for `success?` or `error?` there should not be an issue, but where we don't (i.e. in `controlled_keywords.rb` could throw an error so we should then probably change those methods to check responses for `success?` or `error?`.

app/concerns/controlled_keywords.rb
```diff
@@ -53,9 +53,9 @@ def set_platform_types
   end
 
   def set_instruments
-    @instruments = get_controlled_keyword_short_names(cmr_client.get_controlled_keywords('instruments').fetch('category', []))
```
charleshuang80: In my other comment, I mentioned how the places we don't check for a response `success?` or `error?`, may throw an error if the body is a html document. Here, for instance, if the body is a html document and we are on the `acquisition_information` form, `set_platform_types`, `set_instruments`, and `set_projects` will all throw an error (undefined method 'fetch' for String). We should probably change all these methods to check for `success?` or `error?`

macrouch: You're right, I will keep working on it tomorrow. 

The idea was that MMT shouldn't know anything about the CMR responses. MMT shouldn't know where the errors are in the response, it just wants the response to give it some error messages. I think what we should move towards is MMT knowing as little as possible about CMR. 

For instance, when searching collections, `https://cmr.earthdata.nasa.gov/search/collections.umm_json?pretty=true`, MMT doesn't need to know that the collections are at `response.body.fetch('items', [])`. MMT should just be able to ask the response for the items, `response.items`

That is a huge change to the system, so I didn't want to get into it in this ticket, but I'll see if there is an intermediate step I can take within this ticket.

macrouch: I went back and forth on manipulating the body of the response, I don't really have a concrete reason, but I wanted to leave the original response alone as much as I could

charleshuang80: I think your reasoning makes sense, but I think that means that we should always check the cmr response for `success?` or `error?`, and make sure to set something in the error case that won't cause problems. In this particular file we're not doing that at all, so that's the change I'm suggesting here.

macrouch: In my last commit I searched for `cmr_client.` and made sure every time we are requesting data that we are checking for success or error before looking at the body

app/concerns/chooser_endpoints.rb
```diff
@@ -45,7 +45,8 @@ def get_provider_collections(params = {})
     collection_params['keyword'].concat('*') if collection_params.key?('keyword')
 
     # Retreive the collections from CMR, allowing a few additional parameters
-    cmr_client.get_collections_by_post(collection_params, token).body
+    response = cmr_client.get_collections_by_post(collection_params, token)
+    response.body if response.success?
```
charleshuang80: It seems like we should be providing a response body if the response is an `error?` as well. Here it should probably be a `{}` because the in some cases `render_collections_for_chooser` method above will be called, passing in what is returned from this method.

app/controllers/users_controller.rb
```diff
@@ -76,4 +75,18 @@ def refresh_providers
...
+
+  private
+
+  def cached_providers
+    # retrieve the cached list of all providers
+    response = cmr_client.get_providers
+    response.body if response.success?
+  end
+
+  def all_providers
```
charleshuang80: these provider methods should also have a default response if they are `error?`s, since otherwise we could be comparing the response objects, and in later calling `.map` on `all_providers`

app/helpers/service_entry_helper.rb
```diff
@@ -36,7 +36,8 @@ def collections_from_service_entry_tags(service_entry)
 
     return [] if concept_ids.blank?
 
-    cmr_client.get_collections_by_post({ concept_id: concept_ids, page_size: concept_ids.size }, token).body.fetch('items', [])
+    response = cmr_client.get_collections_by_post({ concept_id: concept_ids, page_size: concept_ids.size }, token)
+    response.body.fetch('items', []) if response.success?
```
charleshuang80: we may need a default response in the case of an `error?` here too.

PR #207, opened 2017-12-22
MMT-1271 configure Faraday to follow redirects

macrouch:


macrouch:
Still looking for a good way to test this

PR #210, opened 2018-01-11
MMT-1265 don't show deprecated order options in new order option assi…

macrouch:
…gnment dropdown


spec/features/order_option_assignments/create_order_option_assignments_spec.rb
```diff
@@ -223,24 +223,11 @@
       VCR.use_cassette('echo_rest/order_option_assignments/display', record: :none) do
```
charleshuang80: maybe the `context` text should be updated, since the test purpose has changed?

PR #213, opened 2018-01-12
MMT-1282, MMT-791: Implement custom error pages with request.uuid provided and Earthdata styling

charleshuang80:


charleshuang80:
In order to check this locally, you can go directly to `/404` or `/500`, OR you can change this setting in `development.rb`:
`config.consider_all_requests_local       = false`
and then go to a url that doesn't exist, or cause some error manually.

PR #214, opened 2018-01-16
MMT-1293 MMT-1294 update gems for security reasons

charleshuang80:


jceaser:
why did lib/cmr/base_client.rb need to be changed?

PR #215, opened 2018-02-02
MMT-1017 MMT-1015 add tests for seeing service drafts lists and deleting service draft

charleshuang80:


spec/features/manage_services/open_drafts_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'rails_helper'
+
+describe 'Drafts listed on the Manage Services page', reset_provider: true do
```
macrouch: We shouldn't need `reset_provider` here, should we?

PR #216, opened 2018-02-15
MMT-1246: Bulk Update Name should be unique within a provider

charleshuang80:
- fix tests failing in Bamboo from uniqueness constraint on bulk update names being turned on
- repopulate bulk update form if a bulk update fails because the bulk update name is not unique
- add ajax validation to check bulk update name on the form
- add tests


config/routes.rb
```diff
@@ -48,6 +48,8 @@
   post '/invite_user' => 'groups#invite', as: 'invite_user'
   get '/accept_invite/:token' => 'groups#accept_invite', as: 'accept_invite'
 
+  get '/bulk_updates/check_task_name' => 'bulk_updates#check_task_name'
```
charleshuang80: remove

spec/features/bulk_updates/bulk_updates_data_centers_spec.rb
```diff
@@ -82,9 +82,10 @@
 
       context 'when viewing the collection' do
         before do
-          within '#bulk-update-status-table' do
-            click_on @find_and_remove_concept_response.body['EntryTitle']
-          end
+          # within '#bulk-update-status-table' do
```
charleshuang80: remove

spec/features/bulk_updates/bulk_updates_instruments_spec.rb
```diff
@@ -183,12 +184,13 @@
 
       context 'when viewing the collection' do
         before do
-          within '#bulk-update-status-table' do
-            click_on @find_and_update_concept_response.body['EntryTitle']
-          end
+          # within '#bulk-update-status-table' do
```
charleshuang80: remove comment

spec/features/bulk_updates/repopulate_form_from_failed_bulk_updates_spec.rb
```diff
@@ -0,0 +1,468 @@
...
+          expect(page).to have_no_content(@used_bulk_update_name)
+        end
+
+        expect(page).to have_content('Error creating bulk update task: Bulk update name needs to be unique within the provider.')
+
+        expect(page).to have_field('bulk_update_name', with: @used_bulk_update_name)
+        expect(page).to have_select('Field to Update', selected: 'Platforms')
+        expect(page).to have_select('Update Type', selected: 'Find & Update')
+        expect(page).to have_field('Short Name', with: 'A340-600')
+        # expect(page).to have_css('li.select2-selection__choice', text: 'DMSP 5B/F3')
```
charleshuang80: remove comment

app/assets/javascripts/bulk_updates.coffee
```diff
@@ -345,6 +349,13 @@ $(document).ready ->
         else
           error.insertAfter(element)
 
+
+    # on load, disable the inputs from partials that are hidden
+    $('.bulk-update-partial').each ->
+      if $(this).hasClass('is-hidden')
+        $update_field_selector = "#" + $(this)[0].id
+        hideAndClear($update_field_selector)
```
macrouch: I think I would shrink this down to one line, `hideAndClear("##{$(this)[0].id}")` (maybe that works?)

charleshuang80: Yeah, I think that'll work. I'll make the change and keep it in if it works.

app/controllers/bulk_updates_controller.rb
```diff
@@ -167,4 +190,19 @@ def ensure_correct_data_center_update_value(task)
...
+
+    reconstructed_bulk_update_object = {
+      name:           params[:name],
+      update_field:   params.fetch(:update_field, '').downcase,
+      update_type:    params[:update_type],
+      find_value:     params[:find_value],
+      update_value:   params[:update_value]
+    }
+
+    Rails.logger.info "!!!!**** !!!! **** reconstructed task #{reconstructed_bulk_update_object.inspect}"
```
macrouch: probably don't need the !!!!**** in the logs

app/helpers/bulk_updates_helper.rb
```diff
@@ -202,13 +202,62 @@ module BulkUpdatesHelper
...
+  def show_updateable_field(field, task)
+    task[:update_field] == field
+  end
+
+  # Requirements from the Bulk Updates Wiki
+  # IF type FIND_AND_REMOVE or FIND_AND_REPLACE or FIND_AND_UPDATE, Find value required
+  def show_find_value(field_to_update, task)
+    return nil unless field_to_update == task[:update_field]
+
+    task[:update_type].upcase.start_with?('FIND')
```
macrouch: could `task[:update_type]` ever be nil? `.upcase` would error out

app/helpers/bulk_updates_helper.rb
```diff
@@ -202,13 +202,62 @@ module BulkUpdatesHelper
...
+  end
+
+  # IF NOT type FIND_AND_REMOVE, New value required
+  def show_update_value(field_to_update, task)
+    return nil unless field_to_update == task[:update_field]
+
+    task[:update_type].upcase != 'FIND_AND_REMOVE'
```
macrouch: same as line #235

spec/features/bulk_updates/bulk_updates_instruments_spec.rb
```diff
@@ -78,12 +78,13 @@
 
       context 'when viewing the collection' do
         before do
-          within '#bulk-update-status-table' do
-            click_on @find_and_remove_concept_response.body['EntryTitle']
-          end
+          # within '#bulk-update-status-table' do
```
macrouch: remove

spec/features/bulk_updates/bulk_updates_location_keywords_spec.rb
```diff
@@ -83,9 +83,10 @@
 
       context 'when viewing the collection' do
         before do
-          within '#bulk-update-status-table' do
-            click_on @find_and_remove_concept_response.body['EntryTitle']
-          end
+          # within '#bulk-update-status-table' do
```
macrouch: remove

also wondering why you switched from clicking on something to using visit?

spec/features/bulk_updates/bulk_updates_location_keywords_spec.rb
```diff
@@ -170,9 +171,10 @@
 
       context 'when viewing the collection' do
         before do
-          within '#bulk-update-status-table' do
-            click_on @add_to_existing_concept_response.body['EntryTitle']
-          end
+          # within '#bulk-update-status-table' do
```
macrouch: remove

spec/features/bulk_updates/bulk_updates_location_keywords_spec.rb
```diff
@@ -270,9 +272,10 @@
 
       context 'when viewing the collection' do
         before do
-          within '#bulk-update-status-table' do
-            click_on @find_and_replace_concept_response.body['EntryTitle']
-          end
+          # within '#bulk-update-status-table' do
```
macrouch: remove

spec/features/bulk_updates/bulk_updates_location_keywords_spec.rb
```diff
@@ -359,9 +362,10 @@
 
       context 'when viewing the collection' do
         before do
-          within '#bulk-update-status-table' do
-            click_on @clear_all_and_replace_concept_response.body['EntryTitle']
-          end
+          # within '#bulk-update-status-table' do
```
macrouch: remove

spec/features/bulk_updates/bulk_updates_platforms_spec.rb
```diff
@@ -82,9 +82,10 @@
 
       context 'when viewing the collection' do
         before do
-          within '#bulk-update-status-table' do
-            click_on @find_and_remove_concept_response.body['EntryTitle']
-          end
+          # within '#bulk-update-status-table' do
```
macrouch: delete

spec/features/bulk_updates/bulk_updates_science_keywords_spec.rb
```diff
@@ -169,9 +169,10 @@
 
       context 'when viewing the collection' do
         before do
-          within '#bulk-update-status-table' do
-            click_on @add_to_existing_concept_response.body['EntryTitle']
-          end
+          # within '#bulk-update-status-table' do
```
macrouch: take out

spec/features/bulk_updates/form_validation_spec.rb
```diff
@@ -18,6 +32,48 @@
...
+      it 'displays the appropriate error messages for Name and Update Field' do
+        expect(page).to have_css('#bulk_update_name-error', text: 'Name is required.')
+        expect(page).to have_css('#update_field-error', text: 'Update Field is required.')
+      end
+    end
+
+    context 'when attempting to enter a duplicate name' do
+      before do
+        fill_in 'bulk_update_name', with: used_bulk_update_name
+        select 'Select a Field to Update', from: 'Field to Update'
```
macrouch: this is probably already selected, right?

app/controllers/bulk_updates_controller.rb
charleshuang80: yeah, I'm removing the line

spec/features/bulk_updates/repopulate_form_from_failed_bulk_updates_spec.rb
```diff
@@ -0,0 +1,468 @@
...
+
+  context 'when starting a new bulk update' do
+    before do
+      login
+
+      visit new_bulk_updates_search_path
+
+      # Search form
+      select 'Entry Title', from: 'Search Field'
+      find(:css, "input[id$='query_text']").set(@concept_response.body['EntryTitle'])
```
macrouch: we probably do this all over the bulk update tests, a helper would be useful, but maybe in the next ticket

app/helpers/bulk_updates_helper.rb
charleshuang80: I'll change this to use `fetch`.
But I think that if `task[:update_type]` was `nil`, so would `task[:update_field]` and since `field_to_update` won't be nil it couldn't get to this line.

spec/features/bulk_updates/bulk_updates_location_keywords_spec.rb
charleshuang80: Apologies if this gets a bit confusing: 
So the `before` blocks with `bulk_update_step_1: true` and  `bulk_update_step_2: true` run in those first 2 `it` blocks that specify those options and they already run the bulk update. The `it` block for this step doesn't have those options so it skips those steps, but the bulk update operation is already executed from the other `it` blocks being run first. Here were are just verifying the operation ran correctly by checking the collection, so we don't need them to run again. Also, if those `before` blocks were run again for this `it` block, we would run into an issue with the unique name constraint (which is why these tests were failing) _and_ for the science keyword facets some options would already have been removed and not selectable.

spec/features/bulk_updates/form_validation_spec.rb
charleshuang80: this is to trigger a `blur` on the Name field to trigger the validation error

app/assets/javascripts/select2.coffee
```diff
@@ -70,33 +70,55 @@ $(document).ready ->
...
 
-    if $dataContactType.length > 0
-      # the select is in the data contact form, so use the data contact type we are in
-      $relatedUrl = $dataContactType.find('.related-urls:first')
+      $relatedUrl = $element.parents('fieldset').find('.related-url')
+      $urlElement = $relatedUrl.find('.bulk-updates-url')
+      $urlContentTypeElement = $relatedUrl.find('.bulk-updates-related-url-content-type-select')
+      $urlTypeElement = $relatedUrl.find('.bulk-updates-related-url-type-select')
+
+      if $urlElement.val() != ''
```
macrouch: how much of this is the same as above? extract whatever you can from the two methods

spec/features/bulk_updates/bulk_updates_location_keywords_spec.rb
macrouch: 👍

spec/features/bulk_updates/form_validation_spec.rb
macrouch: 👍

app/assets/javascripts/select2.coffee
charleshuang80: Yeah, there's a good bit of duplicate code, I'll try and consolidate.

PR #218, opened 2018-02-20
MMT modifications necessary for CMR-4703 - a change to metadata_previ…

littleidiot40:
…ew gem.


jceaser:
is "render" a library you changed? If so, could you instead create a new function which accepts these new parameters and have the older function call the new one with the null values? Then you would be backwards compatible.

littleidiot40:
No. Render is part of Rails' ERB support

jceaser:
I found the definition on line, thanks, I see what is going on here.

PR #221, opened 2018-02-27
MMT-1267 UMM-S Preview page

macrouch:


lib/json_schema_form/umm_form.rb
```diff
@@ -25,22 +27,38 @@ def initialize(form_section_json: {}, json_form: {}, schema: {}, options: {}, ke
...
-        UmmFormElement.new(form_section_json: value, json_form: json_form, schema: schema, options: options, key: full_key, field_value: field_value)
+        UmmFormElement.new(form_section_json: value, json_form: json_form, schema: schema, options: @options, key: full_key, field_value: field_value)
       end
     end
   end
 
   def build_key(fragment, key)
-    key += fragment['key'] if fragment['key'] && !key.ends_with?(fragment['key'])
-    key
+    value = key
```
macrouch: was this needed?

lib/json_schema_form/umm_form_elements/umm_text_field.rb
```diff
@@ -1,10 +1,10 @@
 # :nodoc:
 class UmmTextField < UmmFormElement
   def render_markup
-    text_field_tag(keyify_property_name(form_fragment), text_value, element_properties(schema_fragment))
+    text_field_tag(keyify_property_name, text_value, element_properties(schema_fragment))
   end
 
   def render_preview
-    content_tag(:p, element_value)
+    content_tag(:p, text_value)
```
macrouch: this change was just updating render_preview to use the same value as render_markup, a change from 1014

charleshuang80:
Add breadcrumbs tests for service_drafts previews, empty and valid.

charleshuang80:
My comments we just discussed:
1. We should have some comments on how to use the new UMM Preview classes. They don't need to be very extensive and not all classes need them, but at least some comments on how to use the new classes
2. I  don't really like the other preview element types inheriting from UmmPreviewText, maybe have a UmmPreviewElement and have them all inherit from that?
3. Does the UmmPreviewText (or the possibly new UmmPreviewElement) need to be a new class or can it inherit from UmmPreviewForm, in a similar way the Forms does? (UmmFormElement < UmmForm)
4. Maybe not now, but the preview elements, specifically the ones with the preview card seem less generic than the forms. I think there should be something like a UmmPreviewCard that the related url and the service orgs can utilize, and most of the actual card implementation can be in that preview card class.

PR #223, opened 2018-03-01
MMT-1019 MMT-1273 publish service drafts and published service previe…

macrouch:
…w page

Conflicts:
	lib/json_schema_form/umm_preview.rb

### DIFF
--- a/app/controllers/manage_metadata_controller.rb
+++ b/app/controllers/manage_metadata_controller.rb
@@ -57,6 +57,61 @@ class ManageMetadataController < ApplicationController
     end
   end
 
+  def set_service
+    @concept_id = params[:service_id] || params[:id]
+    @revision_id = params[:revision_id]
+
+    # retrieve the variable metadata with the current umm_var version
+    headers = { 'Accept' => "application/#{Rails.configuration.umm_var_version}; charset=utf-8" }
+    service_concept_response = cmr_client.get_concept(@concept_id, token, headers, @revision_id)
+
+    @service = if service_concept_response.success?
+                  service_concept_response.body
+                else
+                  Rails.logger.error("Error retrieving concept for Variable #{@concept_id} in `set_service`: #{service_concept_response.inspect}")
+                  {}
+                end
+
+    set_service_information
+  end
+
+  def set_service_information
+    # search for service by concept id to get the native_id and provider_id
+    # if the service is not found, try again because CMR might be a little slow to index if it is a newly published record
+    attempts = 0
+    while attempts < 20
+      services_search_response = cmr_client.get_services(concept_id: @concept_id, all_revisions: true)
+
+      service_data = if services_search_response.success?
+                       services_search_response.body.fetch('items', [])
+                     else
+                       []
+                     end
+      service_data.sort! { |a, b| b['meta']['revision-id'] <=> a['meta']['revision-id'] }
+
+      @revisions = service_data
+      latest = service_data.first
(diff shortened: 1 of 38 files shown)


app/controllers/services_controller.rb
```diff
@@ -0,0 +1,150 @@
...
+
+      set_user_permissions
+    end
+
+    @draft = ServiceDraft.where(provider_id: @provider_id, native_id: @native_id).first
+
+    add_breadcrumb breadcrumb_name(@service, 'service'), service_path(params[:id])
+  end
+
+  def edit
```
macrouch: comment out unused methods

app/controllers/services_controller.rb
```diff
@@ -0,0 +1,150 @@
...
+  end
+
+  private
+
+  def set_schema
+    @schema = UmmJsonSchema.new('services', 'umm-s-json-schema.json')
+    @schema.fetch_references(@schema.parsed_json)
+  end
+
+  def set_form
```
macrouch: is this method used

app/controllers/manage_metadata_controller.rb
```diff
@@ -57,6 +57,61 @@ def set_variable_information
...
+                     else
+                       []
+                     end
+      service_data.sort! { |a, b| b['meta']['revision-id'] <=> a['meta']['revision-id'] }
+
+      @revisions = service_data
+      latest = service_data.first
+      meta = latest.nil? ? {} : latest.fetch('meta', {})
+
+      if !@revision_id.nil? && meta['revision-id'].to_s != @revision_id.to_s
```
charleshuang80: I think this might introduce a bug we have for variables where after publishing a new revision, you still might see a banner saying you are viewing an old revision. MMT-1186. I _think_ the solution might just be to add an `else` statement to set the ivar to false

lib/json_schema_form/umm_preview.rb
```diff
@@ -1,13 +1,17 @@
 # :nodoc:
 class UmmPreview < JsonFile
-  attr_accessor :resource, :forms
+  attr_accessor :data, :forms, :draft_id
 
-  def initialize(schema_type:, preview_filename:, resource:)
-    super(schema_type, preview_filename)
+  # schema_type:      published_resource_name (service/variable)
```
charleshuang80: what's the reasoning for this to be the singular form, rather than `plural_published_resource_name`? It seems that all the other UMM classes use the plural, and you're having to call `.pluralize` most of the time `schema_type` is used, including to call `super`.

macrouch: I think it is used 4 places, 2 of them need singular a 2 need plural, I liked the idea of using the simpler version in the param.

spec/features/service_drafts/preview/empty/service_information_preview_spec.rb
```diff
@@ -0,0 +1,77 @@
+require 'rails_helper'
```
charleshuang80: is this one of the tests that got 'lost'?

macrouch: yep,

`service_drafts/preview/*/service_information_preview_spec.rb` and
`service_drafts/preview/*/service_keywords_preview_spec.rb`

lib/json_schema_form/umm_preview.rb
charleshuang80: I do think it's a bit odd to have to change it when you're calling `super`, but I guess it might be more a matter of opinion.

PR #226, opened 2018-03-05
MMT-1021 MMT-1270 search services, sort services search results

macrouch:


app/assets/javascripts/search.coffee
```diff
@@ -43,19 +43,20 @@ $(document).ready ->
     switch recordType
       when 'variables'
         $searchButtonSpan.text('Search Variables').addClass('variable-text')
+      when 'services'
+        $searchButtonSpan.text('Search Services').addClass('service-text')
       else
-        $searchButtonSpan.text('Search Collections').removeClass('variable-text')
+        $searchButtonSpan.text('Search Collections')
```
charleshuang80: should we have a class for collections also, i.e. `collection-text`

app/assets/javascripts/search.coffee
```diff
@@ -43,19 +43,20 @@ $(document).ready ->
...
 
 
   $('#record_type_collections').on 'click', ->
-    $("#search-submit-button-text").text('Search Collections').removeClass('variable-text')
+    $("#search-submit-button-text").text('Search Collections').removeClass('variable-text').removeClass('service-text')
```
charleshuang80: going with the above comment, seems inconsistent that for collections you are removing two classes, but for variables and services you are adding a class and removing a class

app/assets/javascripts/search.coffee
macrouch: the reason there is a variable-text class is to keep the button the same width as it is when 'Search Collections is visible, I missed adding in the css for service-text, so I'll add in collection-text as well

PR #227, opened 2018-03-06
MMT-1275 view revisions for published services

macrouch:


app/views/services/show.html.erb
```diff
@@ -66,9 +66,9 @@
         <%#= link_to 'Download JSON', download_json_service_path(@concept_id, revision_id: @revision_id), class: 'eui-btn--link', target: '_blank' %>
 
         <% if current_provider?(@provider_id) %>
-          <%#= link_to 'Delete Service Record', "#delete-record-modal", class: 'display-modal delete-service eui-btn--link bar-before' %>
+          <%= link_to 'Delete Service Record', "#delete-record-modal", class: 'display-modal delete-service eui-btn--link bar-before' %>
```
charleshuang80: the links currently look a bit odd since there are 2 horizontal bars where a link (to be added later) is missing. the `.bar-before` and `.bar-after` classes add the bars, so I think it's easy enough to add or remove those classes as needed for now. but if everyone else is ok leaving it as is since those tickets might be implemented soon, then I'm ok with that.

charleshuang80:
I am good with this, pending the decision on my comment about the bars.

app/views/services/show.html.erb
macrouch: that ticket is the one I'm picking up as soon as this gets merged, so I am in favor of letting this go for now

@jceaser what do you think?

charleshuang80: well - I can be ok with it then. But I think it _could_ continue to have the lines showing a missing link if none of those classes are changed until we add the last link, so I think we should avoid that.

macrouch: ok, I'll take a look in 1274

PR #230, opened 2018-03-08
MMT-1272 MMT-1276 clone published service records and revert publishe…

macrouch:
…d service records

### DIFF
--- a/spec/features/services/reverting_service_spec.rb
+++ b/spec/features/services/reverting_service_spec.rb
@@ -1,0 +1,153 @@
+require 'rails_helper'
+
+describe 'Reverting to previous services', js: true do
+  before :all do
+    # service for simple reverting service test
+    @simple_revert_ingest_response, @simple_revert_concept_response = publish_service_draft(revision_count: 2)
+
+    # service for reverting service with many revisions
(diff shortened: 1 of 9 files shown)


spec/features/services/reverting_service_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+            visit service_revisions_path(@multiple_revisions_ingest_response['concept-id'])
+
+            within '#service-revisions-table tbody tr:last-child' do
+              # make sure we are clicking on the correct link
+              expect(page).to have_content('1 - Revision')
+
+              click_on 'Revert to this Revision'
+            end
+
+            find('.not-current-provider-link').click
```
charleshuang80: I see here that you're clicking on the not current provider modal, but there don't seem to be any tests for the modal and switching providers for reverting. Should we have that here?

macrouch: yeah, looks like it is missing

spec/features/variables/reverting_variable_spec.rb
```diff
@@ -78,14 +78,16 @@
             wait_for_ajax
           end
 
-          it 'reverts the variable to the correct revision and displays the correct revision information' do
+          it 'reverts the variable to the correct revision and displays the correct revision information and switches provider context' do
```
charleshuang80: sorry to be a pain, but since you're making this match the services one, I think it should have the test for the modal also...

PR #234, opened 2018-03-21
MMT-1356 Improve test time and reliability 

macrouch:


spec/features/providers_list_spec.rb
```diff
@@ -78,7 +78,8 @@
 
       context 'when the user then logs in' do
         before do
-          login
+          # TODO need to do a real login because it fetches providers from CMR
```
charleshuang80: still need the comment?

spec/support/user_helpers.rb
```diff
@@ -1,11 +1,14 @@
 module Helpers
   module UserHelpers
-    def login_admin
-      login(admin: true)
+    def login_admin(provider: 'MMT_2', providers: 'MMT_2')
```
charleshuang80: I kind of like the idea of having a `def real_login`. I think it could be a separate method, or it could call `login` passing in `real_login: true`, similar to how you originally made this `login_admin` method that calls `login`.

PR #242, opened 2018-04-11
MMT-1369 add a filter to the right hand side of the chooser

macrouch:


app/assets/javascripts/widgets/Chooser/Chooser.coffee
```diff
@@ -98,16 +100,17 @@ window.Chooser = (config) ->
     setDefault('filterChars', 1)
     setDefault('resetSize', 50)
     setDefault('showNumChosen', true)
-    setDefault('filterText', 'Enter text to narrow search results')
+    setDefault('fromFilterText', 'Enter text to narrow search results')
```
charleshuang80: I know the text in this box has been this way for a while, so I'm not sure if it needs to change, but I wonder if it would be better if this said something like 'Enter a Short Name to narrow search results' or 'Enter text to narrow search results by Short Name'

spec/features/chooser/chooser_to_filter_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+        # testing 02_01 | My testing title 02
+        find('option[value="C1200189943-MMT_2"]').select_option
+      end
+
+      within '.button-container' do
+        find('.add_button').click
+      end
+    end
+
+    it 'moves two values to the TO list' do
```
charleshuang80: should be 'three values' right?

spec/features/chooser/chooser_to_filter_spec.rb
```diff
@@ -0,0 +1,119 @@
+require 'rails_helper'
+
+describe 'Chooser To Filter', js: true do
+  before do
+    collections_response = Cmr::Response.new(Faraday::Response.new(status: 200, body: JSON.parse(File.read('spec/fixtures/cmr_search.json'))))
```
charleshuang80: I made a small helper, `cmr_success_response`, to make this a bit easier, but I don't think it's used in all places (yet). If you want to use it, this could be changed to `collections_response = cmr_success_response(File.read('spec/fixtures/cmr_search.json'))`

PR #248, opened 2018-05-07
MMT-1431: put URS login behind feature toggle

charleshuang80:


app/controllers/users_controller.rb
```diff
@@ -2,15 +2,19 @@
...
 
   def login
     session[:last_point] = request.referrer
     session[:last_point] = params[:next_point] if params[:next_point]
 
-    redirect_to cmr_client.urs_login_path
+    if urs_login_required?
+      redirect_to cmr_client.urs_login_path
+    else
+      redirect_to root_url
```
charleshuang80: remove

app/controllers/application_controller.rb
```diff
@@ -122,11 +103,66 @@ def store_profile(profile = {})
...
+
+      store_oauth_token unless is_user_logged_in # clear session time and token values if user is not logged in
+      is_user_logged_in
+    end
+
+  end
+
+  def is_logged_in
+    if !urs_login_required? # && !launchpad_login_required?
+      # TODO this should only happen if URS login AND Launchpad login are both turned off (which should not happen)
```
charleshuang80: Log error message

app/controllers/manage_metadata_controller.rb
```diff
@@ -2,6 +2,31 @@
 class ManageMetadataController < ApplicationController
   include ManageMetadataHelper
```
macrouch: I know you didn't add this, but `app/helpers` are supposed to be view helpers. Since you are doing a little bit of refactoring, can you look at removing the need for including ManageMetadataHelper here

app/controllers/application_controller.rb
```diff
@@ -7,13 +7,24 @@ class ApplicationController < ActionController::Base
...
   protected
 
+  # Seconds ahead of the token expiration that the server should
+  # attempt to refresh the token
+  SERVER_EXPIRATION_OFFSET_S = 60
+  # For testing, token expires after 10 seconds
+  # SERVER_EXPIRATION_OFFSET_S = 3590
+
+  def urs_login_required?
+    ENV['urs_login_required'] == 'true'
```
macrouch: I think it might make sense for this to default to true. That will save work for the SA team for having to change environment variables until we actually need to change this.

macrouch:
Add a test for when `urs_login_required?` returns false

app/controllers/application_controller.rb
```diff
@@ -99,7 +69,18 @@ def available_provider?(provider)
...
+  def provider_set?
+    if logged_in? && current_user.provider_id.nil?
+      redirect_to provider_context_path
+    end
+  end
+
   def store_profile(profile = {})
+    # URS login
+
+    # TODO remove this comment with Launchpad feature toggle ticket
```
macrouch: do you know the ticket number for this upcoming ticket?

charleshuang80: 1386 - I didn't think it was necessary to add the ticket number because it was the one that was immediately next.

macrouch: 👍

PR #253, opened 2018-06-01
MMT-1297 add launchpad keep alive call to keep sessions active

macrouch:


app/controllers/saml_controller.rb
```diff
@@ -56,6 +56,7 @@ def acs
       # it should be set to 15 min, the default Launchpad session time
       session[:expires_in] = 900
       session[:logged_in_at] = Time.now.to_i
+      session[:original_logged_in_at] = Time.now.to_i
```
charleshuang80: since you are adding this to the session, maybe should it also be added to the check in `logged_in?`

app/assets/javascripts/launchpad_keep_alive.coffee
```diff
@@ -0,0 +1,53 @@
...
+    fullSessionLength = 840000 # 14 minutes (in milliseconds)
+    lastActiveTime = Date.now()
+
+    # sessionStart is the beginning of the session in which the user was last active
+    sessionStart = Date.now()
+
+    # if the user changes pages in the middle of the session window
+    # the keep alive timer will need to be shortened to ensure
+    # the keep alive call is successful
+    sessionLength = if remainingSessionTime > sessionLength then sessionLength else remainingSessionTime
```
charleshuang80: since this isn't declared (since I thought we could combine the statements) I had thought this might need to be changed to `sessionLength = if remainingSessionTime > fullSessionLength then fullSessionLength else remainingSessionTime`

macrouch: https://gph.is/1W2P1V1

app/controllers/application_controller.rb
```diff
@@ -202,12 +203,29 @@ def refresh_launchpad_if_needed
...
     end
   end
 
+  def refresh_launchpad
+    return { error: 'session older than 8 hours' } if Time.now.to_i - session[:original_logged_in_at] > 28_800
+
+    response = cmr_client.keep_alive(token)
+    Rails.logger.info "launchpad integration keep alive endpoint response: #{response.inspect}" if Rails.env.development?
+    if response.success?
+      session[:launchpad_cookie] = response.headers.fetch('set-cookie', '').split("#{launchpad_cookie_name}=").last
```
charleshuang80: I had added `pull_launchpad_cookie` in `SamlController` in anticipation of needing to use it for this. We can leave this as is, but you could also use that as well.

PR #254, opened 2018-06-06
MMT-1432: prompt user to associate a URS account with their Launchpad account

charleshuang80:
* add actions to prompt user to associate an account, handle URS login for association, and confirm that association
* add appropriate views
* update methods used for logging in
* add/update tests and test helpers


config/routes.rb
```diff
@@ -131,7 +131,11 @@
 
   get 'login' => 'users#login', as: 'login'
   get 'logout' => 'users#logout'
-  get 'urs_callback' => 'oauth_tokens#urs_callback'
+  get 'prompt_urs_association' => 'users#prompt_urs_association'
+  get 'confirm_urs_association' => 'users#confirm_urs_association'
+  post 'associate_urs_and_launchpad_ids' => 'users#associate_urs_and_launchpad_ids'
+  get 'urs_login_callback' => 'oauth_tokens#urs_login_callback'
```
macrouch: I assume you updated all of the URS applications for the new callback uris? SIT UAT and PROD?

charleshuang80: not yet in UAT and PROD (I will check if I have access). I did update it in SIT, that was one of the hiccups this morning.

app/views/shared/_user_info_dropdown.html.erb
```diff
@@ -17,13 +17,13 @@
         </p>
       </a>
       <!-- Display provider context badge -->
-      <a href="#provider-context" id="provider-badge-link" class="display-modal">
-        <div class="provider-context">
-          <% if is_logged_in && current_user.provider_id %>
+      <% if logged_in? && current_user.provider_id %>
```
macrouch: it doesn't look like `logged_in?` is needed here, line #1 checks it

app/views/users/confirm_urs_association.html.erb
```diff
@@ -0,0 +1,33 @@
+<div class="urs-assocation">
+  <div class="row row-content">
+    <div class="content">
+      <section>
+        <h2>Confirm URS and Launchpad account association</h2>
+
+        <fieldset>
+          <p>Dear <%= "#{@profile.fetch('first_name', '')} #{@profile.fetch('last_name', '')}," %> </p>
```
macrouch: Double check with Alicia and Thomas that this is the language we want to use in these two paragraphs. Just thinking that in the past any time we have more than a simple line of text we made sure that it was approved by Kathy first

app/views/users/prompt_urs_association.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="urs-association">
+  <div class="row row-content">
+    <div class="content">
+      <section>
+        <fieldset>
+          <h2>Redirecting to URS</h2>
+          <p>It appears you do not have a URS account linked with your Launchpad account. MMT and CMR permissions (ACLs) are based on your URS account. Please click the button below to log in to the URS account you listed on your NAMS request and would like to associate with your Launchpad account.</p>
```
macrouch: Same with this paragraph.

app/views/users/confirm_urs_association.html.erb
charleshuang80: @lhlyles has a ticket to verify the language with Alicia and update with any text changes, which I think is a better process than what we had previously

app/controllers/users_controller.rb
```diff
@@ -106,19 +142,14 @@ def get_all_providers
   end
 
   def get_urs_profile_from_auid
-    # We are assuming the user has their urs id and auid associated in URS
-    # TODO MMT-1432 will implement directing a user to URS to sign in and associate their urs_uid and auid the first time, if those ids are not associated
-
     urs_profile_response = cmr_client.get_urs_uid_from_nams_auid(session[:auid])
+    if urs_profile_response.success?
+      urs_profile_response.body
+      # TODO: check if user has auid. store auid if not
```
macrouch: this is OK without storing the auid

spec/features/users/user_access_spec.rb
```diff
@@ -1,15 +1,17 @@
...
+    end
 
+    context 'when a logged out user tries to view an access restricted page', js: true do
+      before do
         visit search_path
       end
 
-      it 'redirects them to URS for login' do
+      it 'redirects them to URS to login' do
         expect(page).to have_content('EARTHDATA LOGIN')
```
charleshuang80: So this test still confuses me why it works for Earthdata Login but not Launchpad, but logging `page.current_path` says it is on `/oauth/authorize`, and the screenshot indicates it seems to work:
capybara-201806061814306141569212

app/views/users/confirm_urs_association.html.erb
jceaser: yea, this text does not have to be perfect, we can that updated latter.

PR #257, opened 2018-06-08
Feature/mmt 1443

jceaser:
Chris, James, and added the jQuery UI tooltip to the chooser. Also in this request is a rake command to generate ctags for editing in vim and bbedit.


app/assets/javascripts/order_option_assignments.coffee
```diff
@@ -146,4 +146,6 @@ $(document).ready ->
 
     # select all of the items in the "to" list before submitting
     $('#new-order-option-assignment-form').submit ->
-      $('#collectionsChooser_toList option').prop('selected', true)
\ No newline at end of file
+      $('#collectionsChooser_toList option').prop('selected', true)
+   
+  $(document).tooltip()
```
jceaser: move to a new script, tooltip.coffee

lib/tasks/ctags.rake
```diff
@@ -0,0 +1,9 @@
+# add a rake task to make it easy to navigate ruby code with VIM using ctr-] and
+# to jump and ctr-t to return
+# @author thomas.a.cherry
+namespace :ctags do
+  desc "generate ctags from ruby files to make vim more fun"
+  	task :ctags do
```
jceaser: don't indent

app/assets/javascripts/order_option_assignments.coffee
macrouch: maybe also add some comments to that new file, explaining what it is doing

.editorconfig
```diff
@@ -0,0 +1,4 @@
+[*]
```
charleshuang80: should this file be ignored?

app/assets/javascripts/tooltip.coffee
```diff
@@ -0,0 +1,12 @@
+#
+# This code uses jquery-ui to replace the built in tooltip popup with a nicer
+# one. The popup can be styled but by default is slightly larger and pops up
+# faster. The tooltip function only needs to be called once as it searches
+# the DOM attaching the new UI to any tag with a defined tooltip.
+#
+
+$(document).ready ->
+  if $("#bulk-updates-form").length < 1
```
charleshuang80: This means that the tooltip will only show up on pages that have elements with the `#bulk-updates-form` id. I'm not sure that is what we want...

macrouch: `< 1`, so this does work. I find it confusing too, maybe `unless $("#bulk-updates-form").length > 1`

charleshuang80: ah - I misread it. I like Matthew's proposed change.

.editorconfig
jceaser: I meant for it to be included, it causes vim to use the local coding standards. I picked this up from other open source projects I contribute to (maybe consume is the right word here).

app/assets/javascripts/tooltip.coffee
jceaser: done

spec/features/service_entries/update_service_entry_spec.rb
```diff
@@ -90,8 +90,11 @@
 
           # Remove and add back the second option
           select "lorem_223 | ipsum", from: 'tag_guids_fromList'
+          page.find('body').native.send_keys :escape
+
           find('button[title=remove]').click
           select "lorem_223 | ipsum", from: 'tag_guids_toList'
+          #page.find('body').native.send_keys :escape
```
charleshuang80: if this isn't needed we should remove it.

spec/features/permissions/create_update_delete_permissions_spec.rb
```diff
@@ -188,10 +188,15 @@
             within '#collectionsChooser' do
               # selecting each individually as it seems more robust.
               select(@entry_id_1, from: 'Available Collections')
+              page.find('.from-container').native.send_keys :escape # press esc to hide tool tip, it blocks button
               find('button[title=add]').click
+              
               select(@entry_id_2, from: 'Available Collections')
+              page.find('.from-container').native.send_keys :escape # press esc to hide tool tip, it blocks button
```
charleshuang80: since all 3 instances of using escape here are so close together, maybe just the comment for the first time would be sufficient?

spec/features/service_entries/update_service_entry_spec.rb
```diff
@@ -90,8 +90,11 @@
 
           # Remove and add back the second option
           select "lorem_223 | ipsum", from: 'tag_guids_fromList'
+          page.find('body').native.send_keys :escape
```
charleshuang80: you could add the comment here, though

spec/features/service_entries/update_service_entry_spec.rb
macrouch: :this:

spec/features/permissions/create_update_delete_permissions_spec.rb
jceaser: will do

spec/features/service_entries/update_service_entry_spec.rb
jceaser: done

spec/features/service_entries/update_service_entry_spec.rb
jceaser: agree

PR #261, opened 2018-06-21
MMT-1505 - updating a gem file to make builds happy

jceaser:
Updated one outdated gem. Pretty strait forward. Bamboo is green. Adding Jon V. since he was one of the people reporting this, just so that he sees progress.


jceaser:
got two others to look at this, will merge now

PR #262, opened 2018-06-21
MMT-1418-Fixed the issue where tophat2 was pushing rest of webpage down and hiding part of the footer

zjam1:
Added a div tag inside of wrapper to push down header. Also changed top margins for the wrapper class.


app/assets/javascripts/application.js
```diff
@@ -31,5 +31,6 @@
 //= require jquery.webui-popover/jquery.webui-popover.min
 //= require redactor/redactor.min
 //= require jquery-textcounter/textcounter.min
+//= require tophat_space.coffee
```
charleshuang80: Does this need to be added? I think it should be automatically picked up with the `require_tree` below.

app/assets/javascripts/tophat_space.coffee
```diff
@@ -0,0 +1,5 @@
+$('document').ready ->
+  #checks if tophat exists 
+  tophatExist = document.getElementsByClassName("th-wrapper")
```
charleshuang80: Sorry I didn't notice this before, but you should make this call with JQuery, which you are using above and below

jceaser: yea, that is a good point. https://stackoverflow.com/questions/23648196/converting-document-getelementsbyclassname-to-jquery has an example of this.

zjam1:
I made the changes that Charles requested

app/assets/javascripts/tophat_space.coffee
```diff
@@ -0,0 +1,5 @@
+$('document').ready ->
+  # checks if tophat exists
+  tophatExist = $(".th-wrapper")
+  if tophatExist.length
```
charleshuang80: I was going to say you should add a ` > 0` here, but upon further research it looks like this is a widely accepted way to do this. However, I will suggest one final change, you can shorten this by making it one line: `if $('.th-wrapper).length`

PR #264, opened 2018-06-21
MMT-1218 Collection Data Type Display

zjam1:
When on the collection preview page. If the Collection Data Type is Near Real Time then the NRT badge displays next to the subtitle in the header section.

### DIFF
--- a/app/helpers/collections_helper.rb
+++ b/app/helpers/collections_helper.rb
@@ -34,4 +34,10 @@ module CollectionsHelper
     link_to(download_link_title, download_collection_xml_path(concept_id, download_format, revision_id),
             id: download_format)
   end
+
+  def nrt_badge(metadata)
+    if metadata['CollectionDataType'] == 'NEAR_REAL_TIME'
+      content_tag :span, 'NRT', class: 'eui-badge nrt'
+    end
+  end
 end
--- a/app/helpers/collections_helper.rb
+++ b/app/helpers/collections_helper.rb
@@ -37,7 +37,7 @@ module CollectionsHelper
 
   def nrt_badge(metadata)
     if metadata['CollectionDataType'] == 'NEAR_REAL_TIME'
-      content_tag :span, 'NRT', class: 'eui-badge nrt'
+      content_tag(:span, 'NRT', class: 'eui-badge nrt')
     end
   end
 end
--- a/app/views/collection_drafts/show.html.erb
+++ b/app/views/collection_drafts/show.html.erb
@@ -1,6 +1,9 @@
 <% content_for :header_title do %>
   <h2><%= display_entry_id(get_resource.draft, resource_name) %></h2>
   <p class="subtitle"><%= display_collection_entry_title(get_resource.draft) %></p>
+  <% if get_resource.draft['CollectionDataType'] == "NEAR_REAL_TIME" %>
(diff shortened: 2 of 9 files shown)


jceaser:
need unit test to test for the existence of the tag

app/views/collection_drafts/show.html.erb
```diff
@@ -1,6 +1,9 @@
 <% content_for :header_title do %>
   <h2><%= display_entry_id(get_resource.draft, resource_name) %></h2>
   <p class="subtitle"><%= display_collection_entry_title(get_resource.draft) %></p>
+  <% if get_resource.draft['CollectionDataType'] == 'NEAR_REAL_TIME' %>
```
jceaser: these next three sections look like they mostly repeat, maybe a helper function

spec/features/collection_drafts/breadcrumbs_and_header_spec.rb
```diff
@@ -31,7 +31,7 @@
       visit collection_draft_path(@draft)
     end
 
-    it 'displays the short name' do
+    it 'displays the short name within in breadcrumbs' do
```
charleshuang80: really minor, but I think there's a typo. the 'in' after 'within' should be changed or removed.

app/helpers/collections_helper.rb
```diff
@@ -34,4 +34,10 @@ def render_collection_download_xml_link(download_format, download_link_title, co
     link_to(download_link_title, download_collection_xml_path(concept_id, download_format, revision_id),
             id: download_format)
   end
+
+  def nrt_badge(metadata)
+    if metadata['CollectionDataType'] == 'NEAR_REAL_TIME'
+      content_tag :span, 'NRT', class: 'eui-badge nrt'
```
charleshuang80: I know parentheses aren't required here, but if you look at where we use `content_tag` in the rest of the project, we wrap the arguments in parentheses, so it we should be consistent with that.

app/views/collection_drafts/show.html.erb
```diff
@@ -1,6 +1,7 @@
 <% content_for :header_title do %>
   <h2><%= display_entry_id(get_resource.draft, resource_name) %></h2>
   <p class="subtitle"><%= display_collection_entry_title(get_resource.draft) %></p>
+  <%= nrt_badge get_resource.draft %>
```
charleshuang80: also please use parentheses with this helper to be consistent. as you can see, the view helpers above are called with parentheses.

app/views/collections/revisions.html.erb
```diff
@@ -1,6 +1,7 @@
 <% content_for :header_title do %>
   <h2><%= display_entry_id(@collection, 'collections') %></h2>
   <p class="subtitle"><%= display_collection_entry_title(@collection) %></p>
+  <%= nrt_badge @collection %>
```
charleshuang80: and here

app/views/collections/show.html.erb
```diff
@@ -1,6 +1,7 @@
 <% content_for :header_title do %>
   <h2><%= display_entry_id(@collection, 'collections') %></h2>
   <p class="subtitle"><%= display_collection_entry_title(@collection) %></p>
+  <%= nrt_badge @collection %>
```
charleshuang80: and here

PR #265, opened 2018-06-25
MMT-1491 add scheduler to cleanup expired sessions

htranho:


macrouch:
from Jon  V, a good place to log to would be `${RAILS_HOME}/logs/...`

config/schedule.rb
```diff
@@ -0,0 +1,26 @@
...
+#   rake "some:great:rake:task"
+# end
+#
+# every 4.days do
+#   runner "AnotherModel.prune_old_records"
+# end
+
+# Learn more: http://github.com/javan/whenever
+
+set :output, 'log/session_cleanup_cron.log'
```
jceaser: good to point out that this may need to change in the future, I'll write up a ticket in tech-dept to see if we can fix this in the future.

PR #268, opened 2018-06-28
MMT-873 added a better health check for the status page

zjam1:
Added a more in depth health check that looks to make sure a database connection exists and whether Launchpad is working or not. Changed the tests associated with the status page to work as well.


app/middleware/middleware_healthcheck.rb
```diff
@@ -0,0 +1,32 @@
...
+        db_healthy = ActiveRecord::Migrator.current_version != 0
+      rescue StandardError => e
+        Rails.logger.info "Database error: #{e}"
+        db_healthy = false
+      end
+
+      # checks the health of launchpad
+      cmr_client = Cmr::Client.client_for_environment(Rails.configuration.cmr_env, Rails.configuration.services)
+      launchpad_healthy = cmr_client.launchpad_healthcheck.body == 'OK'.freeze
+      response[2] = ["{\"database\": #{db_healthy}, \"launchpad\": #{launchpad_healthy}}"]
```
jceaser: I think we should check to see if we should add "ok?":"true" if the application is healthy (and all dependencies).

jceaser:
I talked to @ccuadrado, he does not require any other changes or to preview this before we merge.

app/middleware/middleware_healthcheck.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      # If launchpad is disabled then we will not report a 500 error if launchpad still fails
+      unless (ENV['launchpad_login_required'] != 'true' || launchpad_healthy) && db_healthy
```
charleshuang80: I think you should change this to use `if`, it will be easier to be sure what we are checking for.
http://www.railstips.org/blog/archives/2008/12/01/unless-the-abused-ruby-conditional/

app/middleware/middleware_healthcheck.rb
```diff
@@ -0,0 +1,32 @@
...
+      begin
+        db_healthy = ActiveRecord::Migrator.current_version != 0
+      rescue StandardError => e
+        Rails.logger.info "Database error: #{e}"
+        db_healthy = false
+      end
+
+      # checks the health of launchpad
+      cmr_client = Cmr::Client.client_for_environment(Rails.configuration.cmr_env, Rails.configuration.services)
+      launchpad_healthy = cmr_client.launchpad_healthcheck.body == 'OK'.freeze
```
charleshuang80: So when I was testing this call previously, if we were unable to reach launchpad the call would fail pretty quickly. I'm not sure what is different now, but locally, if I am not on the vpn, this call hangs for a while and the page we are expecting (showing the response as text) doesn't show. I am wondering if we should add a timeout so if the call to launchpad doesn't finish within that time it will allow the response to be rendered. Since it is happening locally, it is possible it could happen in our other environments also.
What do you think, @jceaser ?

app/middleware/middleware_healthcheck.rb
charleshuang80: Also, if we don't need to check on launchpad if launchpad login is not required, then we should completely skip the call.

zjam1: I talked with Catalino about this. He said that in this case he would still like to do the launchpad healthcheck but not return a 503 error if the check fails.

app/middleware/middleware_healthcheck.rb
jceaser: yes, I think we should, the difference might be in how it fails, if the client is waiting for a DNS resolution vs the actual server. The VPN may affect this. Lets try specifying a timeout, I think if we don't hear from launchpad in a second, we have a problem.

app/middleware/middleware_healthcheck.rb
```diff
@@ -0,0 +1,36 @@
...
+        db_healthy = ActiveRecord::Migrator.current_version != 0
+      rescue StandardError
+        db_healthy = false
+      end
+
+      # checks the health of launchpad
+      cmr_client = Cmr::Client.client_for_environment(Rails.configuration.cmr_env, Rails.configuration.services)
+      launchpad_healthy = true
+      begin
+        timeout(2) { launchpad_healthy = cmr_client.launchpad_healthcheck.body == 'OK'.freeze }
```
charleshuang80: This timeout seems really short. I would suggest increasing it to at least 10 or 15s. What do you think is appropriate, @jceaser?

app/middleware/middleware_healthcheck.rb
```diff
@@ -5,24 +5,28 @@ def initialize(app)
...
   def call(env)
     if env['PATH_INFO'.freeze] == '/status'.freeze
-      response = [200, { 'Content-Type' => 'application/json' }]
+      response = [503, { 'Content-Type' => 'application/json' }]
 
       # checks the database health
       begin
         db_healthy = ActiveRecord::Migrator.current_version != 0
-      rescue StandardError => e
-        Rails.logger.info "Database error: #{e}"
```
charleshuang80: was there a reason to remove this? seems like it would be helpful to log this error.

zjam1: I originally had it in there for debugging purposes, should I just keep it in there?

charleshuang80: I think it would be helpful if it ever actually happened

PR #269, opened 2018-06-28
MMT-1491 Added rake task as wrapper to call sessions cleanup task

htranho:


lib/tasks/sessions_cleanup.rake
```diff
@@ -0,0 +1,8 @@
+namespace :sessions_cleanup_cron do
+  desc 'Run task to clean up sessions'
+  task :cleanup do
+    Rake::Task['db:sessions:trim'].invoke
+    current_time = Time.now.strftime('%m/%d/%Y %H:%M')
```
charleshuang80: In `config/initializers/time_formats.rb`, we have a list of time/date formats we use. Can you make the format match the default? If this task able to access them, you should be able to get the default with `Time.now.to_s`.

htranho: now using the mmt time format

PR #271, opened 2018-07-03
MMT-1468 added rspec test for deleting collection while switching prvider context and fixed bugs for that case

zjam1:


app/views/collections/show.html.erb
```diff
@@ -73,7 +73,7 @@
           <% if current_provider?(@provider_id) %>
             <%= link_to 'Delete Collection Record', "#delete-record-modal", class: 'display-modal delete-collection eui-btn--link bar-before' %>
           <% elsif available_provider?(@provider_id) %>
-            <%= link_to 'Delete Collection Record', '#not-current-provider-modal', class: 'display-modal not-current-provider eui-btn--link bar-before', data: { provider: @provider_id, record_action: 'delete-collection' } %>
+            <%= link_to 'Delete Collection Record', '#not-current-provider-modal', class: 'display-modal delete-collection not-current-provider eui-btn--link bar-before', data: { provider: @provider_id, record_action: 'delete-collection' } %>
```
macrouch: this still has references to the provider modal, `'#not-current-provider-modal'` and `class: '... not-current-provider ...`

zjam1: I'm pretty sure those still need to exist for the link to function correctly

macrouch: The link path needs to be changed to `#delete-collection`. The class needed to be changed to `delete-collection`, which you already did, so just remove `not-current-provider`

zjam1: I did the changes you asked and I got this ```You need to change your current provider to this collection. Click here to change your provider.``` and the collection didn't delete as well

macrouch: I made those changes and your test passed, so something is wrong either in your test or with the changes that need to be made. Either way, it is confusing as written, the delete link shouldn't be pointing to the not current provider modal.

macrouch: my bad, `#delete-record-modal`

macrouch: After chatting with James, I think this is fine

PR #275, opened 2018-07-10
MMT-1339

jceaser:
Adding a check for non-existing provider orders. This change will cause the view and edit interfaces to display a "friendly" message when a provider order ID does not exist in CMR.

### DIFF
--- a/app/controllers/provider_orders_controller.rb
+++ b/app/controllers/provider_orders_controller.rb
@@ -101,7 +101,8 @@ class ProviderOrdersController < ManageCmrController
 
       order
     else
-      { 'error' => order_response.error_message || "Could not load a provider order due to an unspecified error." }
+      error_message = order_response.error_message || 'Could not load a provider order due to an unspecified error.'
+      { 'error' => error_message }
     end
   end
 end
--- a/app/controllers/provider_orders_controller.rb
+++ b/app/controllers/provider_orders_controller.rb
@@ -96,7 +96,7 @@ class ProviderOrdersController < ManageCmrController
         catalog_items << catalog_item
       end
 
-      order['catalog_items'] = catalog_items.sort { |a, b| a['item_guid'] <=> b['item_guid'] }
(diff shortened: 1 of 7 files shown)


app/controllers/provider_orders_controller.rb
```diff
@@ -54,49 +54,54 @@ def resubmit
...
 
-    order['catalog_items'] = catalog_items.sort { |a, b| a['item_guid'] <=> b['item_guid'] }
-    order['status_messages'] = provider_order.fetch('StatusMessage', '').split("\n").map { |m| m.split(' : ') }
+      order['catalog_items'] = catalog_items.sort_by { |a| a['item_guid'] }
+      order['status_messages'] = provider_order.fetch('StatusMessage', '').split("\n").map { |m| m.split(' : ') }
 
-    order
+      order
+    else
+      { 'error' => order_response.parsed_body.fetch('faultstring', true) }
```
charleshuang80: 1. in `lib/echo/response.rb`, there are a few different methods to get at the response body and errors. that’s where I found `parsed_body`, but it seems there are more detailed/helpful methods for `error_message` and `error_code` among others. We should definitely switch how we are grabbing the error message to use `error_message`, and if it is helpful we may want to use some of the other methods also. we may also want to modify `error_message` because it does not provide a default value to `fetch` which can produce an exception.
2. since the error message is something we are providing to the user to contact OPS with, we should probably have something other than `true`. I had initially thought we should use something similar to our I18n default error flash messages (in `config/locales/en.yml`), but this is not a flash message and is more similar to what we do in CollectionsController#set_collection_error_data, so I think the setup is fine, but we should change the default message to be something more informative than just `true` for this error so OPS has more information to discuss with us in the case where the response does not provide an error message in the way we are expecting.

jceaser:
great changes @charleshuang80, just waiting for a green build now.

jceaser:
I see a green build and two approves.

PR #280, opened 2018-07-20
MMT-1357 revised the MMT invitation email to make more sense

zjam1:


app/views/group_mailer/invite_user.text.erb
```diff
@@ -1,9 +1,5 @@
 Metadata Management Tool Invitation
 
-<%= @invite.user_first_name %>, <%= @invite.manager_name %> from <%= @invite.provider %> would like to add you to a group using NASA's Metadata Management Tool (MMT). To do so, MMT needs to access your Earthdata Login profile. Please authorize MMT to access your profile by clicking the link below. Thank you!
-
-Authorize MMT: <%= accept_invite_url(token: @invite.token) %>
-
-If you do not have an Earthdata Login profile, please create one prior to authorizing MMT.
+<%= @invite.user_first_name %>, <%= @invite.manager_name %> from <%= @invite.provider %> would like to add you to a group using NASA's Metadata Management Tool (MMT). To do so, MMT needs you to create an Earthdata Login profile. To do so click on the link below. Thank You!
```
charleshuang80: have you spoken about this language with Alicia? I am wondering if we should thinking about adding language (possibly at a later date) with information about Launchpad requirements as well.

zjam1: Yeah I talked with Alicia yesterday and she said to just disable this stuff for now because things will with Launchpad

app/controllers/groups_controller.rb
```diff
@@ -3,7 +3,7 @@ class GroupsController < ManageCmrController
   include GroupsHelper
   include GroupEndpoints
   include PermissionManagement
-
+  include Cmr
```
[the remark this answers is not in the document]
zjam1: Yeah, it didn't work without it.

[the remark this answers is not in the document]
charleshuang80: that seems very odd, because if you look through the controller, `cmr_client` is in most of the actions, yet we hadn't had to include it until this change. Can you try again? I will take a look also.

app/views/groups/new.html.erb
```diff
@@ -12,6 +12,6 @@
       <%= render 'form' %>
     <% end %>
 
-    <%= render 'invite_form' %>
+    <%# = render 'invite_form' #%>
```
charleshuang80: maybe add a comment also as to why this is being commented out.

PR #283, opened 2018-07-25
MMT-1361 refactored two classes into one base class

zjam1:
Refactored VariablesController and ServicesController into a new controller called BaseManageController.

### DIFF
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -7,14 +7,23 @@ class ServicesController < BasePublishedRecordController
   before_action :ensure_correct_provider, only: [:edit, :clone, :destroy]
   before_action :set_preview, only: [:show]
 
-  # https://stackoverflow.com/questions/34735540/action-defined-in-applicationcontroller-can-not-be-found
+  # If clone is not defined like this performing the clone action leads to a `action not found error`
+  # we looked into this but couldn't find a way around this without having clone call super
   def clone
     super
   end
 
   private
 
-  def resource_schema
-    's'
+  def resource_schema_file
+    'umm-s-json-schema.json'
+  end
+
+  def resource_form_file
+    'umm-s-form.json'
+  end
+
+  def resource_preview_schema
+    'umm-s-preview.json'
   end
 end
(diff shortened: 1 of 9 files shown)


app/controllers/services_controller.rb
```diff
@@ -1,147 +1,20 @@
...
-      Rails.logger.info("Audit Log: Service Draft for #{draft.entry_title} was created by #{current_user.urs_uid} in provider #{current_user.provider_id}")
-      redirect_to service_draft_path(draft), flash: { success: I18n.t('controllers.draft.service_drafts.create.flash.success') }
-    else
-      Rails.logger.info("User #{current_user.urs_uid} attempted to edit Service #{@concept_id} in provider #{current_user.provider_id} but a Service Draft was not created to edit because there was no native_id (#{@native_id}) found.")
-      # if we cannot locate the native_id for the Service, we should discontinue editing
-      redirect_to service_path(@concept_id, revision_id: @revision_id), flash: { error: I18n.t('controllers.services.edit.flash.native_id_error') }
-    end
-  end
-
+  # https://stackoverflow.com/questions/34735540/action-defined-in-applicationcontroller-can-not-be-found
```
charleshuang80: So this link actually does not seem to quite explain the issue. I would rather you just write a comment describing the error raised, and that we have looked into it and couldn't get it to work without the declaring the method.

app/controllers/services_controller.rb
```diff
@@ -1,147 +1,20 @@
...
-    return if current_provider?(@provider_id)
-
-    set_record_action
-
-    set_user_permissions
-
-    set_preview
-
-    render :show
+  def resource_schema
```
charleshuang80: I was thinking we should declare this as a private constant (see http://www.chrisrolle.com/en/blog/private-constants-in-ruby), and have the whole file name. This works but I think is a little less clear.

zjam1: If we create a private constant then it's not possible to reference the constant anymore within the base class. 

Also there are 3 files names that are slightly different from each other. If you want the whole file name then we are going to need 3 different names.

PR #284, opened 2018-07-30
made some simple changes, making phantom js a link

thomas-cherry:
I've been asked on multiple occasions to test if we can commit from an external fork, this is a test. I created a new account which has no association with anyone, made a change in my fork and now I'm requesting that it be merged. I don't think we need to actually merge it, just see if it is in fact getting through.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ To insert a sample draft with every field completed:
 
 ### OpenSSL Issue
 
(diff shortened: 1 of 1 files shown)


jceaser:
the thomas-cherry account is one created by me today and is not associated with any NASA project or any project at all. It is, as I said, just created.

PR #285, opened 2018-08-02
MMT-1375 prevent wrapping for dropdown items

zjam1:
In specific instances the dropdown items would wrap around and display the items incorrectly.

### DIFF
--- a/app/views/shared/_user_info_dropdown.html.erb
+++ b/app/views/shared/_user_info_dropdown.html.erb
@@ -3,11 +3,11 @@
     <!-- User Info Dropdown Menu -->
     <div id="login-info" class="login-info-dropdown">
       <% unless hide_provider_context %>
-        <%= link_to 'Change Provider', '#provider-context', class: 'display-modal', id: 'change-context' %>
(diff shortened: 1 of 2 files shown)


app/views/shared/_user_info_dropdown.html.erb
```diff
@@ -3,11 +3,11 @@
     <!-- User Info Dropdown Menu -->
     <div id="login-info" class="login-info-dropdown">
```
charleshuang80: would the class you are adding work for the individual items if you added it to this parent `div`, instead of adding it to each of the items?

jceaser: that sounds right, but I think it has to be tried. looks like these are 'a' tags, they should inherit from a parent div, but maybe they have styles that override?

zjam1: I tried that and it look likes none of the characters wrap if the class is in the `div`. So ` Change Provider User's Guide Logout` are all on a single line.

charleshuang80:
Approving, pending a green build.

PR #286, opened 2018-08-06
MMT-36 create custom collections templates

zjam1:
Users are now are able to create collection templates and then use these templates to create new drafts that have fields populated from these templates

### DIFF
--- a/app/views/collection_drafts/show.html.erb
+++ b/app/views/collection_drafts/show.html.erb
@@ -37,7 +37,26 @@
       <% else %>
         <%= link_to 'Publish Collection Draft', publish_collection_draft_path(get_resource), method: :post, class: 'eui-btn--blue spinner' %>
       <% end %>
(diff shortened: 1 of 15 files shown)


app/views/collection_drafts/show.html.erb
```diff
@@ -37,7 +37,27 @@
       <% else %>
         <%= link_to 'Publish Collection Draft', publish_collection_draft_path(get_resource), method: :post, class: 'eui-btn--blue spinner' %>
       <% end %>
-      <%= link_to 'Delete Collection Draft', "#delete-draft-modal", class: 'display-modal' %>
+      <%= link_to 'Save As Template', '#save-as-template-modal', class: 'display-modal' if Rails.configuration.templates_enabled %>
+      <%= '|' if Rails.configuration.templates_enabled %>
```
zjam1: Thomas suggested this but there might be css that already does this

PR #291, opened 2018-08-14
MMT-1551 fix bug trying to log partial launchpad info when logging in with urs

charleshuang80:


jceaser:
chris is not in this week, I think your good to go with the merge.

PR #295, opened 2018-09-06
MMT-1374 Fix issue with display properly error message

htranho:
when delete collection permission without permission


spec/features/permissions/create_update_delete_permissions_spec.rb
```diff
@@ -255,4 +255,31 @@
...
+      visit permission_path(@collection_permission_to_delete_with_failure['concept_id'])
+
+      click_on 'Delete'
+      click_on 'Yes'
+
+      wait_for_cmr
+    end
+
+    it 'not successfully deletes the collection permission' do
+      #save_screenshot('/usr/local/server/cmr/cmr-git/mmt/tmp/capybara/screenshot_name.png')
```
charleshuang80: we don't need these comments anymore, right?

PR #300, opened 2018-10-09
MMT-1589: display the correct number of associated collections

htranho:


spec/features/services/list_collection_associations_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    # publish the service to test
+    @service_ingest_response, _concept_response = publish_service_draft(provider_id: 'LARC', native_id: native_id)
+
+    # get all provider collections
+    @provider_collections = get_collections_by_provider('LARC')
+    #puts 'provider collections: ' + @provider_collections.to_s
+
+    # gets number of collections to be assigned to the service
+    @service_collections = @provider_collections[0..25]
+    #puts 'service collections: ' + @service_collections.to_s
```
charleshuang80: I think we need to remove the logging `puts` here and a bit above

PR #306, opened 2018-11-01
MMT-1591: improve logging of errors, including for searching of metadata records

charleshuang80:


app/controllers/data_quality_summaries_controller.rb
```diff
@@ -59,6 +59,7 @@ def create
     response = echo_client.create_data_quality_summary_definition(token_with_client_id, @summary)
 
     if response.error?
+      Rails.logger.error("Create Data Quality Summary Error: #{response.inspect}")
       flash[:error] = response.error_message
 
       render :new
```
cgokey: you don't want clean_inspect here?

app/controllers/provider_orders_controller.rb
```diff
@@ -27,6 +27,7 @@ def destroy
     if result.success?
       flash[:success] = "#{'Item'.pluralize(catalog_items.count)} successfully #{method}"
     else
+      Rails.logger.error("#{method.capitalize} Provider Order Error: #{result.inspect}")
```
cgokey: likewise here?

app/controllers/collection_drafts_controller.rb
```diff
@@ -147,8 +147,8 @@ def publish
       redirect_to collection_path(concept_id, revision_id: revision_id), flash: { success: I18n.t("controllers.draft.#{plural_resource_name}.publish.flash.success") }
     else
       # Log error message
-      Rails.logger.error("Ingest Metadata Error: #{ingested_response.inspect}")
-      Rails.logger.info("User #{current_user.urs_uid} attempted to ingest draft #{get_resource.entry_title} in provider #{current_user.provider_id} but encountered an error.")
+      Rails.logger.error("Ingest Collection Metadata Error: #{ingested_response.clean_inspect}")
```
jceaser: nice catch

app/controllers/provider_orders_controller.rb
charleshuang80: same answer, this is an `echo_client` response

charleshuang80:
Please see my response to Chris' comment. for `echo_client` responses, there is no token logged and I didn't create a `clean_inspect` method for that response class. Also when it is a `response.body.inspect` that is only the body so won't include the request headers (and won't have a `clean_inspect` method.

PR #315, opened 2018-12-12
Mmt 1611 Reduce badge size and font size

htranho:
Change according to Alicia's feedback


jceaser:
adding Charles, not because I expect him to review this now, but this way he get's an email and can comment after the fact if there is something he does not like.

PR #319, opened 2019-01-15
MMT-1625: migrate MMT tests to use Selenium/Headless Chrome (from Poltergeist/PhantomJS)

charleshuang80:
* use Headless Chrome instead of PhantomJS
* change driver to Selenium from Poltergeist
* modify tests to work with new driver and headless browser


spec/features/chooser/chooser_to_filter_spec.rb
```diff
@@ -52,7 +50,8 @@
 
       it 'hides values that do not match the search text' do
         within '#catalog_item_guid_toList' do
-          expect(page).to have_css('option', count: 2)
+          # Selenium is picking up the '.is-hidden' element(s) for some reason, so we need to specify not having the class
```
jceaser: like the comment

spec/features/collection_drafts/forms/related_urls_form_spec.rb
```diff
@@ -15,7 +11,7 @@
         click_on 'Related URLs', match: :first
       end
 
-      click_on 'Expand All'
```
jceaser: why did this not work?

spec/features/order_options/deprecate_order_options_spec.rb
```diff
@@ -10,11 +7,36 @@
       end
     end
 
-    it 'lists available order options with \'Deprecate\' links.' do
-      page.all('.order-options-table tr') do |tr|
-        within tr do
-          expect(tr).to have_link 'Deprecate'
-        end
+    it "lists available order options with 'Deprecate' links." do
+      within '.order-options-table tbody tr:nth-child(1)' do
```
jceaser: what happened here?

spec/features/permissions/permissions_permissions_spec.rb
```diff
@@ -36,9 +34,10 @@
       end
 
       it 'displays the correct message' do
-        expect(page).to have_content('You don\'t have the appropriate permissions to show this permission.')
+        expect(page).to have_content("You don't have the appropriate permissions to show this permission.")
```
jceaser: why did you need to change from an encoded quote to a "hard quote"?

spec/rails_helper.rb
```diff
@@ -15,20 +15,63 @@
...
+    loggingPrefs: { browser: 'ALL', client: 'ALL', driver: 'ALL', server: 'ALL' }
+  )
+
+  # disable-gpu option is temporarily necessary, possibly only for Windows
+  # https://developers.google.com/web/updates/2017/04/headless-chrome#cli
+  # no-sandbox was necessary for another application's Docker container for CI/CD
+  # https://about.gitlab.com/2017/12/19/moving-to-headless-chrome/
+  # https://developers.google.com/web/updates/2017/04/headless-chrome#faq
+  options = Selenium::WebDriver::Chrome::Options.new(
+    args: %w[headless disable-gpu no-sandbox --window-size=1440,900]
```
jceaser: I assume this is what we would change if we want taller screen shots?

spec/features/bulk_updates/bulk_updates_data_centers_spec.rb
```diff
@@ -37,9 +35,12 @@
     it 'displays the preview information', bulk_update_step_1: true do
       expect(page).to have_content('Preview of New MMT_2 Bulk Update')
 
-      expect(page).to have_content("Name #{bulk_update_name}")
-      expect(page).to have_content('Field to Update Data Centers')
-      expect(page).to have_content('Update Type Find And Remove')
+      expect(page).to have_content('Name')
```
jceaser: talked on slack, but putting it here, we should explore, shortly, if this is the best way.

spec/rails_helper.rb
charleshuang80: yes

spec/features/bulk_updates/bulk_updates_data_centers_spec.rb
charleshuang80: I can't post the screenshot here, but if you look at the page, each of these text pieces are actually on separate lines, so I thought it made sense to actually separate them in the test. (even though they were on separate lines, Poltergeist treated the whitespace differently so it accepted the text on one line)

spec/features/bulk_updates/bulk_updates_data_centers_spec.rb
```diff
@@ -331,7 +347,7 @@
                 expect(page).to have_no_content('ESA/ED')
 
                 expect(page).to have_content('AARHUS-HYDRO', count: 2)
-                expect(page).to have_content('Hydrogeophysics Group, Aarhus University ', count: 2)
+                expect(page).to have_content('Hydrogeophysics Group, Aarhus University', count: 2)
```
cgokey: Why did the old version have a space at the end?   Just a typo or was it on purpose?

spec/features/collection_drafts/forms/add_another_button_spec.rb
```diff
@@ -24,7 +24,7 @@
 
     context 'when clicking the add another button' do
       before do
-        find('button', text: 'Add another Related URL').trigger('click')
+        find('button', text: 'Add another Related URL').click
       end
```
cgokey: Did the previous way not work?

spec/features/collection_drafts/forms/additional_attribute_type_spec.rb
```diff
@@ -57,6 +55,7 @@
         select 'Integer', from: 'Data Type'
         fill_in 'Parameter Range Begin', with: '5'
         fill_in 'Parameter Range End', with: '5'
+        find('body').click
```
cgokey: Was clicking the body not part of the last test?

spec/features/collection_drafts/forms/data_contacts_form/data_contacts_form_data_center_contacts_spec.rb
```diff
@@ -131,7 +133,7 @@
             it 'still displays the Data Center on the preview page' do
               within '.data-centers-cards' do
                 expect(page).to have_content(data_center_short_name)
-                expect(page).to have_content(data_center_long_name)
+                expect(page).to have_content(data_center_long_name_without_space)
```
cgokey: Why do you test for both here?

spec/features/collection_drafts/forms/validation_spec.rb
```diff
@@ -522,16 +516,17 @@
 
       within '#draft_spatial_extent_horizontal_spatial_domain_geometry_bounding_rectangles_0' do
         fill_in 'North', with: 'asdf'
-        fill_in 'West', with: ''
-        fill_in 'East', with: ''
-        fill_in 'South', with: ''
+        find('#draft_spatial_extent_horizontal_spatial_domain_geometry_bounding_rectangles_0_west_bounding_coordinate').click
+        find('#draft_spatial_extent_horizontal_spatial_domain_geometry_bounding_rectangles_0_east_bounding_coordinate').click
+        find('#draft_spatial_extent_horizontal_spatial_domain_geometry_bounding_rectangles_0_south_bounding_coordinate').click
```
cgokey: The prior way of doing it no longer works?

spec/features/bulk_updates/bulk_updates_data_centers_spec.rb
charleshuang80: The KMS (both in the local CMR and the other environments) has this trailing whitespace in the long name keyword. I brought this up with @jceaser, you can ask him about tickets and what the plan is for it. What it means for our tests is that previously it was rendered, but with the change in webdriver and headless browser, it's not rendered or picked up anymore so needs to be removed. However, when it is in an input field, because it's being populated from the controlled keywords, the space is still there.

spec/features/collection_drafts/forms/additional_attribute_type_spec.rb
charleshuang80: As I mentioned, previously using `fill_in` would also trigger a blur, which would allow validation or other JS activity we need to occur. Somehow with the change over, that is not happening so I need to click on something to trigger the blur.

spec/features/collection_drafts/forms/add_another_button_spec.rb
charleshuang80: see the link I posted in the chat, it talks about this change being necessary.

spec/features/bulk_updates/bulk_updates_location_keywords_spec.rb
```diff
@@ -38,11 +36,15 @@
...
+      expect(page).to have_content('Name')
+      expect(page).to have_content(bulk_update_name)
+      expect(page).to have_content('Field to Update')
+      expect(page).to have_content('Location Keywords')
+      expect(page).to have_content('Update Type')
+      expect(page).to have_content('Find And Remove')
+
       within '.find-values-preview' do
-        expect(page).to have_content('CATEGORY: ANY VALUEARCTIC')
+        expect(page).to have_content('CATEGORY: ANY VALUE ARCTIC', normalize_ws: true)
```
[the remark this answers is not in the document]
charleshuang80: https://www.rubydoc.info/github/jnicklas/capybara/Capybara/Node/Matchers#has_text%3F-instance_method, we are using the `:normalize_ws` option, which says 'When true replace all whitespace with standard spaces and collapse consecutive whitespace to a single space'.

Again, there is a difference of how science and location keywords get rendered between the previous webdriver/headless browser and the new one. Now the spaces are there (as you can see previously the space was not there). The styling for these keywords has not changed and still looks good, I'm just adding the normalize_ws to make sure all the different whitespace characters get treated the same because it will all try to render on one line.

spec/features/collection_drafts/forms/data_contacts_form/data_contacts_form_data_center_contacts_spec.rb
charleshuang80: see comment above about the whitespace in the KMS long name and how it's treated in an input field and otherwise.

spec/features/collection_drafts/forms/validation_spec.rb
charleshuang80: Also as I mentioned, previously when filling in with an empty string that served to tell the page that the input field was visited and could trigger validation. Now that is not happening, so the click serves to 'visit' the field but not enter anything.

spec/rails_helper.rb
```diff
@@ -55,6 +98,8 @@
     puts "Rails Version: #{Rails.version}"
   end
 
+  Capybara.default_max_wait_time = (ENV['CAPYBARA_WAIT_TIME'] || 15).to_i
```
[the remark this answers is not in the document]
jceaser: the chance that it is ms and not sec I think is pretty low, if you also feel that it is seconds, that is good enough for me.

I just want a general idea of what is going on in tests: 3000 (made up, from "memory") test, all waiting (at most) 15 seconds = 12 hours (I know a crazy/unrealistic amount of time, but that would be worst case, nothing worked time which helps me understand why at least we would have a 1 hour test time).

spec/features/bulk_updates/bulk_updates_data_centers_spec.rb
jceaser: I can see that, and I can see how different algorithms would treat 'space newline space' different then 'space space space' when dealing with HTML, with the old way we knew that "Field to Update" was followed by "Instruments" with only white space between them, where as here we are assuming that. But if normalize white space did not resolve this, then we should just keep it in mind that it could be faked or easily confused and move on.

So my question is, did the normalize white space parameter that I saw used in other cases not work here?

spec/features/bulk_updates/bulk_updates_location_keywords_spec.rb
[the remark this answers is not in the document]
jceaser: yea, there seams to be a different opinion between these tools on what to do with white space. I like the new code better.

spec/features/collection_drafts/forms/add_another_button_spec.rb
jceaser: web browsers are very funny with how they handle events, at least if your an old hat like me and did it all by hand. I can see how APIs would not always get it right so switching methods I believe would work.

spec/features/bulk_updates/bulk_updates_location_keywords_spec.rb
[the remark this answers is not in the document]
charleshuang80: right but in this case, Poltergeist/PhantomJS actually didn't even show a space in between the two values.

spec/features/collection_drafts/forms/add_another_button_spec.rb
charleshuang80: there's a better explanation [here](https://about.gitlab.com/2017/12/19/moving-to-headless-chrome/), apparently `trigger` was a work around that Poltergeist provided (because the rendering was less accurate).

PR #321, opened 2019-02-06
MMT-1740: add feature toggle for uvg, turn on for dev, test, and sit only

charleshuang80:


config/environments/test.rb
```diff
@@ -62,11 +62,14 @@
   # Feature Toggle for UMM-S
   config.umm_s_enabled = true
 
+  # Feature Toggle for UMM-Var Generation
+  config.uvg_enabled = true
```
jceaser: this is what bamboo uses I am assuming, we on the product side will need to keep this in mind.

config/environments/test.rb
```diff
@@ -62,11 +62,14 @@
...
+
   config.cmr_env = 'sit'
   config.echo_env = 'sit'
   config.urs_register_url = 'https://sit.urs.earthdata.nasa.gov/users/new'
 
   config.middleware.use RackSessionAccess::Middleware
 
-  config.colorize_logging = false
+  # config.colorize_logging = false
```
jceaser: a non-blocking "why" (will approve but I just want to know what's up here)

config/environments/test.rb
charleshuang80: Yes, this is what will be used for our tests locally and in Bamboo

config/environments/test.rb
charleshuang80: So I hadn't actually checked our test logs until fixing all the headless chrome tests. Those logs aren't colored like our dev logs and I hadn't realized this is where that was controlled, so I thought it would be nice to see if having them be colored would be helpful. I would like to keep this commented for a little while to determine that.

PR #322, opened 2019-02-11
MMT-1647 Updating MMT to add GTM html code

rosycordova:
Update html to add Google Tag Manager (GTM) code to head and body tags.


config/environments/sit.rb
```diff
@@ -99,6 +99,7 @@
   config.cmr_env = 'sit'
   config.echo_env = 'sit'
   config.analytics_id = 'UA-62340125-5'
+  config.tag_manager_id = 'GTM-TESTTEST'
```
jceaser: what was the logic for not using the ID for SIT and UAT? I see why to not use it for development and test. Not yet suggesting a change, just want to ask what lead us to here.

app/views/shared/_google_tag_manager_body.html.erb
```diff
@@ -0,0 +1,7 @@
+<!-- Google Tag Manager (noscript) -->
+<noscript>
+<iframe src="https://www.googletagmanager.com/ns.html?id=<%= Rails.configuration.tag_manager_id %>"
```
charleshuang80: I feel like we should wrap both the `body` and `head` partial contents within an if statement similar to how we had it for the `analytics_id`:
`<% if Rails.configuration.respond_to?(:tag_manager_id) -%>`
That way we wouldn't need to add fake ids, and not need to add them for development or test.

config/environments/sit.rb
rosycordova: Peggy from EMS replied that same ID can be used in all environments.  Updated code & resubmitted.

app/views/shared/_google_tag_manager_body.html.erb
rosycordova: added if statement & resubmitted.

PR #325, opened 2019-03-08
MMT-1751: allow sending UVG augmentation request

charleshuang80:
* add augmentation selection and pages
* add calls to augmentation endpoints
* add controller actions and update views
* add and update tests

### DIFF
--- a/app/views/variable_generation_processes/edit.html.erb
+++ b/app/views/variable_generation_processes/edit.html.erb
@@ -4,8 +4,7 @@
 
     <fieldset>
       <h4>Augmentation</h4>
-      <!-- figure out a box -->
-      <%= form_tag(variable_generation_process_path, method: :put, class: 'something') %>
+      <%= form_tag(variable_generation_process_path, method: :put) %>
 
       <!-- These are currently radio buttons, but eventually should be changed
         to be checkboxes, so the user can choose all the augmentations they want
@@ -42,7 +41,6 @@
         <%= hidden_field_tag 'variables_json', @variables_string %>
 
         <%= submit_tag 'Submit', class: 'eui-btn--blue', id: 'submit-for-augmentation' %>
-      <%# end %>
     </fieldset>
   </section>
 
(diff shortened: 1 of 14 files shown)


app/views/variable_generation_processes/edit.html.erb
```diff
@@ -0,0 +1,49 @@
+<div class="row row-content">
+  <section>
+    <h2>Variable Generation for <%= current_user.provider_id %></h2>
+
+    <fieldset>
+      <h4>Augmentation</h4>
+      <!-- figure out a box -->
+      <%= form_tag(variable_generation_process_path, method: :put, class: 'something') %>
```
cgokey: something?

charleshuang80: hah. i'll remove this. It doesn't actually need a class, but I thought not having one was causing an error.

lib/cmr/uvg_client.rb
```diff
@@ -9,5 +10,79 @@ def uvg_generate(params)
...
+            else
+              '/augment/definitions_stub'
+            end
+
+      post(url, params.to_json)
+    end
+
+    def uvg_augment_keywords_stub(params)
+      url = if Rails.env.development? || Rails.env.test?
+              'http://localhost:4000/augment/keywords_stub'
```
cgokey: I see this code sprinkled all through mmt, if development || test do this else do that.   Why not just declare a host variable in env config to use instead?

charleshuang80: I'm not sure I follow what you're suggesting - can you give an example?

cgokey: I'm not suggesting change it (and there is probably reasons why it isn't this way), just wondering why this code can't be something like:
```
	      url = if Rails.env.development? || Rails.env.test?
	              'http://localhost:4000/generate_stub'
	            else
	      post(url, params.to_json)
	    end
```
```
post(Rails.application.config.generate_var_url, params.to_json)
```
To me seems like having specific service endpoints might be a good thing to have in config.

charleshuang80: I think we can discuss this further, but I'm not going to address it in this PR.

There are a lot of endpoints that we call, and I'm not sure how much of a benefit it would be to maintain a separate list of them, especially if they might be different between dev/test and the other environments. A big reason we have these is for the local cmr endpoints, which aren't always just a clean addition of some localhost port at the front.

PR #327, opened 2019-03-12
MMT-1756 - UMM-V version 1.3 -> 1.4

cgokey:
Upgrade from umm-var 1.3 to 1.4


lib/assets/schemas/variables/umm-var-form.json
```diff
@@ -357,17 +357,32 @@
...
-                  "type": "section",
-                  "htmlClass": "col-6",
-                  "items": [{"key": "/AvgCompressionRateASCII", "field": true, "label": "Avg Compression Rate ASCII"}]
-                },
-                {
-                  "type": "section",
-                  "htmlClass": "col-6",
-                  "items": [{"key": "/AvgCompressionRateNetCDF4", "field": true, "label": "Avg Compression Rate NetCDF4"}]
+                  "type": "UmmMultiItems",
+                  "label": "Average Compression Informations",
```
cgokey: This should not be plural

lib/assets/schemas/variables/umm-var-json-schema.json
```diff
@@ -29,15 +29,15 @@
           "items": {
             "type": "number"
           },
-          "minItems": 0
+          "minItems": 2
```
cgokey: Need to confirm this Eric it is meant to be there and check related tests.

lib/assets/schemas/variables/umm-var-json-schema.json
```diff
@@ -29,15 +29,15 @@
...
         },
         "LonRange": {
           "description": "The LonRange consists of an index range for longitude.",
           "type": "array",
           "items": {
             "type": "number"
           },
-          "minItems": 0
+          "minItems": 2
```
cgokey: Likewise check Erich on this one.

spec/features/variable_drafts/forms/validation_spec.rb
```diff
@@ -343,4 +339,25 @@
...
+        fill_in 'variable_draft_draft_characteristics_index_ranges_lon_range_0', with: '1'
+        within '.nav-top' do
+          click_on 'Save'
+        end
+      end
+
+      it 'displays validation error messages' do
+        expect(page).to have_css('#variable_draft_draft_characteristics_index_ranges_lat_range_error', text: 'Lat Range has too few items')
+        expect(page).to have_css('#variable_draft_draft_characteristics_index_ranges_lon_range_error', text: 'Lon Range has too few items')
+      end
```
charleshuang80: I feel like we should add another item to one or both ranges, to test that the error goes away.

app/assets/javascripts/validation.coffee
```diff
@@ -349,6 +347,13 @@ $(document).ready ->
               params: {}
             errors.push newError
 
+  validateLatLonRanges = (errors) ->
+    if $("[id^=variable_draft_draft_characteristics_index_ranges_lat_range_]").length > 0
```
charleshuang80: It looks like all you are doing is visiting the field if the parent id exists on the page. So those fields will be visited every time the page is validated, right? We usually only validate a field if it has already been visited, but here it seems the field will be marked as visited even if the user hasn't visited it.

spec/features/variable_drafts/forms/validation_spec.rb
htranho: Yes, will add that test

app/assets/javascripts/validation.coffee
htranho: Added fix in method validateLatLonRanges

app/assets/javascripts/validation.coffee
```diff
@@ -349,6 +347,15 @@ $(document).ready ->
               params: {}
             errors.push newError
 
+  validateLatLonRanges = ->
```
charleshuang80: do you think this can be done in the `visitField` function on line 536?

htranho: Now moved to visitField

PR #330, opened 2019-03-20
MMT-1752: add ability to save UVG generated variables as variable drafts

charleshuang80:
* add mass import gem
* add saving as variable drafts functionality utilizing mass import
* update buttons placement and style
* update helper methods
* attempt to resolve timeout issue
* add tests for saving drafts

### DIFF
--- a/app/controllers/variable_generation_processes_controller.rb
+++ b/app/controllers/variable_generation_processes_controller.rb
@@ -1,6 +1,7 @@
 # :nodoc:
 class VariableGenerationProcessesController < ManageVariablesController
-  before_action :uvg_enabled?
+  before_action :uvg_enabled?, :log_before_timestamp
+  after_action :log_after_timestamp
 
   def create
     payload = {}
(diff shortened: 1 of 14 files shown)


charleshuang80:
turn off feature toggle in SIT

app/controllers/variable_generation_processes_controller.rb
```diff
@@ -65,4 +67,72 @@ def update
...
+                     id_to_start = first_var.id + 1
+
+                     mass_import_result = import_remaining_variable_drafts(generated_variables, id_to_start)
+                     Rails.logger.info "need to save this somehow. mass_import_result: #{mass_import_result.inspect}"
+                   end
+
+    [mass_import_result, drafts_saved]
+  end
+
+  def create_single_variable_draft(generated_var)
```
cgokey: Why does the first variable not have entry id, short name, and native id just curious?   can't these fields be nil so you don't have to have a separate method for the first and all the others?

lib/cmr/response.rb
```diff
@@ -9,7 +9,7 @@ def initialize(faradayResponse)
     end
 
     def error?
-      status >= 400 || (body.is_a?(Hash) && body['errors'])
+      status >= 400 || (body.is_a?(Hash) && body['errors'] && !body['errors'].blank?)
```
cgokey: I think you can say: ```status >= 400 || (body.is_a?(Hash) && !body['errors'].blank?``` 
as ```nil.blank? == true```

cgokey: Actually scratch that.

app/controllers/variable_generation_processes_controller.rb
htranho: Just curios why the process of saving variables in two steps: the first single one and then the rest.

charleshuang80: I'm using the gem to do a mass import/save of all the variables, so there are fewer SQL transactions (because we can have up to 3000 records on the high end). However, because of the way the gem works, the before and after hooks are not run properly (specifically https://github.com/nasa/mmt/blob/5819e06be1f17de8a0719435ae7c332030fc3b65/app/models/draft.rb#L9). So I need to first create one the regular way, and use the id for the `native_id`s of the other drafts.

PR #333, opened 2019-04-02
MMT-1744: UVG Generate request works with live endpoint in development

charleshuang80:
* add collection and granule to local cmr that will work with UVG API generate endpoint
* UVG generate request to API in development works
* update tests to reflect use non-stubbed endpoint


lib/test_cmr/test_data.json
```diff
@@ -286,4 +286,10 @@
   "collection": "https://cmr.earthdata.nasa.gov/search/concepts/C115003861-NSIDC_ECS/76",
   "ingest_count": 1,
   "granule": null
+}, {
+  // 57
```
jceaser: what is this?

PR #334, opened 2019-04-04
MMT-1743: allow UVG augmentation of science keywords request

charleshuang80:
* allow augmentation request for science keywords
* display correct statistics
* update tests and stubs
* improve handling of save as drafts failure

### DIFF
--- a/app/controllers/variable_generation_processes_controller.rb
+++ b/app/controllers/variable_generation_processes_controller.rb
@@ -50,8 +50,12 @@ class VariableGenerationProcessesController < ManageVariablesController
     uvg_parameters = { payload: payload }
 
     @operation = "augment_#{augmentation_type}"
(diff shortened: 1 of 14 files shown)


app/controllers/variable_generation_processes_controller.rb
```diff
@@ -100,17 +107,21 @@ def save_variable_drafts
 
   def import_variable_drafts(generated_variables)
     total_record_count = generated_variables.count
-    first_generated_var = generated_variables.shift
+    # in case of failure, we need to preserve the original variables
+    vars_to_save = generated_variables.dup
```
cgokey: I think by default dup is shallow, so I assume these are only 1 level deep?

charleshuang80: We only care about the top level because the variables are not being manipulated in any way. The only thing that is happening is that we are dropping one of the variables when we try to import them, and want to preserve the original array in case of error.

PR #335, opened 2019-04-04
MMT-1776 - is a good year... added support for umm-var 1.5 namely acquisition source name.

cgokey:


spec/features/variable_drafts/preview/invalid/acquisition_source_name_preview_spec.rb
```diff
@@ -0,0 +1,39 @@
+describe 'Invalid Variable Draft Acquisition Source Name Preview', :js => true do
```
charleshuang80: `js: true` is preferred (hash rocket is the older way)

spec/features/variable_drafts/preview/invalid/acquisition_source_name_preview_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  before do
+    login
+    @draft = create(:invalid_variable_draft, user: User.where(urs_uid: 'testuser').first)
```
charleshuang80: for the tests in this folder, I thought it was trying to test if there was invalid data in the draft. Are you able to add an invalid value in the draft factory being used here?

spec/features/variable_drafts/preview/valid/acquisition_source_name_preview_spec.rb
```diff
@@ -0,0 +1,39 @@
+describe 'Invalid Variable Draft Acquisition Source Name Preview', :js => true do
```
charleshuang80: `js: true` is preferred

charleshuang80: this is in the `valid` folder, but text starts with 'Invalid Variable Draft'

PR #338, opened 2019-04-19
MMT-1775 Set sub field preview circle depends on sibling

htranho:


jceaser:
as no one is here, I think this is good to go out

PR #340, opened 2019-05-02
Feature/mmt-1811 - cache calls to CMR when looking up users

jceaser:


lib/echo/order.rb
```diff
@@ -86,6 +86,12 @@ def billing_address
 
     private
 
+    def cached_owner
+      Rails.cache.fetch("owners.#{owner_guid}", expires_in: 15.minutes) do
+        @client.get_user_names(@token, owner_guid).parsed_body
```
charleshuang80: I think we should add logging just before and after this call.
before, we should put: `Rails.logger.info "Starting get_user_names request sent at Time #{Time.now.to_i} with owner_guid #{owner_guid}"`
and after: `Rails.logger.info "Response from get_user_names request received at Time #{Time.now.to_i}"`

PR #341, opened 2019-05-03
MMT-1807: add additional logging for better information on Tracking Orders

charleshuang80:


app/controllers/errors_controller.rb
```diff
@@ -9,10 +9,20 @@ class ErrorsController < ApplicationController
...
   def not_found
+    Rails.logger.error "User #{current_user_id} encountered a 404 not found error"
+
     render(status: 404)
   end
 
   def internal_server_error
-    render(status: 500)
+    logger.tagged("#{current_user_id}") do
+      Rails.logger.error "User #{current_user_id} encountered an error at Time #{Time.now.to_i} with request method #{request.request_method} to path #{request.original_fullpath}"
```
cgokey: Do you need {current_user_id} in line 19 if the logs are tagged with {current_user_id} ?

charleshuang80: You're right, I think it is redundant, but I'm also trying to be consistent with other error messages we have.

PR #343, opened 2019-05-08
MMT-1815 - adding Top Hat configuration

jceaser:
Created a configuration for using SIT and UAT versions of TopHat in SIT and UAT


config/environments/test.rb
```diff
@@ -69,6 +69,8 @@
   config.echo_env = 'sit'
   config.urs_register_url = 'https://sit.urs.earthdata.nasa.gov/users/new'
 
+  config.tophat_url = 'https://cdn.sit.earthdata.nasa.gov/tophat2/tophat2.js'
```
charleshuang80: should we keep this in here if it is never rendered in test?

PR #344, opened 2019-05-10
MMT-1821  logic for app layer to handle timeouts before nginx.   Also pre-caches user guids to improve performance.

cgokey:
By pre-caching owner guids in application layer, we cut down on thousands of soap calls.    MMT-1407 related ticket now finishes in 200+ seconds, < 5 minute timeout now.

### DIFF
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -46,13 +46,14 @@ class OrdersController < ManageCmrController
 
     end
   rescue Faraday::Error::TimeoutError
-    flash[:alert] = 'The order request timed out retrieving results.  Perhaps limit your criteria to a smaller time frame.'
+    flash.now[:alert] = 'The order request timed out retrieving results.  Limit your search criteria and try again.'
     render :index
   end
 
   private
 
   # sets up initial values to track time spent issuing faraday requests.
+  # initial budget of time allowed to complete request is 270 seconds (300-30 see below)
   # echo_client.timeout as of 5/16/19 is 300 seconds, subtracting 30 seconds for any potential processing,
   # the rest of the remaining time will be used for faraday requests.
   def init_time_tracking_variables
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -19,15 +19,9 @@ class OrdersController < ManageCmrController
     end
   end
 
-  def time_left
-    return @timeout_duration - (Time.new - @request_start)
-  end
-
   def search
     Rails.logger.info("starting request - #{request.uuid} timeout=#{echo_client.timeout}")
-    @timeout_duration = echo_client.timeout - 30
-    @request_start = Time.new
-
+    init_time_tracking_variables
     guids = determine_order_guids
 
     time "search #{guids.class == Array ? guids.count : 1} orders" do
@@ -49,6 +43,7 @@ class OrdersController < ManageCmrController
         @orders.sort_by!(&:created_date)
         render :search
       end
+
     end
   rescue Faraday::Error::TimeoutError
     flash[:alert] = 'The order request timed out retrieving results.  Perhaps limit your criteria to a smaller time frame.'
@@ -57,6 +52,22 @@ class OrdersController < ManageCmrController
 
   private
 
+  # sets up initial values to track time spent issuing faraday requests.
+  # echo_client.timeout as of 5/16/19 is 300 seconds, subtracting 30 seconds for any potential processing,
+  # the rest of the remaining time will be used for faraday requests.
+  def init_time_tracking_variables
+    @timeout_duration = echo_client.timeout - 30
+    @request_start = Time.new
+  end
+
+
+  # returns the time remaining for the request to complete, used as a timeout value for faraday connections.
+  def time_left
+    return @timeout_duration - (Time.new - @request_start)
+  end
+
+  # this speeds things up dramatically by collecting all the owner guids and performing 1 query to grab ALL the
+  # user info details, it will then cache them and this cache is used further down in the processing.
   def precache_owner_guids
     time "time to retreive guids" do
       owner_guids = []
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -1,6 +1,6 @@
 # :nodoc:
 class OrdersController < ManageCmrController
-  include ApplicationHelper
+  include LogTimeSpentHelper
   add_breadcrumb 'Track Orders', :orders_path
 
   def index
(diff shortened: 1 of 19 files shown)


app/controllers/orders_controller.rb
```diff
@@ -21,22 +21,49 @@ def show
...
 
   def search
-    logger.tagged("#{current_user.urs_uid} #{controller_name}_controller") do
-      @orders = Echo::Orders.new(client: echo_client, echo_provider_token: echo_provider_token, guids: determine_order_guids).orders
+    Timeout.timeout(Rails.configuration.orders_timeout) do
+      logger.tagged("#{current_user.urs_uid} #{controller_name}_controller") do
+        request_start = Time.new
+        start = Time.new
+        guids = determine_order_guids
+        Rails.logger.info("Elapsed time to retrieve guids for #{guids.count} orders=#{Time.new - start}")
```
cgokey: It occurred to me it is doing this twice, it doesn't take very long, but needs to be moved up above line 25

app/controllers/orders_controller.rb
```diff
@@ -21,22 +21,49 @@ def show
   end
 
   def search
-    logger.tagged("#{current_user.urs_uid} #{controller_name}_controller") do
-      @orders = Echo::Orders.new(client: echo_client, echo_provider_token: echo_provider_token, guids: determine_order_guids).orders
+    Timeout.timeout(Rails.configuration.orders_timeout) do
+      logger.tagged("#{current_user.urs_uid} #{controller_name}_controller") do
+        request_start = Time.new
+        start = Time.new
+        guids = determine_order_guids
```
cgokey: It occurred to me it is doing this twice, it doesn't take very long, but needs to be moved up above line 25

app/controllers/orders_controller.rb
```diff
@@ -13,56 +14,110 @@ def index
...
       render :show
     end
   end
 
   def search
-    logger.tagged("#{current_user.urs_uid} #{controller_name}_controller") do
-      @orders = Echo::Orders.new(client: echo_client, echo_provider_token: echo_provider_token, guids: determine_order_guids).orders
+    Rails.logger.info("starting request - #{request.uuid} timeout=#{echo_client.timeout}")
+    init_time_tracking_variables
+    guids = determine_order_guids
```
charleshuang80: if you run `determine_order_guids` outside of the `logger.tagged` block, it won't have those logging tags.

app/controllers/orders_controller.rb
```diff
@@ -13,56 +14,110 @@ def index
...
+  rescue Faraday::Error::TimeoutError
+    flash.now[:alert] = 'The order request timed out retrieving results.  Limit your search criteria and try again.'
+    render :index
   end
 
   private
 
+  # sets up initial values to track time spent issuing faraday requests.
+  # initial budget of time allowed to complete request is 270 seconds (300-30 see below)
+  # echo_client.timeout as of 5/16/19 is 300 seconds, subtracting 30 seconds for any potential processing,
```
charleshuang80: I'm not sure if the date is necessary, but saying why it is 300s might be helpful

app/controllers/provider_orders_controller.rb
```diff
@@ -131,4 +151,19 @@ def generate_provider_order(guid)
       { 'error' => error_message }
     end
   end
+
+  # sets up initial values to track time spent issuing faraday requests.
+  # initial budget of time allowed to complete request is 270 seconds (300-30 see below)
+  # echo_client.timeout as of 5/16/19 is 300 seconds, subtracting 30 seconds for any potential processing,
+  # the rest of the remaining time will be used for faraday requests.
+  def init_time_tracking_variables
```
charleshuang80: you're declaring these time helper methods twice. it might be better to put it higher in the inheritance chain (maybe `manage_cmr_controller` so both controllers can access them. if you do that, explaining that the time tracking is for orders would be helpful.

app/helpers/application_helper.rb
```diff
@@ -22,4 +22,15 @@ def actions_table_header(actions)
 
     content_tag :th, 'Actions', colspan: (actions.count if actions.count > 1)
   end
+
+  # Simple method to time a block and log its elapsed time in logs
+  def time(name, &block)
```
charleshuang80: I still think this method name is a bit vague. I like some combination/version of these ideas: `with_time`, `time_log`, `with_time_logging`

Also, I think this method could maybe live in a different helper file. I don't think it matters if it is the only method in the helper file. I think like `orders_helper`, `time_helper`, or `timeout_helper` might work.

lib/echo/base.rb
```diff
@@ -54,5 +58,18 @@ def wrap_with_envelope(content)
         end
       end
     end
+
+    # sets the timeout used for faraday connections
+    def timeout=(value)
```
charleshuang80: We discussed and you commented about trying to test these timeout functions for the echo_client

app/controllers/provider_orders_controller.rb
```diff
@@ -1,24 +1,33 @@
...
   add_breadcrumb 'Track Orders', :orders_path
 
   def show
+    init_time_tracking_variables
     logger.tagged("#{current_user.urs_uid} #{controller_name}_controller") do
       @provider_order = generate_provider_order(params['id'])
-
       render :show
     end
+  rescue Faraday::Error::TimeoutError
```
charleshuang80: I'm not sure if the rescue for the `show` and `edit` actions here are necessary. But if we do it here, is there a reason that it's not being done for the `show` and `index` actions in `orders_controller`?

app/controllers/orders_controller.rb
cgokey: I fixed it, thanks.

app/controllers/orders_controller.rb
cgokey: Agree and done.

app/controllers/provider_orders_controller.rb
cgokey: I was one the fence about this one, as usually bad idea to have redundant code.   Agree, I moved into manage_cmr_controller.

app/helpers/application_helper.rb
cgokey: I used log_time_spent and moved it into its own helper class.

app/controllers/provider_orders_controller.rb
cgokey: I added for show, but index doesn't perform any faraday request so left as is.

lib/echo/base.rb
cgokey: I wrote a small test to ensure the timeouts are all set in each delegate.   We should discuss more tomorrow if we need any other tests.   I sorted of hacked it a bit though, as @services wasn't exposed to the public as had to add a attr_reader to expose it to the publicto ensure all the timeouts where set.  Not sure I'm crazy about doing that though.

PR #345, opened 2019-05-21
feature/MMT-1822 uses faraday timeouts instead of rails

cgokey:


spec/cmr/cmr_client_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+
+        response = cmr_client.get_collections({ 'short_name' => 'term' }, 'bad-bad-bad-token-bad-bad-bad-token:somethingsomethingso')
+      end
+    end
+  end
+
+  context 'headers' do
+    let(:basic_client) do
+      # New class with a public request method
+      Class.new(Cmr::BaseClient) do
```
charleshuang80: this `context` seems to be testing `BaseClient`, maybe we should move it into its own file `base_client_spec.rb`?

lib/echo/base.rb
```diff
@@ -17,7 +17,7 @@ def build_connection
         conn.use :instrumentation
 
         # Set timeout to 300s to match nginx timeout
-        conn.options[:timeout] = NGINX_TIMEOUT
+        conn.options[:timeout] = NGINX_TIMEOUT - 10
```
charleshuang80: in the order tracking timeout catching we start with `echo_client.timeout - 30`, so should that be adjusted slightly if we are changing this?

cgokey: Well I was thinking that code dealing with the -30 was a slightly more computation intensive process, why I did the -30.   The -10 here are defaults.

PR #349, opened 2019-05-31
MMT-1787 Fix help popup and change UI for Lat and Lon Range

htranho:


lib/json_schema_form/umm_form_elements/umm_set_multi_item.rb
```diff
@@ -14,13 +14,13 @@ def default_value
...
 
   def render_markup
-    content_tag(:div, class: "multiple simple-multiple #{form_class}") do
+    content_tag(:div, class: "multiple simple-multiple #{form_class}", id: idify_property_name) do
       indexes = options.fetch('indexes', [])
 
       values = Array.wrap(element_value)
       values = [] if values.empty?
       values += Array.new(number_of_items)
-      values[0..2].each_with_index do |value, index|
```
htranho: Check if StreetAddress still works

htranho: StreetAdress UI works

PR #352, opened 2019-06-11
Mmt 1817

ryanmiller-1:
Added an app/views/drafts/show.json.erb file that responds with the json of a draft when x_drafts/id.json url is used.  A test for each of collections, variables, and services are included in their x_draft_creation_spec file.  Changes passed all expected tests (inherited and novel) locally.


spec/features/collection_drafts/collection_draft_creation_spec.rb
```diff
@@ -54,6 +54,16 @@
             expect(page).to have_content('123')
           end
         end
+
+        context 'when accessing a collection draft\'s json' do
+          before do
+            visit collection_draft_path(CollectionDraft.first, :format => 'json')
```
charleshuang80: the preferred way to do the hash key that is a symbol would be `format: 'json'`

spec/features/service_drafts/service_draft_creation_spec.rb
```diff
@@ -37,6 +37,16 @@
       it 'displays a confirmation message' do
         expect(page).to have_content('Service Draft Created Successfully!')
       end
+
+      context 'when accessing a service draft\'s json' do
+        before do
+          visit service_draft_path(ServiceDraft.first, :format => 'json')
```
charleshuang80: same comment

spec/features/variable_drafts/variable_draft_creation_spec.rb
```diff
@@ -35,6 +35,16 @@
       it 'displays a confirmation message' do
         expect(page).to have_content('Variable Draft Created Successfully!')
       end
+
+      context 'when accessing a variable draft\'s json' do
+        before do
+          visit variable_draft_path(VariableDraft.first, :format => 'json')
```
charleshuang80: again

app/controllers/base_drafts_controller.rb
```diff
@@ -19,6 +19,10 @@ def show
     end
 
     add_breadcrumb breadcrumb_name(get_resource.draft, resource_name), send("#{resource_name}_path", get_resource)
+
+    respond_to do |format|
```
charleshuang80: I think this is not working properly, have you tested both html and json versions with all draft types locally?
1. I think this is missing a `format.html`
2. The way this is called by the collection_drafts_controller is causing some issues that need to be resolved. If you want to discuss it at all, let me know.

PR #355, opened 2019-06-12
Mmt 1804 Add 3 new tabs to collection preview

htranho:


app/assets/javascripts/preview.coffee
```diff
@@ -104,3 +106,17 @@ $(document).ready ->
...
+
+  $('.tab-label').on 'click', (e) ->
+    $('.tab-label').removeClass('active')
+    $currentTab = $(this)
+    $currentTab.addClass('active')
+    panelId = $currentTab.attr('for')
+    $('.tab-panel').addClass('is-invisible')
+    $('#' + panelId).removeClass('is-invisible')
+
+    # is-invisible class is not working for the spatial box
```
charleshuang80: I think we don't need this comment. It is true that `is-invisible` is not working, but that is because of the way the class interacts with the element. More importantly, I also tried using `show()` and `hide()`, which worked, but this way ends up being a cleaner way to do it because of clicking in between tabs.

app/assets/stylesheets/components/_metadata-preview.scss
```diff
@@ -202,71 +202,79 @@ ul.sub-field {
...
+      background: linear-gradient(to bottom,#fdfdfe 0%,#eff1f4 100%);
+      border: 1px solid #c9dce7;
+      color: #7f7f7f;
+      cursor: pointer;
+      font-size: 16px;
+      font-weight: bold;
+      margin-right: 5px;
+      padding: 12px 15px;
+      transition: background 0.1s, color 0.1s;
+      // width: 100%;
```
charleshuang80: I think we should clean up the commented css we don't need anymore

app/assets/javascripts/preview.coffee
htranho: removed

app/assets/stylesheets/components/_metadata-preview.scss
htranho: done

PR #365, opened 2019-07-03
feature/MMT-1844 now uses legacy-services url instead of echo one.

cgokey:


config/services.yml
```diff
@@ -18,7 +18,7 @@ earthdata:
     # uvg_root: "https://uvg.uat.earthdata.nasa.gov"
   sit:
     urs_root: "https://sit.urs.earthdata.nasa.gov"
-    echo_root: "https://testbed.echo.nasa.gov"
+    echo_root: "https://cmr.sit.earthdata.nasa.gov/legacy-services"
```
cgokey: change uat and prod api endpoints.

PR #367, opened 2019-07-08
Feature/MMT-1817

ryanmiller-1:
I tried to merge master into 1817 and something weird happened.  These changes are the same as those changes, but without replicating all of the preview gem commits/changes.

### DIFF
--- a/app/controllers/base_drafts_controller.rb
+++ b/app/controllers/base_drafts_controller.rb
@@ -19,6 +19,11 @@ class BaseDraftsController < DraftsController
     end
 
     add_breadcrumb breadcrumb_name(get_resource.draft, resource_name), send("#{resource_name}_path", get_resource)
+
+    respond_to do |format|
+      format.html { return }
+      format.json { render json: JSON.pretty_generate(get_resource.draft) }
+    end
   end
 
   def new
--- a/spec/features/collection_drafts/collection_draft_creation_spec.rb
+++ b/spec/features/collection_drafts/collection_draft_creation_spec.rb
@@ -54,6 +54,16 @@ describe 'Collection Draft creation', js: true do
             expect(page).to have_content('123')
           end
         end
+
+        context 'when accessing a collection draft\'s json' do
+          before do
+            visit collection_draft_path(CollectionDraft.first, format: 'json')
+          end
+
+          it 'returns the json' do
+            expect(page).to have_content('123')
+          end
+        end
       end
     end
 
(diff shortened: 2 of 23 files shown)


ryanmiller-1:
Had a conflict in one of the test files that I modified in a branch already merged in.

app/controllers/base_drafts_controller.rb
```diff
@@ -19,6 +19,11 @@ def show
     end
 
     add_breadcrumb breadcrumb_name(get_resource.draft, resource_name), send("#{resource_name}_path", get_resource)
+
+    respond_to do |format|
+      format.html { return }
```
cgokey: Is this line necessary?   if format matches html, won't it just return anyhow?

spec/features/collection_drafts/collection_draft_creation_spec.rb
```diff
@@ -54,6 +54,16 @@
...
           end
         end
+
+        context 'when accessing a collection draft\'s json' do
+          before do
+            visit collection_draft_path(CollectionDraft.first, format: 'json')
+          end
+
+          it 'returns the json' do
+            expect(page).to have_content('123')
```
cgokey: wonder if you should check for a json tag here rather than just content?

app/controllers/base_drafts_controller.rb
ryanmiller-1: The format.html { } is necessary or rails generates an unknown format error.  The return can be omitted.  I'd argue that { return } is more transparent than { }.

cgokey: Agree, thanks.

spec/features/collection_drafts/collection_draft_creation_spec.rb
ryanmiller-1: I certainly accede to the argument that I should check for a json tag.  My question is if there is value in checking that the content is as expected or if we trust the content is verified elsewhere.

cgokey: I'm inclined to think a quick check that a json tag exists, as the purpose of the ticket is produce some kind of json output, seems a check with that regard would be useful in case something changes later and breaks it.

ryanmiller-1: Selenium doesn't support checking the header tags.  I have updated the tests so that they are actually differentiating between what they expect from an html response versus a json response.  As Thomas described it, it's duck type testing.  I do not think this satisfies the heart of your request, but I think this is the best available compromise.

PR #372, opened 2019-07-25
MMT-1654 set up CollectionDraftProposal model

charleshuang80:
* set checks for draft mmt configuration
* write and run migration
* write new model, set up inheritance chain but use new table
* write tests for new model


app/models/collection_draft.rb
```diff
@@ -30,24 +30,25 @@ def get_next_form(name, direction)
...
-        draft = CollectionDraft.create
+        draft = self.create
         draft.entry_title = "#{new_entry_title} - Cloned"
         collection['EntryTitle'] = "#{new_entry_title} - Cloned"
         draft.short_name = nil
         collection.delete('ShortName')
         collection.delete('MetadataDates')
       end
-      draft.user = user
-      draft.provider_id = user.provider_id # TODO is this problematic for collections editing permissions?
```
charleshuang80: relook at this.

app/models/collection_draft.rb
```diff
@@ -30,24 +30,25 @@ def get_next_form(name, direction)
     end
 
     def create_from_collection(collection, user, native_id)
+      # TODO: try to refactor this method for all drafts
       new_entry_title = (collection['EntryTitle'].blank?) ? nil : collection['EntryTitle']
 
       if native_id
         # Edited record
-        draft = CollectionDraft.find_or_create_by(native_id: native_id)
+        draft = self.find_or_create_by(native_id: native_id)
```
cgokey: so just using "self" worked well, no need to subclass or anything these, like the solution.

PR #373, opened 2019-07-25
Feature/mmt 36

ryanmiller-1:


app/views/collection_templates/index.html.erb
```diff
@@ -0,0 +1,55 @@
+<div class="row row-content">
+  <section>
+    <h2><%= current_user.provider_id %> <%= plural_resource_name.titleize %></h2>
+
+    <%= link_to "Create a #{resource_name.titleize}", send("new_#{resource_name}_path"), class: 'eui-btn--blue space-bot' %>
+
```
ryanmiller-1: Everything beyond line 5 in this file is likely to change, in whole or in part, in subsequent tickets.  

Minimally, the plan is to make action column have two sub columns that contain "Edit" and "Delete" with appropriate links.  Additionally, the titles will link to the show pages, not to making a draft.

app/helpers/collections_cmr_helper.rb
```diff
@@ -0,0 +1,23 @@
+module CollectionsCmrHelper
```
ryanmiller-1: Per Charles comment about helpers largely being for views, we should discuss if/where we want to make this function available without having to manage two copies of it.

app/controllers/collection_templates_controller.rb
```diff
@@ -0,0 +1,119 @@
...
+      collection
+    else
+      nil
+    end
+  end
+
+  # Provide a list of names currently in the database for validation.
+  def names_list(id = nil)
+    policy_scope(resource_class).map { |template| template['template_name'] unless template.id == id }
+  end
```
ryanmiller-1: This provides the correct behavior for this ticket, but will be changed in the edit ticket.  A reduce function should provide correct behavior for both.  I can update this ticket with that change, if desired.

ryanmiller-1: I updated the function when I made a bug fix today.

app/controllers/collection_templates_controller.rb
```diff
@@ -0,0 +1,121 @@
+# frozen_string_literal: true
+
+class CollectionTemplatesController < CollectionDraftsController
+  include CollectionsCmrHelper
+  before_action :templates_enabled?
+  before_action :set_resource, only: %i[create_draft destroy edit update show]
```
charleshuang80: if these `before_action`s are already in the super class, you probably don't need to declare them again

app/controllers/collection_templates_controller.rb
```diff
@@ -0,0 +1,121 @@
...
+    set_resource(resource_class.new(user: current_user, provider_id: current_user.provider_id, draft: new_draft.to_camel_keys))
+
+    super
+  end
+
+  def new_from_existing
+    set_resource(resource_class.new(user: current_user, provider_id: current_user.provider_id, draft: fetch_source_data))
+    authorize get_resource
+
+    @draft_forms = CollectionDraft.forms
```
charleshuang80: We are making changes like this for MMT-1667. I think that `@draft_forms` (plural and singular) should be changed to the corresponding `@forms`, and `CollectionDraft` should be changed to the generic method to call the class, as the forms should be inherited.

app/helpers/collections_cmr_helper.rb
charleshuang80: this is very subjective/semantic, but I think `CmrCollectionsHelper` might be a little better (no need to change if you disagree)

charleshuang80: I think this can be moved into the `concerns` folder as it's not being used for views

spec/factories/collection_template.rb
```diff
@@ -0,0 +1,831 @@
...
+          'Type' => 'CREATE',
+          'Date' => '2010-12-25T00:00:00Z'
+        }, {
+          'Type' => 'UPDATE',
+          'Date' => '2010-12-30T00:00:00Z'
+        }]
+      }}
+    end
+
+    draft {{
```
charleshuang80: I wonder if there is a way to save this draft metadata and have it be used by multiple factory methods, since it is fairly long and we might also want to use it for draft proposals.

app/controllers/collection_drafts_controller.rb
```diff
@@ -8,14 +8,14 @@ class CollectionDraftsController < BaseDraftsController
   layout 'collection_preview', only: [:show]
 
   def new
-    set_resource(CollectionDraft.new(user: current_user, provider_id: current_user.provider_id, draft: {}))
+    set_resource(resource_class.new(user: current_user, provider_id: current_user.provider_id, draft: {}))
```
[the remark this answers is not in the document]
ryanmiller-1: All of these sorts of changes are made to generalize the collection drafts.  They are using things that are already being done (e.g. resource_class isn't something I made).  Unless you are specifically avoiding resource_class and its ilk, I believe this should work for proposals as well.

app/controllers/collection_templates_controller.rb
```diff
@@ -0,0 +1,121 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
ryanmiller-1: At some point during the process, I told rubocop to fix the file automagically and it added this (among other things).  At least in Atom, it complains about this particular magic comment in many of our files.  The code works without it, but the internet suggests that this magic comment should help performance.

app/controllers/collection_templates_controller.rb
```diff
@@ -0,0 +1,121 @@
...
+
+class CollectionTemplatesController < CollectionDraftsController
+  include CollectionsCmrHelper
+  before_action :templates_enabled?
+  before_action :set_resource, only: %i[create_draft destroy edit update show]
+  before_action :load_umm_schema, only: %i[new edit show new_from_existing]
```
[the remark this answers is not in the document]
ryanmiller-1: This is something that rubocop advocates [https://github.com/rubocop-hq/ruby-style-guide#i](here).

app/assets/stylesheets/components/_callout-box.scss
```diff
@@ -99,7 +99,9 @@
 
   /* Form Elements
   ------------------------------------------*/
-  form {
+/*  This does not appear to be used anywhere.  Should we retain this?
```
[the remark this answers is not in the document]
ryanmiller-1: @jceaser @cgokey @hvtranho

app/controllers/collection_templates_controller.rb
```diff
@@ -0,0 +1,119 @@
...
+
+  def create_draft
+    authorize get_resource
+    draft = CollectionDraft.create_from_template(get_resource, current_user)
+    Rails.logger.info("Audit Log: Collection Draft #{draft.entry_title} was created from a template by #{current_user.urs_uid} in provider: #{current_user.provider_id}")
```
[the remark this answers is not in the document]
ryanmiller-1: I suspect it was chosen because it mirrors the logging in collection drafts, but I will try to craft a more useful message to log when I update that feature in MMT-38.

PR #375, opened 2019-07-31
feature/MMT-37 Updating index page gui to mirror existing index pages

ryanmiller-1:


spec/features/collection_templates/template_index_spec.rb
```diff
@@ -0,0 +1,34 @@
+# The tests in this file verify that the index page contains the expected elements
+
+describe 'When visiting the template index page', js:true do
+  before do
```
charleshuang80: I think you should wrap all these contents in another context block: `context 'when there is one record' do...`

spec/features/collection_templates/template_index_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  it 'has expected links for one record' do
+    # Create button is actually a link
+    expect(page).to have_link('Create a Collection Template', href: '/collection_templates/new')
+    expect(page).to have_link('An Example Template', href: '/collection_templates/1')
+    expect(page).to have_link('Edit', href: '/collection_templates/1/edit')
+    expect(page).to have_link('Delete', href: '#delete-template-modal-0')
+  end
+
+  context 'and has multiple records' do
```
charleshuang80: then the text here should change to 'when there are multiple records'. If you look at the usage of context blocks in the feature tests, they pretty much all start with 'when'

PR #377, opened 2019-08-01
MMT-1667 Create a collection draft in draft-only-MMT

htranho:


app/controllers/collection_drafts_controller.rb
```diff
@@ -8,14 +8,14 @@ class CollectionDraftsController < BaseDraftsController
   layout 'collection_preview', only: [:show]
 
   def new
-    set_resource(CollectionDraft.new(user: current_user, provider_id: current_user.provider_id, draft: {}))
+    set_new_resource
```
jceaser: what resource is new? not descriptive enough

htranho: well, that was a new name just for refactoring, we need to come up with a descriptive name.

app/controllers/collection_drafts_controller.rb
```diff
@@ -170,6 +169,10 @@ def collection_draft_params
 
   private
 
+  def set_resource_by_model
+    set_resource(CollectionDraft.new(user: current_user, provider_id: current_user.provider_id, draft: {}))
```
charleshuang80: what happens when this is a proposal or a template? shouldn't you one of the more generic methods for the model name?

app/controllers/proposal/collection_draft_proposals_controller.rb
```diff
@@ -0,0 +1,18 @@
+module Proposal
+  class CollectionDraftProposalsController < CollectionDraftsController
+
+    def publish
```
charleshuang80: did you see my comment in Ryan's other PR where he overwrote this method?

app/policies/collection_draft_proposal_policy.rb
```diff
@@ -0,0 +1,29 @@
+class CollectionDraftProposalPolicy < DraftPolicy
```
charleshuang80: This works for when we were working on the first ticket, but I think we need to come up with a programmatic way of allowing or blocking access. I am thinking along the lines of using the environmental configuration and possibly checking that they are logged in through EDL, as those are the requirements to access proposals. `publish?` should probably just stay as `false` though.

app/helpers/manage_metadata_helper.rb
```diff
@@ -33,6 +33,8 @@ def current_manage_title
       'manage_variables'
     elsif is_services_controller?
       'manage_services'
+    elsif controller.lookup_context.prefixes.include?('proposal/manage_collection_proposals')
```
charleshuang80: I feel like this can be even more generic. For the time being anything namespaced under `proposal` should give this title. Also, does this miss everything in CollectionDraftProposalsController? seems to me like it would

app/controllers/collection_drafts_controller.rb
ryanmiller-1: I believe this is trying to solve the same problem as set_user_and_provider.  The stated reasoning during the review was that proposals do not have providers.  Reading over the feature branch, you solved that elsewhere.

app/helpers/manage_metadata_helper.rb
htranho: Now it looks for an entry with prefix 'proposal'

app/controllers/proposal/collection_draft_proposals_controller.rb
htranho: Check publish() in collection_templates_cotroller.rb

app/controllers/collection_drafts_controller.rb
htranho: Added comment in method: We don't want to sent provider id

app/controllers/proposal/collection_draft_proposals_controller.rb
htranho: Raised exception and redirect to normal mmt path

app/policies/collection_draft_proposal_policy.rb
htranho: Now check for proposal mode and if user logged in

app/controllers/application_controller.rb
```diff
@@ -10,6 +10,7 @@ class ApplicationController < ActionController::Base
   before_action :refresh_urs_if_needed, except: [:login, :logout, :refresh_token] # URS login
   before_action :refresh_launchpad_if_needed, except: [:login, :logout] # Launchpad login
   before_action :provider_set?
```
htranho: should taken out in proposal controller

app/controllers/proposal/collection_draft_proposals_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      flash[:error] = 'Collection Draft Proposal cannot be published.'
```
htranho: check pundit policy

app/controllers/proposal/collection_draft_proposals_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      redirect_to manage_collections_path
```
htranho: to the proposal mode!

app/helpers/manage_metadata_helper.rb
```diff
@@ -24,6 +24,15 @@ def is_services_controller?
     (SERVICES_CONTROLLERS & controller.lookup_context.prefixes).any? || (controller.lookup_context.prefixes.include?('collection_associations') && params[:service_id])
   end
 
+  def is_collection_draft_proposal_controller?
+    controller.lookup_context.prefixes.each do |item|
+      if item.start_with?('proposal')
+        return true
+      end
+      return false
```
htranho: fix this

app/controllers/application_controller.rb
htranho: Done

app/controllers/proposal/collection_draft_proposals_controller.rb
htranho: done

app/helpers/manage_metadata_helper.rb
htranho: Fixed

PR #379, opened 2019-08-05
Mmt 1888

ryanmiller-1:
Deleting the last data contact was not properly renumbering visible elements and was failing to properly renumber the select2 element.


jceaser:
bamboo is green

PR #382, opened 2019-08-07
feature/MMT-40 Updating branch to not have preview gem changes

ryanmiller-1:


app/controllers/collection_templates_controller.rb
```diff
@@ -46,6 +52,21 @@ def create
     end
   end
 
+  def update
+    if names_list(params[:id]).include?(params[:draft][:template_name])
+      flash[:error] = 'A template with that name already exists.'
+      redirect_to manage_collections_path
+      return
```
charleshuang80: if you need to add a return after a redirect, I really like `redirect_to <some path> and return`

app/controllers/collection_templates_controller.rb
```diff
@@ -46,6 +52,21 @@ def create
...
+    end
+
+    begin
```
charleshuang80: How many checks for duplicate template names are there? can they be consolidated?

Also, shouldn't our front end validation (if working correctly) also prevent a user from saving/updating if a template name already exists? If that is the case, we probably should still have a just in case check in the controller, but we should definitely consolidate.

spec/factories/collection_template.rb
```diff
@@ -24,808 +24,14 @@
...
-          'DetailedClassification' => 'test 2 Detailed Classification',
-          'Period'                 => 'test 2 Period'
-        }, {
-          'Eon' => 'test 3 Eon text 1'
-        }]
-      }],
-      'ISOTopicCategories' => %w[farming climatologyMeteorologyAtmosphere health]
-    }}
+    draft {
+      collection_one.merge(
```
charleshuang80: I am debating if we should try and make the corresponding change in collection draft factories, or write a tech debt ticket to consolidate later, as the proposals probably could do this also.

spec/features/collection_templates/template_edit_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  context 'when editing a draft' do
+    before do
+      draft = create(:full_collection_template, template_name: 'An Example Template2')
+      visit edit_collection_template_path(draft)
+    end
+
+    it 'provides error validation for template names' do
+      expect(page).to have_content('An Example Template2')
+      fill_in 'Template Name', with: 'An Example Template'
+      # Need to do something else to trigger the validation because the fill_in
```
charleshuang80: You can always just find another element and click on it, I have had to do that for other tests

spec/features/collection_templates/template_index_spec.rb
```diff
@@ -1,36 +1,50 @@
...
+      expect(page).to have_content('Metadata Preview')
+    end
+    
+    context 'when there are multiple records' do
+      before do
+        create(:full_collection_template, template_name: 'An Example Template2')
+        visit collection_templates_path
+      end
+
+      it 'has expected links for both records' do
```
charleshuang80: The way this is done is fine. Another way (and possibly more brittle) might be https://github.com/nasa/mmt/blob/90df72221f0948fead9df71457abee7deb9c81af/spec/features/groups/filtering_groups_spec.rb#L38.

app/controllers/collection_templates_controller.rb
ryanmiller-1: Currently, the UI, the controller, and DB all check.  The front end validation cannot be trusted because the list of names is created when the page is loaded.  For instance, submitting, going back (browser) and then submitting again circumvents this validation.  The controller one can probably be removed, but I consider the DB one mandatory.

app/controllers/collection_templates_controller.rb
ryanmiller-1: I don't have a strong opinion on it, but I believe that rubocop style guide bans the use of 'and' and 'or'.

spec/features/collection_templates/template_index_spec.rb
ryanmiller-1: I would definitely be more inclined to do that as the number of things I was testing increased.  Thank you for sharing.

app/controllers/collection_templates_controller.rb
charleshuang80: yeah, but this is a specific pattern for rails, which is why I think it's ok and we have made an exception for it

app/controllers/collection_templates_controller.rb
charleshuang80: Ok, so if the DB check covers the other one in the controller, let's just keep the one. However, the way this is written seems to break rails convention in the interest of keeping the method of the superclass intact. I think we should try to use a model validation, which should catch this error. If that means we need to tweak or rewrite some of the code in the superclass for the method to follow the rails convention then I think that is acceptable.

charleshuang80: https://guides.rubyonrails.org/v4.2/active_record_validations.html

app/views/collection_drafts/edit.html.erb
```diff
@@ -50,6 +50,8 @@
         var globalJsonSchema = <%= raw @json_schema.to_json %>;
 
         var urlContentTypeMap = <%= raw DraftsHelper::URLContentTypeMap.to_json %>;
+
+        var globalTemplateNames = <%= raw @template_names.to_json %>;
```
jceaser: wait, why raw? Show me that this is safe?

app/controllers/collection_drafts_controller.rb
```diff
@@ -89,7 +59,11 @@ def create
       end
     else # record update failed
       flash[:error] = I18n.t("controllers.draft.#{plural_resource_name}.create.flash.error")
+      load_umm_schema
```
ryanmiller-1: render loads the view without calling the function.  This means that many of the variables that are set in the new function are set, so they need to be set now.  render is used so that the data the user did change can be retained.

app/controllers/collection_drafts_controller.rb
```diff
@@ -403,4 +380,42 @@ def validate_picklists(errors, metadata)
   def set_language_codes
     @language_codes = cmr_client.get_language_codes.to_a
   end
+
+  def new_view_setup
```
ryanmiller-1: As this and the edit_view_setup are used twice, they have been extracted here.

app/models/collection_draft.rb
```diff
@@ -97,11 +97,11 @@ def update_draft(params, editing_user_id)
 
       if new_draft
         self.draft = new_draft
-        save
+        return save
       elsif self.draft != {}
         # draft had content, but now new_draft is nil/empty so any data has been deleted and saved in the forms
         self.draft = {}
-        save
+        return save
```
ryanmiller-1: This function would save, then regardless of the outcome return true (below).  These mean that the response from save is provided back to the calling function for evaluation.

app/views/collection_drafts/new.html.erb
```diff
@@ -5,14 +5,20 @@
...
 <% if get_resource.errors.any? %>
-  <div id="error_explanation">
-    <h2><%= pluralize(get_resource.errors.count, 'error') %>) prohibit this draft from being saved:</h2>
-    <ul>
-      <% get_resource.errors.full_messages.each do |msg| %>
-        <li><%= msg %></li>
-      <% end %>
-    </ul>
-  </div>
+  <section class="errors">
```
ryanmiller-1: When considering what we want to try to reuse in partials, I think this is a good candidate.

app/controllers/collection_drafts_controller.rb
charleshuang80: nice refactoring

app/controllers/collection_drafts_controller.rb
```diff
@@ -116,8 +89,10 @@ def update
       end
     else # record update failed
       # render 'edit' # this should get get_resource_form
-      # Remove
       flash[:error] = I18n.t("controllers.draft.#{plural_resource_name}.update.flash.error")
```
charleshuang80: We've discussed this, but now it seems that the template specific error will not go in the flash, the flash message will be a generic message. But because there is a save error for the model object, we that error will still be presented to the user. That seems like a good solution.

app/views/collection_drafts/new.html.erb
charleshuang80: sounds good. feel free to do that either now or in the next ticket.

app/controllers/collection_drafts_controller.rb
```diff
@@ -89,6 +59,9 @@ def create
       end
     else # record update failed
       flash[:error] = I18n.t("controllers.draft.#{plural_resource_name}.create.flash.error")
+      load_umm_schema
+      new_view_setup
+      @template_names = names_list if resource_name == 'collection_template'
```
charleshuang80: could this be moved into the subclass method?

app/controllers/collection_drafts_controller.rb
```diff
@@ -116,8 +89,10 @@ def update
       end
     else # record update failed
       # render 'edit' # this should get get_resource_form
-      # Remove
       flash[:error] = I18n.t("controllers.draft.#{plural_resource_name}.update.flash.error")
+      load_umm_schema
+      edit_view_setup
+      @template_names = names_list(params[:id]) if resource_name == 'collection_template'
```
charleshuang80: could this be moved into the subclass method?

app/controllers/collection_drafts_controller.rb
charleshuang80: It just feels a little bit weird to be adding a call in the superclass method that is required and will only be run by the subclass.

Do the names need to be set at this particular point? Could they be set before super is called in the subclass? Or potentially with a before_action in the subclass?

ryamil1: They could be called in the template's create before the super call.  I'll make the change in the morning.

ryanmiller-1: I've moved it before the super call in the latest commit.

PR #393, opened 2019-08-16
Feature/mmt 1891

ryanmiller-1:
These codes changes include: 
- Updating how model errors are shown per Alicia's feedback
- Updating draft/draft_forms => unpublished resource
- Partialing out model errors
- Partialing out invalid draft modal
- Partialing out invalid template modal
- Updating validation error display to show template instead of draft when it is a template
- Adding back in the template reminder
- Updating to include set_resource_by_model changes
  - Refactoring the collection template controller
- Adding a pundit policy for publish that is always false
- Fixing another bar problem on the collections page


app/assets/javascripts/validation.coffee
```diff
@@ -248,10 +248,17 @@ $(document).ready ->
...
+    # This modal is loaded on every new/edit page for templates, but not drafts
+    if $('#invalid-template-modal').length > 0
+      summary = $('<div/>',
+        class: 'eui-banner--danger summary-errors'
+        html: '<h4><i class="fa fa-exclamation-triangle"></i> This template has the following errors:</h4>'
+      )
+    else
+      summary = $('<div/>',
+        class: 'eui-banner--danger summary-errors'
+        html: '<h4><i class="fa fa-exclamation-triangle"></i> This draft has the following errors:</h4>'
```
charleshuang80: I'm wondering if we can make this a little more succinct by setting a variable (just an example but maybe `resource_type`) to 'template' or 'draft', then interpolating it into the html section (i.e. `"... This #{resource_type} ..."`)

app/assets/javascripts/validation.coffee
```diff
@@ -605,18 +612,20 @@ $(document).ready ->
...
-      return switch this.type
-        when 'radio'
-          # don't want to save fields that aren't translated into metadata
-          this.name? and this.checked
-        else
-          this.value
-    .each (index, element) ->
-      visitField($(element).attr('id'))
+    # Do not display validation errors on page load if model errors are showing
+    if $('.errors').length == 0
```
charleshuang80: I would prefer to add and use a more specific class for this check, maybe `.model-errors`

app/views/collection_drafts/forms/_model_errors.html.erb
```diff
@@ -0,0 +1,5 @@
+<% if get_resource.errors.any? %>
+  <!-- Model Errors Flag -->
+  <section class="errors">
```
charleshuang80: is this class used for anything besides the check commented on above? If so, can we add a more specific class to use for that check, i.e. `model-errors`?

app/controllers/collection_drafts_controller.rb
```diff
@@ -414,4 +414,18 @@ def edit_view_setup
...
+
+  def generate_model_error
+    return unless get_resource.errors.any?
+    model_error_message = ''
+    get_resource.errors.full_messages.each do |msg|
+      model_error_message += ', ' unless model_error_message == ''
+      model_error_message += msg.downcase
+    end
+    model_error_message
+  end
```
charleshuang80: This is going to get really ruby condensed (which some people don't like it as much), but can we try replacing lines 424 - 429 with:
`get_resource.errors.full_messages.reject{ |msg| msg.blank? }.map{ |msg| msg.downcase }.join(', ')`

config/locales/en.yml
```diff
@@ -20,6 +20,10 @@
 # available at http://guides.rubyonrails.org/i18n.html.
 
 en:
+  activerecord:
+    errors:
+      messages:
+        taken: 'must be unique within a provider context'
```
charleshuang80: is this default message tested anywhere?

ryanmiller-1: spec/features/collection_templates/template_creation_spec.rb 189

app/assets/javascripts/validation.coffee
ryanmiller-1: Done

app/views/collection_drafts/forms/_model_errors.html.erb
ryanmiller-1: Done

app/assets/javascripts/validation.coffee
ryanmiller-1: Done.  Didn't know coffeescript had included an alternate way of doing interpolation.  JS's backticks were not cooperating when I tried it originally.

app/controllers/collection_drafts_controller.rb
ryanmiller-1: I would be more wary of it with larger datasets because you're walking it twice, but it is expected to be small here.

ryanmiller-1: Also rubocop suggests     get_resource.errors.full_messages.reject(&:blank).map(&:downcase).join(', ') instead

app/assets/javascripts/validation.coffee
```diff
@@ -605,18 +607,20 @@ $(document).ready ->
...
-    .each (index, element) ->
-      visitField($(element).attr('id'))
+    # Do not display validation errors on page load if model errors are showing
+    if $('.model-errors').length == 0
+      # "visit" each field with a value on page load
+      $('.validate').not(':disabled').filter ->
+        return switch this.type
+          when 'radio'
+            # don't want to save fields that aren't translated into metadata
+            this.name? and this.checked
```
cgokey: why use a switch when if/else will do?

PR #403, opened 2019-09-04
MMT-1898 Data Center Contact without Data Center Short Name

htranho:
Fixed issue: Data Contact without Data Center Short Name was removed when updating Data Center.


htranho:
ticket is deferred

PR #407, opened 2019-09-09
MMT-1435 Updating System and Provider permissions to block if another…

ryanmiller-1:
… user has edited the permissions rather than overwrite.
In the situation:
User A opens the page, User B opens the page, makes changes, and then submits the form, then User A submits the form.
the result should be that User A's overlapping changes are rejected and User A is informed of the changes that succeed/fail.
This required passing information about current revision IDs, and verifying that information while preparing the list of ACLs to change.


app/controllers/provider_identity_permissions_controller.rb
```diff
@@ -57,10 +59,10 @@ def edit
 
   def update
     @group_id = params[:id]
-    permissions_params = params[:provider_permissions]
+    permissions_params = params[:provider_permissions] || {}
     redirect_to provider_identity_permissions_path and return if permissions_params.nil?
```
ryanmiller-1: remove 63

app/concerns/permission_management.rb
```diff
@@ -74,52 +97,96 @@ def assemble_permissions_for_updating(full_permissions:, type:, group_id:)
...
+            # removing permissions to the target, need to remove the group from the acl
+            targets_to_remove_group << target
+            target_revision_ids[target] = params["#{target}_revision_id"]
+          else
+            targets_to_fail << target
+          end
         elsif permissions_params[target] == [] && acl_info['num_groups'] == 1
-          # removing permissions, but group is only group in the acl, need to delete the acl
-          targets_to_delete << target
+          # *** CONCERN ***
```
ryanmiller-1: Is this really true?

ryanmiller-1: Adding the revision ID to the header as we do in update does not change the CMR's willingness to delete ACLs.  Given that, and that the history is that we had to specifically ask for revision ID on updates, I believe that they have not made a similar change to deletes.

app/concerns/permission_management.rb
```diff
@@ -74,52 +97,96 @@ def assemble_permissions_for_updating(full_permissions:, type:, group_id:)
...
+          # There is no way to guarantee delete does not overwrite changes from other users
+          # because the CMR does not accept a revision_id on delete calls. Comparing two
+          # revision ids before we send the request is the best I can see to do in the current
+          # state.
+          if params["#{target}_revision_id"].present? && params["#{target}_revision_id"] == (current_revisions[target]).to_s
+            # removing permissions, but group is only group in the acl, need to delete the acl
+            targets_to_delete << target
+            target_revision_ids[target] = params["#{target}_revision_id"]
+          else
+            # If there is no revision_id but the acl_info says we should be deleting here,
```
charleshuang80: update comment

app/concerns/permission_management.rb
```diff
@@ -28,6 +28,29 @@ def get_permissions_for_identity_type(type:, group_id: nil)
...
+        revision_ids[permission['acl']["#{type}_identity"]['target']] = permission['revision_id']
+      end
+    else
+      Rails.logger.error("Get #{type.titleize} Revisions Fetch Error: #{permissions_response.clean_inspect}")
+      flash[:error] = revisions_response.error_message
+    end
+    revision_ids
+  end
+
   def assemble_permissions_for_table(permissions:, type:, group_id:)
```
charleshuang80: check if revision ids can be populated within this function

app/concerns/permission_management.rb
```diff
@@ -74,52 +97,96 @@ def assemble_permissions_for_updating(full_permissions:, type:, group_id:)
...
       if acl_info['matched_to_group'] # group is currently in the target acl
         if acl_info['granted_permissions'] == permissions_params[target]
           # permissions are the same, do nothing
           targets_to_skip << target
         elsif permissions_params[target] == [] && acl_info['num_groups'] > 1
-          # removing permissions to the target, need to remove the group from the acl
-          targets_to_remove_group << target
+          # The acl_info is information obtained AFTER submission of the form
+          # The revision_id from the parameters is obtained BEFORE submission of the form
+          if params["#{target}_revision_id"].present?
```
charleshuang80: can this be a part of permissions_params?

app/concerns/permission_management.rb
```diff
@@ -220,7 +294,7 @@ def create_target_permissions(targets_to_create:, permissions_params:, identity_
     end
   end
 
-  def construct_permission_object(group_id:, permissions:, target:, identity_type:, target_id: nil)
+  def construct_permission_object(group_id:, permissions:, target:, identity_type:, target_id: nil) #target_id: is unused here?
```
charleshuang80: address comment

app/controllers/provider_identity_permissions_controller.rb
```diff
@@ -69,9 +70,13 @@ def update
...
 
-    targets_to_add_group, targets_to_update_perms, targets_to_remove_group, targets_to_create, targets_to_delete = sort_permissions_to_update(assembled_all_permissions: selective_provider_permission_info, permissions_params: permissions_params)
-
+    targets_to_add_group, targets_to_update_perms, targets_to_remove_group, targets_to_create, targets_to_delete, targets_to_fail, target_revision_ids = sort_permissions_to_update(assembled_all_permissions: selective_provider_permission_info, permissions_params: permissions_params, type: 'provider')
+    next_revision_ids = {}
+    target_revision_ids.each do |key, value|
+      next_revision_ids[key] = (Integer(value) + 1).to_s
+    end
     successes = []
+    overwrite_fails = targets_to_fail
```
charleshuang80: maybe just change it above on line 73

app/concerns/permission_management.rb
ryanmiller-1: With modifications to get_permissions and both assemble functions, I was able to remove the function I added.

app/concerns/permission_management.rb
ryanmiller-1: With the revision ids provided by the assemble function, the overwriting checks can and are being done in the controller.  As a consequence, there is no longer a need to pass revision ids in or use them from the params.

app/controllers/system_identity_permissions_controller.rb
```diff
@@ -97,11 +112,23 @@ def update
...
+      overwrite_fails: overwrite_fails,
+      revision_ids: next_revision_ids
     )
 
-    flash[:success] = 'System Object Permissions were saved.' unless successes.blank?
-    flash[:error] = "#{fails.join(', ')} permissions were unable to be saved." unless fails.blank?
+    unless successes.blank?
+      flash[:success] = successes.join(', ').titleize + ' permissions were saved.'
+    end
+    unless fails.blank? && overwrite_fails.blank?
```
charleshuang80: I think this `unless` block is a little more convoluted than necessary as is.
Line 123 can be removed as 124 and 125 are already doing the `unless` check.
For the rest, this is just a suggestion, but to make it more readable, I might try:
```
if error_message.present? && overwrite_error_message.present?
  ...
elsif error_message.present?
  ...
elsif overwrite_error_message.present?
  ...
end
```

app/controllers/provider_identity_permissions_controller.rb
```diff
@@ -98,11 +113,23 @@ def update
...
+      overwrite_fails: overwrite_fails,
+      revision_ids: next_revision_ids
     )
 
-    flash[:success] = 'Provider Object Permissions were saved.' unless successes.blank?
-    flash[:error] = "#{fails.join(', ')} permissions were unable to be saved." unless fails.blank?
+    unless successes.blank?
+      flash[:success] = successes.join(', ').titleize + ' permissions were saved.'
+    end
+    unless fails.blank? && overwrite_fails.blank?
```
charleshuang80: see my comment for system permissions

PR #408, opened 2019-09-10
MMT-1917 replacing entry_id with short name in Draft MMT

ryanmiller-1:


spec/features/proposal/collection_draft_proposals/collection_draft_proposal_list_spec.rb
```diff
@@ -16,13 +16,13 @@
...
-    it 'displays versions correctly' do
-      expect(page).to have_content('An Example Proposal_5')
+    it 'displays short names correctly' do
+      expect(page).to have_content('An Example Proposal')
       expect(page).to have_content('An Example Proposal2')
     end
 
     it 'displays entry titles' do
-      expect(page).to have_content('An Example Proposal Title')
+      expect(page).to have_content('An Example proposal Title')
```
charleshuang80: let's have more words to increase uniqueness, rather than expecting folks to know that the difference in case is for uniqueness

PR #412, opened 2019-09-16
MMT-1641 Removed logging for status response when successful and upda…

ryanmiller-1:
…ted headers to include app name and version.

### DIFF
--- a/app/middleware/middleware_healthcheck.rb
+++ b/app/middleware/middleware_healthcheck.rb
@@ -6,7 +6,7 @@ class MiddlewareHealthcheck
   def call(env)
     if env['PATH_INFO'.freeze] == '/status'.freeze
       Rails.logger.tagged "middleware health check" do
-        response = [503, {'Content-Type' => 'application/json'}]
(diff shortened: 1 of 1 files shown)


app/middleware/middleware_healthcheck.rb
```diff
@@ -37,7 +37,6 @@ def call(env)
         if db_healthy && (ENV['launchpad_login_required'] != 'true' || launchpad_healthy)
           response[0] = 200
         end
-        Rails.logger.info "The Status page returned a #{response[0]} Response:#{response.inspect}"
```
charleshuang80: this looks like it is removing the logging for _both_ a 200 status response as well as a 503 status response.

jceaser:
I see the content when I tried this myself.

app/middleware/middleware_healthcheck.rb
ryanmiller-1: The final log, yes, but there are logs associated with each of the other failure conditions.  If we want to log the failure response here, in addition to those, I could add an else to the preceding if for this logging.

PR #413, opened 2019-09-17
MMT-1909: require Non-NASA Draft User ACL to access Draft only MMT and CRUD proposals

charleshuang80:
* add Pundit policy
* refactor policies
* refactor ACL checking methods
* add tests for requiring Non-NASA Draft User ACL


app/policies/provider_order_policy.rb
```diff
@@ -1,6 +1,7 @@
 # :nodoc:
-class ProviderOrderPolicy < ProviderPolicy
+class ProviderOrderPolicy < ApplicationPolicy
   def resubmit?
-    user_has_permission_to('create', 'PROVIDER_ORDER_RESUBMISSION')
+    # user_has_permission_to('create', 'PROVIDER_ORDER_RESUBMISSION')
```
charleshuang80: remove comment

app/controllers/manage_metadata_controller.rb
```diff
@@ -311,4 +311,12 @@ def get_page(fields)
     TemporalKeywords
     PaleoTemporalCoverages
   )
+
+  def ensure_non_nasa_draft_user
+    unless is_non_nasa_draft_user?(user: current_user, token: token)
+      clear_session_and_token_data
+
+      redirect_to root_url, flash: { error: "It appears you are not provisioned with the proper permissions to access the MMT for Non-NASA Users. Please try again or contact #{view_context.mail_to('support@earthdata.nasa.gov', 'Earthdata Support')}." }
```
jceaser: is this the exact same message as in collection_draft_proposals_controller? should this be in some sort of internationalization file so that you can reuse it? No is fine, just a suggestion for reducing duplication.

app/concerns/permission_checking.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+# These methods are used for checking Permissions (aka ACLs) to determine if
+# a user has access to System or Provider Identity Permissions
+# these will be used by the controllers as well as Pundit
+module PermissionChecking
+  extend ActiveSupport::Concern
+
+  def is_non_nasa_draft_user?(user:, token:)
```
cgokey: I don't see an "else" if all the conditions above are false

cgokey: I guess you don't. have to explicitly return false in rails

app/controllers/manage_metadata_controller.rb
charleshuang80: I think that even if I tried to use the en.yml file for the error message, I'd have to have it in there twice. I think the messages are tied to controllers by action.

app/concerns/permission_checking.rb
charleshuang80: No, I think it works without, but you're right it would be better to have it be there explicitly.

charleshuang80: Actually, I don't think there can be an else case here, it is just running through all the environments. The true or false is coming from the different `user_has_provider_permission_to` calls for the environments.

PR #414, opened 2019-09-18
MMT-1933 Updating UAT and Prod Config to enable templates.

ryanmiller-1:


jceaser:
we need to do a hot fix for uat.

jceaser:
bamboo is green, will merge for ryan as he is out.

PR #415, opened 2019-09-19
MMT-1258 Providing warning for users that Bulk Updates will convert r…

ryanmiller-1:
…ecords to UMM-JSON


jceaser:
questions on which button should be the default button can be addressed in future tickets.

PR #417, opened 2019-09-20
MMT-1920: fix configuration settings for MMT proper and Draft MMT to run on same release

charleshuang80:


spec/cmr/client_spec.rb
```diff
@@ -6,7 +6,8 @@
       env = Rails.configuration.cmr_env
       service_configs = Rails.configuration.services
       service_config = service_configs['earthdata'][env]
-      urs_client_id = service_configs['urs'][Rails.env.to_s][service_config['urs_root']]
+      mmt_mode = Rails.configuration.proposal_mode ? 'mmt_proposal_mode' : 'mmt_proper'
```
cgokey: Since this code is copies in 3 places, should you create a helper method called mmt_mode

charleshuang80: I would argue that it is a very small method that is in 3 very different places (the cmr client classes, the test for cmr_client, and the controllers) where it would be not as simple to have a helper method shared in all 3 locations.

cgokey: How about in environment.rb add this:
def self.mmt_mode
  Rails.configuration.proposal_mode ? 'mmt_proposal_mode' : 'mmt_proper'
end

And then use it all 3 places?

charleshuang80: Discussed this with Chris. Not sure I like the options of where to put the method, so going to leave it as is, but think about if there is a good place to refactor this.

PR #421, opened 2019-10-14
MMT-1993 Adding the progress view page for dMMT.

ryanmiller-1:
Added a route, controller action, and view for displaying the current status of a proposal on a separate page.  Added migration that includes request_type (not needed here, but will be needed next sprint) and status_history.  Updated status badge to be a link to this new page.  Added proposal status to index page as links to the new page.  Took confirmation modals used on the show page (for rescind and submit) and put them in a partial to be reused.  Added the rest of the states to AASM (but not events).  Added a spec helper to mock state transitions.  Used new helper to make some old tests a little more intuitive.


app/assets/stylesheets/overrides/_eui.scss
```diff
@@ -14,3 +14,7 @@
     color: #963b3b;
   }
 }
+
```
ryanmiller-1: The eui 1.1.7 change changed our formatting and this override returns us to what it used to look like.

app/controllers/proposal/collection_draft_proposals_controller.rb
```diff
@@ -62,11 +67,47 @@ def destroy
...
+      authorize get_resource
+
+      @status_history = get_resource.status_history || {}
+      approver_feedback = get_resource.approver_feedback || {}
+
+      if get_resource.in_work?
+        @first_stage = 'In Work'
+
+        # This applies the same validation as is applied on the preview/show page
+        show_view_setup
```
ryanmiller-1: I think it is inconsistent to make submit the only action that cannot be done on this page, so I think this validation is necessary.  I extracted it into a helper function in the original class.

app/helpers/proposals_helper.rb
```diff
@@ -0,0 +1,72 @@
+module ProposalsHelper
+  # For the functions relating to the nodes on the progress page, there are
+  # two classes which control the CSS that display the 'graphic' of the progress
+  # The node class controls whether the blue dots are faded (this stage hasn't
+  # been reached), visible (stage has been completed), or has a ring (current).
+  # The line class controls whether the lines are faded and their placement
+  # (since the active ones need to be placed slightly lower).
+
+  # Determine what the first node should look like on the progress page
+  def submit_node
```
ryanmiller-1: I extracted the if/elsif/elses from the view to try to clean it up and put them in this helper.  I tried to add comments to explain my logic, but I am unsure if what I wrote is really adding much.

app/views/collection_drafts/show.html.erb
```diff
@@ -46,7 +46,7 @@
           <% end %>
         <% end %>
 
-        <%= link_to "Delete #{name_to_title(resource_name)}", "#delete-draft-modal", class: 'display-modal' %>
+        <%= link_to "Delete #{name_to_title(resource_name)}", "#delete-draft-modal", class: 'display-modal eui-btn--link' %>
```
ryanmiller-1: As a consequence of the eui change, I noticed that this link was slightly off.  This fixes it.

app/assets/stylesheets/application.scss
```diff
@@ -71,6 +71,7 @@
 @import "components/cursor";
 @import "components/umm-version-label";
 @import "components/variable_generation";
+@import "components/timeline";
```
jceaser: did you say you tried the gem and then ended up doing it all by hand? Does this mean you need to remove this or you ended up using a part of it?

ryanmiller-1: I tried a prebuilt node component, but there are no remnants of it in this PR.  This is how we incorporate our css into sass.

app/views/proposal/collection_draft_proposals/progress.html.erb
```diff
@@ -0,0 +1,74 @@
+
+<div class="row row-content">
+  <section class="action wide-content">
+    <div class="wide-content-inside">
+      <span>
+        <span id="proposal-status-display-left" class="eui-badge eui-badge--split ">Proposal Status: <span><%= get_resource.proposal_status.titleize %></span></span>
+        <span class="progress-update"> Last Updated: <%= "#{get_resource.updated_at.utc} UTC" %></span>
```
charleshuang80: I would use the default_with_time_zone time format here as well

app/helpers/proposals_helper.rb
```diff
@@ -0,0 +1,72 @@
...
+    # If the proposal has been rejected and has not been submitted,
+    # Display the old rejection information.  At the time of the original
+    # implementation, this is the only time the active node should display its line
+    if @status_history['rejected'] && @status_history['submitted'].nil?
+      '<div class="timeline-node-active timeline-line-active"></div>'
```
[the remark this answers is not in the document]
charleshuang80: this may help also https://blog.bigbinary.com/2017/08/23/new-syntax-for-tag-helpers-in-rails-5-1.html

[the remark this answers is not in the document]
ryanmiller-1: I believe that the <%== %> is from a gem (erubis) and is why it did not need the html_safe.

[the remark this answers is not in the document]
ryanmiller-1: The <%== %> syntax that I was using is the equivalent of raw.

PR #422, opened 2019-10-15
MMT-1964: add Non-NASA Draft Approver Permissions

charleshuang80:
* add ACL to table
* update ACL tests
* add ACL check method
* add ACL check to pundit and before actions
* update tests


app/helpers/provider_identity_permissions_helper.rb
```diff
@@ -12,6 +12,7 @@ module ProviderIdentityPermissionsHelper
   GROUP_PERMISSIONS = %w[create read].freeze
   INGEST_MANAGEMENT_ACL_PERMISSIONS = %w[read update].freeze
   NON_NASA_DRAFT_USER_PERMISSIONS = %w[create read update delete].freeze
+  NON_NASA_DRAFT_APPROVER_PERMISSIONS = %w[create read update delete].freeze
```
ryanmiller-1: Can we swap the orders of these to alphabetize?

lib/tasks/acls.rake
```diff
@@ -92,6 +92,27 @@ namespace :acls do
 
       print_result(cmr_client.add_group_permissions(provider_perm, get_acls_token(admin: true)), '`Non-NASA Draft User` permissions added to the group.')
     end
+
+    desc 'Creates a group and grans permission for Non-NASA Draft Approver'
```
ryanmiller-1: grans => grant

app/controllers/proposal/collection_draft_proposals_controller.rb
```diff
@@ -4,7 +4,7 @@ class CollectionDraftProposalsController < CollectionDraftsController
 
     # TODO: Limit this to only the things a user is supposed to do, also need a new one for approver
     # TODO: Also need one for functions which both can perform (e.g. show)
-    before_action :ensure_non_nasa_draft_user
+    before_action :ensure_non_nasa_draft_permissions
```
ryanmiller-1: I think there is value in having a user_permissions and approver_permissions before action.  Specifically, these before actions can be used to ensure that users do not have access to approve, for instance.  Also, approvers do not need to be able to CRUD proposals.

app/helpers/provider_identity_permissions_helper.rb
charleshuang80: updated

lib/tasks/acls.rake
charleshuang80: fixed

app/controllers/proposal/collection_draft_proposals_controller.rb
charleshuang80: I do not disagree. However, those are not yet needed and without clearly knowing how they will be used I would rather do what is necessary for now, knowing it might need to change as we add functionality - which is why I left your comments above and added comments to the tests' configuration_helpers file.

PR #424, opened 2019-10-22
Mmt 1998 Remove Google Analytics Javascript

htranho:


jceaser:
there was a question on testing this code, you can check the network inspector in the browser (safari works here) for hits going out to google. They drop off in this version.

vendor/assets/javascripts/eui-1.0.0/eui.js
```diff
@@ -183,16 +183,7 @@
     });
   });
 }(jQuery));
-;(function () {
```
charleshuang80: I guess it is too late now, but I don't think this should have been removed. This seems to be the google analytics info from EUI since it is in their JS package, not from MMT. But at the same time, I think we are trying to remove this package to use the more recent version so I guess it is not a big deal.

PR #426, opened 2019-10-24
MMT-1989 Create a delete metadata proposal request

ryanmiller-1:
Added the ability to create and rescind a delete metadata proposal request.  Updated the look of the badges per PO feedback (to include the request type).  Updated permissions for search/collections pages in dMMT.

### DIFF
--- a/app/concerns/permission_checking.rb
+++ b/app/concerns/permission_checking.rb
@@ -40,11 +40,11 @@ module PermissionChecking
     is_non_nasa_draft_user?(user: user, token: token) || is_non_nasa_draft_approver?(user: user, token: token)
   end
 
-  def authorize_proposal_all_user_actions(user:, token:)
+  def proposal_mode_all_user_actions_allowed?(user:, token:)
     redirect_to manage_collection_proposals_path unless Rails.configuration.proposal_mode && either_non_nasa_user_or_approver?(user: user, token: token)
   end
 
-  def authorize_public_actions(user:, token:)
+  def multi_mode_actions_allowed?(user:, token:)
     redirect_to manage_collection_proposals_path unless !Rails.configuration.proposal_mode || (Rails.configuration.proposal_mode && either_non_nasa_user_or_approver?(user: user, token: token))
   end
 
--- a/app/concerns/permission_checking.rb
+++ b/app/concerns/permission_checking.rb
@@ -36,6 +36,18 @@ module PermissionChecking
     end
   end
 
+  def either_non_nasa_user_or_approver?(user:, token:)
+    is_non_nasa_draft_user?(user: user, token: token) || is_non_nasa_draft_approver?(user: user, token: token)
+  end
+
+  def authorize_proposal_all_user_actions(user:, token:)
+    redirect_to manage_collection_proposals_path unless Rails.configuration.proposal_mode && either_non_nasa_user_or_approver?(user: user, token: token)
+  end
+
+  def authorize_public_actions(user:, token:)
+    redirect_to manage_collection_proposals_path unless !Rails.configuration.proposal_mode || (Rails.configuration.proposal_mode && either_non_nasa_user_or_approver?(user: user, token: token))
+  end
+
   def user_has_provider_permission_to(user:, action:, target:, token:, specific_provider: nil)
     user_has_permission_to(user: user, action: action, target: target, type: 'provider', token: token, specific_provider: specific_provider)
   end
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -3,7 +3,7 @@ class CollectionsController < ManageCollectionsController
   include CMRCollectionsHelper
 
   before_action :set_collection
-  before_action :ensure_correct_collection_provider, only: %i[edit clone revert destroy]
+  before_action :ensure_correct_collection_provider, only: [:edit, :clone, :revert, :destroy]
 
   layout 'collection_preview', only: [:show]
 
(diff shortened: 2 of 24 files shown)


app/concerns/permission_checking.rb
```diff
@@ -36,6 +36,18 @@ def is_non_nasa_draft_approver?(user:, token:)
...
 
+  def either_non_nasa_user_or_approver?(user:, token:)
+    is_non_nasa_draft_user?(user: user, token: token) || is_non_nasa_draft_approver?(user: user, token: token)
+  end
+
+  def authorize_proposal_all_user_actions(user:, token:)
+    redirect_to manage_collection_proposals_path unless Rails.configuration.proposal_mode && either_non_nasa_user_or_approver?(user: user, token: token)
+  end
+
+  def authorize_public_actions(user:, token:)
```
charleshuang80: this seems slightly imprecise because the actions aren't 'public' (and because the method isn't doing the authorizing). 

I would move the 'allowed' or 'authorized' to the end again, and then use something that denotes 'both modes' or 'all modes' or 'multi mode', so might end up with something more like `all_modes_action_authorized?`

app/controllers/collections_controller.rb
```diff
@@ -98,6 +94,13 @@ def download_xml
     send_data collection_download, type: content_type, disposition: "attachment; filename=#{concept_id}.#{download_format}"
   end
 
+  def delete_proposal
```
charleshuang80: This reads like it should be deleting something, but we are creating a delete proposal, so I think it should be `create_delete_proposal`

app/controllers/collections_controller.rb
```diff
@@ -3,11 +3,7 @@ class CollectionsController < ManageCollectionsController
   include CMRCollectionsHelper
 
   before_action :set_collection
-  before_action :ensure_correct_collection_provider, only: [:edit, :clone, :revert, :destroy]
-
-  skip_before_action :proposal_mode_enabled?, only: [:show]
-  # the only functionality that should be available in Draft MMT should be
-  # to view the collection preview
+  before_action :ensure_correct_collection_provider, only: %i[edit clone revert destroy]
```
charleshuang80: I know rubocop was complaining to change this to use `%i`, but I think we made a decision a while ago to keep before_actions without that, because that is how Rails has been doing it for quite a while.

app/controllers/collections_controller.rb
```diff
@@ -153,4 +156,17 @@ def set_collection_error_data
     @collection_format = ''
     @error = 'This collection is not available. It is either being published right now, does not exist, or you have insufficient permissions to view this collection.'
   end
+
+  def proposal_mode_enabled?
+    # actions available in both dMMT and MMT
```
charleshuang80: this may just be a preference thing, but it seems less confusing if the comment is nested under the `if` (and similar for the other comments here)

app/controllers/proposal/collection_draft_proposals_controller.rb
```diff
@@ -38,9 +37,20 @@ def submit
...
 
-      remove_status_history('submitted')
-
-      if get_resource.rescind && get_resource.save
+      # An in_work delete request does not make sense.  If a delete request is
+      # rescinded, delete the request instead.
+      if get_resource.request_type == 'delete'
+        short_name = get_resource.draft['ShortName']
+        if get_resource.rescind && get_resource.destroy
+          Rails.logger.info("Audit Log: #{resource_name.titleize} #{get_resource.entry_title} was rescinded and destroyed by #{current_user.urs_uid}")
```
charleshuang80: per our conversation today, we should also capture the type of request

app/views/proposal/collection_draft_proposals/progress.html.erb
```diff
@@ -3,7 +3,9 @@
   <section class="action wide-content">
     <div class="wide-content-inside">
       <span>
-        <span id="proposal-status-display-left" class="eui-badge eui-badge--split ">Proposal Status: <span><%= get_resource.proposal_status.titleize %></span></span>
+        <span class="eui-badge eui-badge--split proposal-status-badge"><%= get_resource.request_type == 'create' ? 'Draft Proposal Submission:' : "#{get_resource.request_type.titleize} Metadata Request:" %>
```
charleshuang80: I had suggested seeing if the text in the badge could be underlined, and if that made it look more like a link.

spec/features/proposal/collection_draft_proposals/collection_draft_proposal_creation_spec.rb
```diff
@@ -76,4 +76,20 @@
       end
     end
   end
+
+  context 'when viewing a collection' do
```
charleshuang80: We will also want to create an update request from viewing a collection. Therefore, either we can add text to this line to say something like 'when viewing a collection to create a delete metadata proposal', _or_ we can add another context on/before line 87 for the delete request so the before block can be used later for update requests also.

spec/features/proposal/collection_draft_proposals/collection_draft_proposal_submit_spec.rb
```diff
@@ -97,11 +94,40 @@
...
+  context 'when looking at a delete metadata request' do
+    before do
+      set_as_proposal_mode_mmt(with_required_acl: true)
+      @collection_draft_proposal = create(:full_collection_draft_proposal, draft_request_type: 'delete')
+      mock_submit(@collection_draft_proposal)
+      visit collection_draft_proposal_path(@collection_draft_proposal)
+    end
+
+    it 'can be rescinded and deleted' do
+      short_name = @collection_draft_proposal['short_name']
```
charleshuang80: discussed having these actions in a `before` block, to keep a consistent structure for our tests.

app/models/collection_draft_proposal.rb
```diff
@@ -31,6 +31,26 @@ class CollectionDraftProposal < CollectionDraft
     end
   end
 
+  class << self
+    def create_request(collection, user, native_id, type, username = nil)
```
cgokey: I'm just curious what does the class << self syntax do in rails?

PR #427, opened 2019-10-24
CMRQ-2029 Add detailed logging in Bulk Update

htranho:


charleshuang80:
@rosycordova if you are satisfied with this, can you approve? I am fine with it, but I think you should weigh in with whether this satisfies your needs.

rosycordova:
The information requested was added to the log.  I'll need to query splunk once the changes are deployed.

PR #433, opened 2019-11-01
MMT-1989 Updating request type to be validated when a user loads a page or saves data.  Migrating DB to populate request_type.

ryanmiller-1:
Migrates the DB for all nil values of request_type.
Validates on save.


app/controllers/proposal/collection_draft_proposals_controller.rb
```diff
@@ -155,5 +156,13 @@ def get_progress_message(action)
       end
       "#{action == 'done' ? 'Published' : action.titleize}: #{action_time} By: #{action_username}"
     end
+
+    def validate_proposal
```
charleshuang80: As long as the migration is run and the model validation runs, this before action should not be necessary, right?

ryanmiller-1: This protects against other sources of bad data as well; humans mucking about in the DB, rails calls that don't respect the validation rules (e.g. update_all).

spec/models/collection_draft_proposal_spec.rb
```diff
@@ -210,10 +210,10 @@
     # not be possible if the other actions are successfully blocked
 
     it 'creating a draft proposal should fail to save' do
```
charleshuang80: do we have a test in this file that tests creating a proposal without the request_type?

PR #435, opened 2019-11-07
MMT-2044 Adding the Manage Proposals tab, tab highlighting, view, and basic logic for displaying approved proposals in MMT proper.

ryanmiller-1:


app/views/manage_proposal/show.html.erb
```diff
@@ -0,0 +1,64 @@
+<div class="row row-content manage-proposals-mmt">
+  <section>
+    <h2><%= "#{@category_of_displayed_proposal} Proposals" %></h2>
```
charleshuang80: This seems fine, but I'm not sure that it needs to be changeable

spec/features/proposal/access_permissions/no_permissions_for_draft_mmt_spec.rb
```diff
@@ -158,3 +158,38 @@
...
+    end
+  end
+
+  context 'when trying to access the manage proposals page directly' do
+    before do
+      visit manage_proposals_path
+    end
+
+    it 'the user cannot access the Manage Proposals tab' do
+      expect(page).to have_link('Manage Collections')
```
charleshuang80: I would maybe add a check for the highlighting under Manage Collections to verify that's where we got redirected to

spec/features/proposal/access_permissions/no_permissions_for_draft_mmt_spec.rb
```diff
@@ -158,3 +158,38 @@
     end
   end
 end
+
+describe 'No Permissions for Non-NASA Workflow in MMT', js: true do
```
charleshuang80: since this is testing what is happening in MMT Proper, I think you should make a folder for 'Manage Proposals', and move all these new permissions tests into their own files.

spec/features/manage_proposals/manage_proposal_view_spec.rb
```diff
@@ -0,0 +1,120 @@
+describe 'Proposals listed on the Manage Proposals (MMT) page', js: true do
```
cgokey: remove js:true?

spec/features/manage_proposals/no_permissions_for_draft_mmt_spec.rb
```diff
@@ -0,0 +1,42 @@
+describe 'No Permissions for Non-NASA Workflow in MMT', js: true do
```
cgokey: likewise here, I think it slows down the tests.

spec/features/manage_proposals/manage_proposal_view_spec.rb
ryanmiller-1: The four files in features/manage_proposals require the js:true because it is either manipulating modals or the tab highlighting (features we are testing).

PR #438, opened 2019-11-12
MMT-2032 -- Fix Rails 5.1 test case failures due to gem incompatabili…

jdf-eos:
Updated gem files for a consistent rails 5.1 gem set. 

Modified code for Faker upgrade breakage.


config/environments/test.rb
```diff
@@ -45,7 +45,7 @@
   config.action_mailer.default_url_options = { host: 'localhost', port: '3000' }
 
   # Randomize the order test cases are executed.
-  config.active_support.test_order = :random
+  config.active_support.test_order = :sorted
```
charleshuang80: was this a necessary change?

spec/factories/collection_template.rb
```diff
@@ -26,10 +26,10 @@
 
     draft {
       collection_one.merge(
-        'TemplateName' => template_name || "#{Faker::Number.number(6)}_#{Faker::Superhero.name}",
-        'ShortName'    => draft_short_name || "#{Faker::Number.number(6)}_#{Faker::Superhero.name}",
+        'TemplateName' => template_name || "#{Faker::Number.number(digits: 6)}_#{Faker::Superhero.name}",
```
charleshuang80: I'm glad you took care of this. I wanted to make this change in my other PR but I felt it would have blown it up, in the sense that there would be so many extraneous files to look at.

config/environments/test.rb
jdf-eos: No.   But it makes it easier to do diffs on different versions of the Rspec output. With random, it isn't simple to do this.   I can change it back if you feel that is better.

charleshuang80: I think it makes sense to leave it as you have it while working through all the upgrades. As long as we change it back once we have everything upgraded and working that should be good.

jdf-eos: That sounds right. Thanks.

PR #442, opened 2019-11-15
Mmt 1974 - Fetch Approved Proposals from dMMT

ryanmiller-1:
Added a new controller to handle requests related to approved proposals in dMMT.
Updated manage proposals controller in MMT to handle real data instead of dummy data.
Added monkey-patch to enable dev/test to use one instance of the server.
Added dMMT client for handling requests from MMT to dMMT.
Added URS token validation for dMMT to use tokens from MMT.
Updated tests for Manage Proposals to use real data instead of dummy data.
Added tests to verify that approved proposals is mostly inaccessible in a browser.


lib/cmr/dmmt_client.rb
```diff
@@ -0,0 +1,20 @@
+module Cmr
+  class DmmtClient < BaseClient
+    def dmmt_get_approved_proposals(params, token, host, port)
+      url = if Rails.env.development?
+              "http://localhost:3000/approved_proposals"
```
ryanmiller-1: Investigate alternatives

lib/cmr/base_client.rb
```diff
@@ -7,6 +7,8 @@ class BaseClient
     # include Cmr::QueryTransformations
     include Cmr::Util
 
+    # who cares about this client ID?  As a consequence, do we need to change it
+    # to include dmmt?
```
ryanmiller-1: Investigate who consumes this

app/controllers/manage_proposal_controller.rb
```diff
@@ -8,29 +8,30 @@ def show
...
-    # TODO: When we have the end points set up, this will probably be receiving
-    # a JSON blob that gets converted to an array of records. If this is true,
-    # dummy_data should be replacable with that array.
-    proposals = if sort_dir == 'ASC'
-                  dummy_data.sort { |a, b| a[sort_key] <=> b[sort_key] }
-                else
-                  dummy_data.sort { |a, b| b[sort_key] <=> a[sort_key] }
-                end
-
+    dmmt_response = cmr_client.dmmt_get_approved_proposals({}, token, request.ip, request.port)
```
ryanmiller-1: Pass whole request object?

app/controllers/manage_proposal_controller.rb
```diff
@@ -8,29 +8,30 @@ def show
...
+      proposals = if sort_dir == 'ASC'
+                    dmmt_response.body['proposals'].sort { |a, b| a[sort_key] <=> b[sort_key] }
+                  else
+                    dmmt_response.body['proposals'].sort { |a, b| b[sort_key] <=> a[sort_key] }
+                  end
+    else
+      if dmmt_response.status == 401
+        flash[:error] = "Your token could not be authorized. Please try refreshing the page before contacting #{view_context.mail_to('support@earthdata.nasa.gov', 'Earthdata Support')} about #{request.uuid}."
+        Rails.logger.error("#{request.uuid}: A user who was logged in with Non-NASA draft approver privileges was not authenticated or authorized correctly.  Refer to dMMT logs for further information: #{dmmt_response.body['request_id']}")
+      else
```
ryanmiller-1: elsif

app/controllers/manage_proposal_controller.rb
```diff
@@ -8,29 +8,30 @@ def show
...
+
+    if dmmt_response.success?
+      Rails.logger.info("MMT successfully received approved proposals from dMMT at #{current_user.urs_uid}'s request.")
+      proposals = if sort_dir == 'ASC'
+                    dmmt_response.body['proposals'].sort { |a, b| a[sort_key] <=> b[sort_key] }
+                  else
+                    dmmt_response.body['proposals'].sort { |a, b| b[sort_key] <=> a[sort_key] }
+                  end
+    else
+      if dmmt_response.status == 401
```
ryanmiller-1: More human readable?

spec/features/manage_proposals/manage_proposal_view_spec.rb
```diff
@@ -4,31 +4,53 @@
...
-#
-#    it 'has the appropriate text' do
-#      expect(page).to have_content('No Approved Proposals found')
-#    end
-#  end
-
-  context 'when using the dummy data' do
+  context 'when the user can be authenticated, has approver permissions, and there are two pages of proposals' do
     before do
+      set_as_proposal_mode_mmt
```
ryanmiller-1: Maybe monkey path the controller to shorten this test?

ryanmiller-1: More thinking doesn't really change my thoughts here; I think readability is more important than the ~10s that it could save us.

lib/cmr/dmmt_client.rb
ryanmiller-1: Anything involving https gives certification verification errors.
mmt.localtest fails to connect on 80.
Mimicking the test set up gives a bad request error because a plain http request is sent to port 443.

config/environments/test.rb
```diff
@@ -16,7 +16,16 @@
   # Do not eager load code on boot. This avoids loading your whole application
   # just for the purpose of running a single test. If you are using a tool that
   # preloads Rails for running tests, you may have to set it to true.
-  config.eager_load = false
+  # config.eager_load = false
+
+  # This configuration change is needed to permit dMMT and MMT to talk to each
```
charleshuang80: I would maybe add a 'NOTE:' or 'TODO:' with these comments as an additional reminder to come back to it later.

app/controllers/proposal/approved_proposals_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+    def approved_proposals
+      passed_token = request.headers.fetch('Echo-Token', ':').split(':')[0]
+
+      # Navigate a browser elsewhere.
+      redirect_to root_path and return if passed_token.blank?
+
+      token_response = cmr_client.validate_token(passed_token)
+
+      if token_response.success?
+        authenticated_user = User.new(urs_uid: token_response.body['uid'])
```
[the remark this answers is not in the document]
ryanmiller-1: This is not saving the user.  The call on the following line expects a user object (or more specifically, an object that responds to .urs_uid).

lib/cmr/urs_client.rb
```diff
@@ -52,6 +52,15 @@ def associate_urs_uid_and_auid(urs_uid, auid)
       response
     end
 
+    # Function to allow dMMT to authenticate a token from MMT.
+    def validate_token(token)
+      services = Rails.configuration.services
+      config = services['earthdata'][Rails.configuration.cmr_env]
+      mmt_client_id = services['urs']['mmt_proper'][Rails.env.to_s][config['urs_root']]
```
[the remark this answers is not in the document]
ryanmiller-1: done

config/environments/test.rb
ryanmiller-1: done

app/controllers/proposal/approved_proposals_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+    # environments.
+
+    skip_before_action :ensure_user_is_logged_in
+    skip_before_action :refresh_urs_if_needed, :refresh_launchpad_if_needed, :provider_set?, :proposal_approver_permissions
+
+    def approved_proposals
+      passed_token = request.headers.fetch('Echo-Token', ':').split(':')[0]
+
+      # Navigate a browser elsewhere.
+      redirect_to root_path and return if passed_token.blank?
```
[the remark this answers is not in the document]
ryanmiller-1: Initially I tried a respond_to block, but that still permits things like approved_proposals.json

lib/cmr/dmmt_client.rb
ryanmiller-1: The next step, to me, is to investigate self-signed certs for dev.  Is that something we want to pursue?

ryanmiller-1:
Force push to acquire fix for MMT-2044 in master

lib/cmr/dmmt_client.rb
charleshuang80: My opinion is that it is not necessary. I think we may want to add a comment that you may need to change the port depending on if you run rails s on a different port.

spec/features/proposal/approved_proposals/approved_proposals_visibility_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  context 'when logged into MMT with approver and user permissions' do
+    before do
+      login
+      allow_any_instance_of(PermissionChecking).to receive(:is_non_nasa_draft_approver?).and_return(true)
+      allow_any_instance_of(PermissionChecking).to receive(:is_non_nasa_draft_user?).and_return(true)
+      visit approved_proposals_approved_proposals_path
+    end
+
+    it 'cannot reach the approved_proposals page' do
```
charleshuang80: so if a user has approver and user permissions, they should not be able to reach the page? just verifying that is the desired behavior.

spec/features/manage_proposals/manage_proposal_view_spec.rb
charleshuang80: couldn't you use the `with_draft_approver_acl` option here?

app/controllers/proposal/approved_proposals_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    # environments.
+
+    skip_before_action :ensure_user_is_logged_in
+    skip_before_action :refresh_urs_if_needed, :refresh_launchpad_if_needed, :provider_set?, :proposal_approver_permissions
+
+    def approved_proposals
+      token_and_client_id = request.headers.fetch('Echo-Token', ':').split(':')
+      passed_token = token_and_client_id[0]
+      passed_client_id = token_and_client_id[1]
+
```
cgokey: You can say passed_token, passed_client_id = ...
so you don't need to intermediate variable

app/controllers/proposal/approved_proposals_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      token_response = cmr_client.validate_token(passed_token, passed_client_id)
+
+      if token_response.success?
+        authenticated_user = User.new(urs_uid: token_response.body['uid'])
+        requester_has_approver_permissions = is_non_nasa_draft_approver?(user: authenticated_user, token: passed_token)
+      else
+        requester_has_approver_permissions = false
+      end
+
```
cgokey: rubocopy didn't complain here?   
requester_has_approver_permissons = if ... else ...

spec/features/proposal/approved_proposals/approved_proposals_visibility_spec.rb
ryanmiller-1: That is the behavior that I intend.  I don't intend the /approved_proposals end point to be user accessible.

charleshuang80: oh right, I was confusing this with the Manage Proposals tab

app/controllers/proposal/approved_proposals_controller.rb
ryanmiller-1: No, the first part of the if has two statements.  I've reworked it to address that though.

app/controllers/proposal/approved_proposals_controller.rb
ryanmiller-1: I thought that a, b = ['a', 'b', 'c'] would yield a = 'a' and b = ['b', 'c'], but I was mistaken.  Changed.

spec/features/manage_proposals/manage_proposal_view_spec.rb
ryanmiller-1: done

app/controllers/proposal/approved_proposals_controller.rb
cgokey: I think you can do something like this, correct if me I'm wrong:

```
    requester_has_approver_permissions = if token_response.success?
                                           authenticated_user = User.new(urs_uid: token_response.body['uid'])
                                           is_non_nasa_draft_approver?(user: authenticated_user, token: passed_token)
                                         else
                                           false
                                         end
```

ryanmiller-1: I think that would work too, but I changed the later statement that was relying on the authenticated user, so this has been changed slightly differently.
Another way of stating the first sentence of my other comment is: I think rubocop only balks in the most trivial cases.  I didn't mean to imply it wouldn't work.

PR #445, opened 2019-11-21
MMT-1984: add ability to create a metadata update proposal

charleshuang80:
* add link, modal, and controller actions for creating an update proposal
* move methods out of controller into model
* add tests

### DIFF
--- a/app/views/proposal/collections/show.html.erb
+++ b/app/views/proposal/collections/show.html.erb
@@ -17,12 +17,22 @@
     <section class="action wide-content">
       <div class="wide-content-inside">
         <span>
+          <%= link_to 'Create Update Proposal', '#update-request-modal', class: 'display-modal eui-btn--link bar-after' %>
           <%= link_to 'Submit Delete Request', '#delete-request-modal', class: 'display-modal eui-btn--link' %>
         </span>
       </div>
     </section>
 
 
+    <div id="update-request-modal" class="eui-modal-content">
+      <a href="javascript:void(0);" class="modal-close float-r"><i class="fa fa-times"></i><span class="is-invisible">Close</span></a>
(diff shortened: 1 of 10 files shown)


app/views/proposal/collections/show.html.erb
```diff
@@ -17,12 +17,22 @@
     <section class="action wide-content">
       <div class="wide-content-inside">
         <span>
+          <%= link_to 'Create Update Proposal', '#update-request-modal', class: 'display-modal eui-btn--link bar-after' %>
```
charleshuang80: I checked with the PO, she prefers these actions to stay as links

spec/features/proposal/collection_draft_proposals/collection_draft_proposal_update_spec.rb
```diff
@@ -1,4 +1,4 @@
-describe 'Collection Draft Proposal Update', js: true do
+describe 'Collection Draft Proposal Update', reset_provider: true, js: true do
```
cgokey: I mentioned this in of Ryan's pull requests, does adding/keeping js:true slow down the tests in your opinion?   I thought I noticed a significant difference one day when I was testing.

charleshuang80: Yes it does. I think we should remove it where possible, but if your integration test requires things that the js is doing, then it is necessary. I have also found it necessary for capybara screenshots, but that can be used temporarily.

charleshuang80: this test _might not_ need `js: true` but I'd rather not change it at the moment. with one of the future draft mmt tickets, I'll review if all the tests for proposals need it or not.

PR #448, opened 2019-12-05
MMT-2000 Odds and ends update based on PO walkthrough

ryanmiller-1:
Did a find and replace for the CSS on the progress page: timeline-faded-node => timeline-node-faded for consistency with other classes.  Similarly for timeline-faded-line => timeline-line-faded.
Changed a before action that was checking approver ACLs to also happen in proposal mode
Made a series of changes to reduce how often we were going to CMR to check approver acls
Added a badge that displays approver/user status in dMMT
Added submitter column to index pages (sorting deferred)
Added ingest error display when publishing a proposal fails
Made a series of changes to maintain output consistency (e.g. cancel, word order)
Updated variable names draft => proposal in proposal files

### DIFF
--- a/app/concerns/permission_checking.rb
+++ b/app/concerns/permission_checking.rb
@@ -42,7 +42,7 @@ module PermissionChecking
 
   # shortcut helper methods
   def approver?
-    is_non_nasa_draft_approver?(user: current_user, token: token)
+    @user_has_approver_permissions || is_non_nasa_draft_approver?(user: current_user, token: token)
   end
 
   def non_nasa_user?
--- a/app/views/manage_proposal/show.html.erb
+++ b/app/views/manage_proposal/show.html.erb
@@ -1,4 +1,21 @@
 <div class="row row-content manage-proposals-mmt">
+  <% if @ingest_errors && !@ingest_errors.empty? %>
+    <section class="errors">
+      <div class="eui-banner--danger">
+        <div class="eui-banner__message">
+          <h4><i class="fa fa-exclamation-triangle"></i> This proposal has the following errors:</h4>
+          <ul class="no-bullet">
+            <% @ingest_errors.each do |error| %>
+              <li>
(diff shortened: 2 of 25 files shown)


ryanmiller-1:
Force push to rebase with 1962 in master.  Additional commit was added to make tests added in that ticket consistent with changes in this one and to address a spacing issue identified in PR.

app/concerns/permission_checking.rb
```diff
@@ -42,7 +42,7 @@ def is_non_nasa_draft_approver?(user:, token:)
 
   # shortcut helper methods
   def approver?
-    is_non_nasa_draft_approver?(user: current_user, token: token)
+    @user_has_approver_permissions || is_non_nasa_draft_approver?(user: current_user, token: token)
```
jdf-eos: Aren't these 2 operands identical?

ryanmiller-1: The right side goes out to CMR to check if they have the ACL, but in some cases we will have already checked and set it.  If so, we save a call to the CMR.

app/views/manage_proposal/show.html.erb
```diff
@@ -1,4 +1,21 @@
 <div class="row row-content manage-proposals-mmt">
+  <% if @ingest_errors && !@ingest_errors.empty? %>
+    <section class="errors">
+      <div class="eui-banner--danger">
+        <div class="eui-banner__message">
+          <h4><i class="fa fa-exclamation-triangle"></i> This proposal has the following errors:</h4>
+          <ul class="no-bullet">
+            <% @ingest_errors.each do |error| %>
```
jceaser: Haven't we seen examples of this where the errors are more then one screen's worth? Going forward what if we limited to the first 20 or so errors?

jceaser: I"m thinking this should be a future ticket

PR #449, opened 2019-12-11
Mmt 2034

jdf-eos:
Hopefully, this is the last round of bug fixes for the Rails 5.1 baseline of MMT.

### DIFF
--- a/app/controllers/conversions_controller.rb
+++ b/app/controllers/conversions_controller.rb
@@ -14,7 +14,7 @@ class ConversionsController < ApplicationController
     Rails.logger.info "converting shapefile #{request.headers["Content-Type"]}"
 
     # Define local_path method required by Faraday
-    upload = params.require(:upload)
+    upload = safe_hash(:upload)
     def upload.local_path
       @tempfile
     end
--- a/app/controllers/conversions_controller.rb
+++ b/app/controllers/conversions_controller.rb
@@ -19,7 +19,7 @@ class ConversionsController < ApplicationController
       @tempfile
     end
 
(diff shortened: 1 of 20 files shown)


app/controllers/conversions_controller.rb
```diff
@@ -14,14 +14,15 @@ def convert
...
       @tempfile
     end
 
-    response = OgreClient.convert_shapefile(params)
+    response = OgreClient.convert_shapefile(params.permit!.to_h)
+    
 
     # render(json: current_user.to_fileupload(name, style), content_type: request.format)
-    render text: response.body, content_type: request.format
+    render :json => response.body
```
ryanmiller-1: I think render json: response.body is the preferred syntax.

jdf-eos: This produces a syntax error:
SyntaxError
/Users/jdfarley/Work/mmt/app/controllers/conversions_controller.rb:26: syntax error, unexpected tIDENTIFIER, expecting keyword_end
    render :json response.body
                 ^~~~~~~~
Extracted source (around line #26):

It took several tries to find "render :json => response.body" worked correctly.   render has gone through some changes in Rails 5

Thoughts?

ryanmiller-1: Does 'json:' cause a syntax error?  json: response.body should be interpreted as a hash here.

jdf-eos: Sorry,  mis-typed!  render JSON: response.body  does work.  I'll incorporate this change.

PR #450, opened 2019-12-11
MMT-2028 Updating logging for create/update of drafts for metrics.

ryanmiller-1:


ryanmiller-1:
Force push to update with rebase so that it can build in bamboo (needed datepicker test changes)

PR #453, opened 2019-12-19
MMT-2001 Adding cron to delete in work proposals last updated > 30 days ago.

ryanmiller-1:


config/schedule.rb
```diff
@@ -1,3 +1,5 @@
+require File.expand_path(File.dirname(__FILE__) + "/environment")
```
ryanmiller-1: Added to make the check for proposal mode

lib/tasks/proposals.rake
```diff
@@ -0,0 +1,7 @@
+namespace :delete_proposals do
+  desc 'Delete proposals that are older than specified date old'
+  task :expired_proposals => :environment do
```
charleshuang80: `expired_proposals: :environment`

lib/tasks/proposals.rake
```diff
@@ -0,0 +1,7 @@
...
+    count = CollectionDraftProposal.delete_all("proposal_status = 'in_work' AND updated_at < '#{30.days.ago}'") if Rails.configuration.proposal_mode
```
charleshuang80: This selection was verified with the PO, right?

ryanmiller-1: Yes, another ticket was written to capture the work for other cases.

jceaser: There is no test for this, which may be hard to do, but I got a demo and I think the logic is good. updated_at < 30 days ago when looking at timestamp would work as expected.

PR #456, opened 2019-12-20
hotfix/MMT-2001 updating where proposal mode is checked to enable dep…

ryanmiller-1:
…loyment

### DIFF
--- a/lib/tasks/proposals.rake
+++ b/lib/tasks/proposals.rake
@@ -1,7 +1,9 @@
 namespace :delete_proposals do
   desc 'Delete proposals that are older than specified date old'
   task expired_proposals: :environment do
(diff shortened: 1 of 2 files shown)


lib/tasks/proposals.rake
```diff
@@ -1,7 +1,9 @@
 namespace :delete_proposals do
   desc 'Delete proposals that are older than specified date old'
   task expired_proposals: :environment do
-    count = CollectionDraftProposal.delete_all("proposal_status = 'in_work' AND updated_at < '#{30.days.ago}'") if Rails.configuration.proposal_mode
-    Rails.logger.info("Automated check for proposals that are 'in_work' and 30 days old deleted #{count || 0} proposals.")
+    if Rails.configuration.proposal_mode
```
charleshuang80: Is rubocop complaining that you should use a guard clause? I either think this way _or_ the guard clause would be fine, was just thinking we should change it if rubocop is complaining

PR #457, opened 2019-12-27
MMT-2057 - MMT in Docker

jceaser:
Adding a docker file and a helper script called harbor.sh


jceaser:
Adding a lot of people here because I felt others may want too see the technology

jceaser:
I'll add the --system note

PR #460, opened 2020-01-06
Mmt 2035

jdf-eos:


config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -0,0 +1,38 @@
+# Be sure to restart your server when you modify this file.
```
ryanmiller-1: Did you test any of the settings in this file?  If they're commented out and the application.rb file is loading 5.1 defaults, are we done?

jdf-eos: WRT new  defaults, I have been writing tickets for new items which may be useful such as the enhanced security and CSRF.   For optional items, I tend to keep the current behavior unless there is a clear reason to change, or if the item comes with a deprecation. If deprecated, we are going to make the change eventually so I made the change as part of this upgrade.   For this file, I didn't see a reason to make any of these changes at this time.  Let me know if you see an item that should be used.

PR #461, opened 2020-01-06
MMT-1719 Adding e-mail to reject proposal.

ryanmiller-1:


app/views/proposal_mailer/proposal_rejected_notification.text.erb
```diff
@@ -0,0 +1,6 @@
+<%= @user[:name] %>,
+
+<%= "The collection draft proposal #{@short_name}_#{@version} has been reviewed and rejected for the following reasons: '#{@rejected_reasons}'." %>
+<%= "Additional feedback was provided: '#{@rejected_note}'" %>
```
jdf-eos: Should the "Additional feedback" line be conditional based on rejected_note being non-blank?

Also, it looks like there should be some spacing between line 3 and line 4,  as you did in the html file.

ryanmiller-1: The note is not currently optional when rejecting.

I closed this PR because it was the wrong ticket number.  I will update the spacing in the correct PR.

PR #462, opened 2020-01-06
MMT-1718 Adding e-mail to reject proposal.

ryanmiller-1:


ryanmiller-1:
John reviewed this when I had it with the wrong ticket number.  (See: https://github.com/nasa/mmt/pull/461 ).  I am content to say this one has two reviews, without objections.

app/mailers/proposal_mailer.rb
```diff
@@ -20,4 +20,18 @@ def proposal_approved_notification(user, short_name, version, id)
 
     mail(to: "#{@user[:name]} <#{@user[:email]}>", subject: email_subject)
   end
+
+  def proposal_rejected_notification(user, proposal)
+    @user = user
```
cgokey: I like how you just pass the proposal object into this rather than all the individual fields, why kind of wondering if you should have done that for the other methods.

ryanmiller-1:
Force pushed because of a rebase to capture MMT-2067's submitter_id column being added.

PR #463, opened 2020-01-07
MMT-1719 adding emails for approvers to submission.

ryanmiller-1:


app/controllers/proposal/collection_draft_proposals_controller.rb
```diff
@@ -101,6 +101,10 @@ def submit
         Rails.logger.info("Audit Log: User #{current_user.urs_uid} successfully submitted #{resource_name.titleize} with title: '#{get_resource.entry_title}' and id: #{get_resource.id} (a #{get_resource.request_type} metadata request).")
 
         ProposalMailer.proposal_submitted_notification(get_user_info, get_resource.draft['ShortName'], get_resource.draft['Version'], params['id']).deliver_now
+        approver_users = get_approver_emails(ENV['approvers_urs'].split(','))
```
charleshuang80: call the method without params

app/controllers/proposal/collection_draft_proposals_controller.rb
```diff
@@ -217,6 +221,15 @@ def user_from_resource
       { name: "#{proposal_urs_user['first_name']} #{proposal_urs_user['last_name']}", email: proposal_urs_user['email_address'] } unless proposal_urs_user.blank?
     end
 
+    def get_approver_emails(approvers_urs)
+      approver_urs_users = retrieve_urs_users(approvers_urs)
```
charleshuang80: since you are using an env var, you can just use it here

app/controllers/proposal/collection_draft_proposals_controller.rb
ryanmiller-1: done

app/controllers/proposal/collection_draft_proposals_controller.rb
ryanmiller-1: done

ryanmiller-1:
Force pushed to capture rebase to master to fix the datepicker tests.

PR #468, opened 2020-01-13
Feature/mmt 1598

jdf-eos:


config/application.rb
```diff
@@ -17,6 +18,9 @@
 
 module Mmt
   class Application < Rails::Application
+    # Initialize configuration defaults for originally generated Rails version.
+    config.load_defaults 5.1
```
ryanmiller-1: This setting controls the same things as the initializers 'new_framework_defaults...'.  In my mind, one of two things is happening:
1. This setting is overriding those initializers and those files can be deleted.
2. This setting is being overridden by the initializer for 5.1 which is currently completely false.

I suspect it is the former, but we should understand which is happening.

app/controllers/welcome_controller.rb
```diff
@@ -10,7 +10,11 @@ class WelcomeController < ApplicationController
   before_action :redirect_if_logged_in
 
   # Skip all filters for status
-  skip_filter *_process_action_callbacks.map(&:filter), only: [:status]
+  # TODO Changed skip_filter to skip_before_action per Rails 5.1 upgrade 
+  # TODO note that the raise: flag was added to avoid an exception. This is 
+  # temporary so that testing can continue. The raise flag should be removed and the exception
+  # should be tracked down.
+  skip_before_action *_process_action_callbacks.map(&:filter), only: [:status], raise: false
```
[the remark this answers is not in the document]
jdf-eos: This was already fixed. I'll remove the comment.

config/routes.rb
```diff
@@ -102,19 +102,19 @@
 
   resources :variable_drafts, controller: 'variable_drafts', draft_type: 'VariableDraft' do
     member do
-      get 'edit', path: 'edit(/:form)'
+      get :edit, path: "edit(/:form)"
```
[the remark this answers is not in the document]
jdf-eos: As I recall,  this was a rubocop change.  I'm hesitant about changing this without another full test cycle.   Talk about this at scrum.

spec/features/collection_drafts/forms/saving_data_centers_data_contacts_spec.rb
```diff
@@ -1,3 +1,6 @@
+# TODO  rails_helper require was added for RAILS5 upgrade to eliminate a race condition
```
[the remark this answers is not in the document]
jdf-eos: removed "TODO"

config/application.rb
jdf-eos: 1) is true.  However,  the initializer files were left around so we can refer to them as needed.  The intent was to leave them in-place until we are completely finished with Rails upgrade. At this point, they should probably be removed after the 6.0 upgrade, or after we decide to postpone the 6.0 upgrade.

config/routes.rb
```diff
@@ -102,19 +102,19 @@
...
     end
   end
 
   resources :service_drafts, controller: 'service_drafts', draft_type: 'ServiceDraft' do
     member do
-      get 'edit', path: 'edit(/:form)'
+      get :edit, path: "edit(/:form)"
```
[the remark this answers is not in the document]
jdf-eos: ditto.

config/routes.rb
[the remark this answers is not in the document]
jdf-eos: Did the test cycle and fixed/commited.

config/routes.rb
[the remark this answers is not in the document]
jdf-eos: ditto

Gemfile
```diff
@@ -42,9 +42,9 @@ gem 'awrence' # convert snake_case hash keys to CamelCase hash keys
 gem 'bootstrap3-datetimepicker-rails'
 gem 'breadcrumbs_on_rails'
 gem 'builder'
-gem 'carmen' # countries and subdivisions
+gem 'carmen', '~>1.0.2'  # countries and subdivisions
 gem 'database_cleaner' # added to provide a solution to Capybara's problems with js=>true
-gem 'factory_girl_rails'
+gem 'factory_girl_rails', :require => true
```
jceaser: was this the gem who's location was disputed? Was that resolved?

Gemfile
```diff
@@ -99,7 +100,7 @@ group :development do
 end
 
 group :test do
-  gem 'capybara'
+  gem 'capybara', '3.19.0'
```
jceaser: we don't want 3.19.1?

app/controllers/bulk_updates_controller.rb
```diff
@@ -54,14 +56,15 @@ def new
 
   def preview
     add_breadcrumb 'Preview', bulk_update_preview_path
-
-    @task = construct_task(params)
+    permitted = params.to_unsafe_h unless params.nil?# need to understand what this is doing more, think related to nested parameters not permitted.
```
jceaser: do we have a TODO syntax for these comments? Would this be relevant here?

Gemfile
ryanmiller-1: I don't think we have had the conversation John wanted to have about it yet.  It is resolved insomuch as it can be built and deployed correctly.

jdf-eos: This isn't perfect, but it works and Ryan is currently working another ticket to replace FactoryGirl with FactoryBot.  So, we are going with this.

config/environments/test.rb
```diff
@@ -43,14 +28,19 @@
...
+
   # Tell Action Mailer not to deliver emails to the real world.
   # The :test delivery method accumulates sent emails in the
   # ActionMailer::Base.deliveries array.
   config.action_mailer.delivery_method = :test
   config.action_mailer.default_url_options = { host: 'localhost', port: '3000' }
 
   # Randomize the order test cases are executed.
-  config.active_support.test_order = :random
+  config.active_support.test_order = :sorted
```
jceaser: to be random or not to be random, is that the question? - why was this changed?

Gemfile
jdf-eos: 3.19.0 is what we have been testing with.   Capybara is up to 3.30, or something like that.   I think that we should wait until we get all of the other Gem upgrades & removals completed then we should test with the latest version capybara.

config/environments/test.rb
jdf-eos: Strictly for easily repeating the same test sequence after fixing a problem.  This could go be set back to random on this point.  Let me know if you want me to make that change and I'll put it in.

app/controllers/bulk_updates_controller.rb
jdf-eos: We were using TODO RAILS5 for awhile, but all of those have been removed.   There are a few RAILS5 comments left around, such as those around the "datetime" changes.

charleshuang80: I'm seeing comments like this in quite a few places (not having gone through everything yet). Some use `to_unsafe_h`, and some use the `safe_hash` method above. I'm wondering if there is a difference. Also, these should be addressed by MMT-2030, right?

config/environments/test.rb
jdf-eos: Thought about this some more and decided to put it back to random.  It is committed.

config/cable.yml
```diff
@@ -0,0 +1,10 @@
+development:
+  adapter: async
+
+test:
+  adapter: async
+
+production:
+  adapter: redis
+  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
```
charleshuang80: does this mean we have Redis available now?

app/controllers/bulk_updates_controller.rb
jdf-eos: This is a difficult one.  safe_hash and to_unsafe_h work differently,  but I suspect that only safe_hash should be used in our code.   Unfortunately,  to_unsafe_h is part of rails so we can't remove it. 

I recommend that we leave these as they are and I will add some notes to MMT-2030 about the issues.  Then we can decide if/when/how to fix them.  

Note: I had thought that MMT-2030 was OBE, since we fixed the permit issues as part of the general bug fixing.

ryanmiller-1: safe_hash is defined in the application controller.  I believe that it takes a key, ensures that the key is there and then mass permits the keys.  to_unsafe_h just makes a hash of the keys.

config/environments/production.rb
```diff
@@ -49,14 +61,15 @@
   config.log_level = :info
 
   # Prepend all log lines with the following tags.
-  # config.log_tags = [ :subdomain, :uuid ]
-
-  # Use a different logger for distributed setups.
-  # config.logger = ActiveSupport::TaggedLogging.new(SyslogLogger.new)
+  config.log_tags = [ :request_id ]
```
charleshuang80: I assume this will change how our logs look in Splunk. I think we should take a note to take a look once this hits production and see if we need to adjust.

config/environments/test.rb
```diff
@@ -1,12 +1,6 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
-  # Precompile assets before running the test suite
```
charleshuang80: just curious - does removing this actually affect anything?

app/controllers/bulk_updates_controller.rb
charleshuang80: I think that plan makes sense. But in my opinion 2030 is not OBE, because I think we need to actually name all the params we want to permit as Rails is trying to force us to do.

config/cable.yml
jdf-eos: Rails 5.2 includes an option for a  Redis cache store.  See https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-rediscachestore

We do not have this turned on.  Our default is still the mem cache.

PR #469, opened 2020-01-13
Hotfix/MMT-2067 Fix sorting by submitter in MMT proper proposal workflow

ryanmiller-1:
Sort by submitter function did not work properly with hashes.  The data that MMT proper sorts is a hash, not active records.  Created a second sort by submitter function that will work on hashes.  Added tests to verify this behavior in MMT proper

### DIFF
--- a/app/controllers/manage_proposal_controller.rb
+++ b/app/controllers/manage_proposal_controller.rb
@@ -166,4 +166,26 @@ class ManageProposalController < ManageMetadataController
     proposal_urs_user = retrieve_urs_users([proposal['submitter_id']])[0]
     proposal_urs_user.blank? ? nil : { name: "#{proposal_urs_user['first_name']} #{proposal_urs_user['last_name']}", email: proposal_urs_user['email_address'] }
   end
+
+  # This function is mirrored in the proposal draft controller to sort an
+  # array of records instead of an array of hashes. They should be kept similar
+  # to maintain consistency in behavior for the proposal workflow.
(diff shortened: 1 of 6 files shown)


app/controllers/manage_proposal_controller.rb
```diff
@@ -166,4 +166,26 @@ def submitter_from_proposal(proposal)
...
+  # to maintain consistency in behavior for the proposal workflow.
+  def sort_hashes_by_submitter(proposals,user_hash = {})
+    @query = {}
+    @query['sort_key'] = params['sort_key'] unless params['sort_key'].blank?
+
+    sorted_proposals = proposals.sort do |a, b|
+                         # Making these arrays allows empty submitter_ids to sort last/first for ASC/DESC
+                         # Using 0 for 'has submitter id' and 1 for 'does not have submitted id'
+                         # allows these to naturally sort last in ASC order, and first in DESC order
+                         a_name = user_hash[a['submitter_id']] ? [0, user_hash[a['submitter_id']]] : [1, user_hash[a['submitter_id']]]
```
cgokey: I wonder if you could reuse the same function, and just have an "if" statement here (pseudocode)
a_submitted_id = if a instance_of hash ? a['submitted_id'] : a.submitted_id
b_submitted_id = if b instance_of hash ? b['submitted_id'] : b.submitted_id

ryanmiller-1: I implemented this.  I'm not fond of instance_of, but I prefer maintaining the logic once.

PR #470, opened 2020-01-15
MMT-1930 Updating UMM-V

ryanmiller-1:


lib/assets/schemas/variables/umm-var-form.json
```diff
@@ -453,9 +469,13 @@
                   "items": [
                     {
                       "type": "section",
-                      "items": [{"key": "/MeasurementSource", "field": true, "type": "UmmSelect"}]
+                      "items": [{"key": "/MeasurementObjectURI", "field": true, "htmlClass": "uri"}]
```
ryanmiller-1: remove htmlClass

lib/assets/schemas/variables/umm-var-search-results-json-schema.json
```diff
@@ -0,0 +1,131 @@
+{
```
ryanmiller-1: do we need any of this?

app/concerns/controlled_keywords.rb
```diff
@@ -162,4 +162,22 @@ def set_country_codes
     united_states = country_codes.delete(Carmen::Country.named('United States'))
     @country_codes = country_codes.unshift(united_states)
   end
+
+  def set_measurement_names
```
ryanmiller-1: add comment to explain structure

ryanmiller-1:
PO feedback was to remove the auto-selection behavior from the non-required fields.  This answers the question populating multiple fields with the same value.

lib/assets/schemas/variables/umm-var-form.json
```diff
@@ -583,5 +603,47 @@
         }
       ]
     }
-  ]
+  ],
+  "definitions": {
```
charleshuang80: what was the reason to create this definitions section and define the two fields here?

ryanmiller-1: At some point while trying to understand how to make the modifications I wanted, separating them worked better.  I have updated the schema to not include the definitions section now.

PR #472, opened 2020-01-17
MMT-1719 Approver E-mails when a Proposal is Submitted

ryanmiller-1:
Updated how I was getting the approvers' URS to send the submission e-mails.  This method would tie the e-mails to whomever has the create permissions for approvers in SCIOPS.

### DIFF
--- a/app/controllers/proposal/collection_draft_proposals_controller.rb
+++ b/app/controllers/proposal/collection_draft_proposals_controller.rb
@@ -233,7 +233,34 @@ module Proposal
     # Get urs users and get emails from urs for approvers
     # Returns array of hashes for each user
     def get_approver_emails
-      approver_urs_users = retrieve_urs_users(ENV['approvers_urs'].split(','))
+      provider_target = Rails.env.development? ? 'MMT_2' : 'SCIOPS'
+
+      # Find the ACL for correct provider in this env.
+      find_acl_response = cmr_client.get_permissions({ 'provider' => provider_target }, token)
+      return [] unless find_acl_response.success?
+
+      approver_acl = find_acl_response.body['items'].select { |acl| acl['name']&.include?('NON_NASA_DRAFT_APPROVER') }
+      return [] if approver_acl.blank?
+
+      # Find the groups that have this ACL in this env.
+      find_groups_response = cmr_client.get_permission(approver_acl[0]['concept_id'], token)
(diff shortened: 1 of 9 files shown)


app/controllers/proposal/collection_draft_proposals_controller.rb
```diff
@@ -226,6 +230,44 @@ def submitter_from_resource
       { name: "#{proposal_urs_user['first_name']} #{proposal_urs_user['last_name']}", email: proposal_urs_user['email_address'] } unless proposal_urs_user.blank?
     end
 
+    # Get urs users and get emails from urs for approvers
```
ryanmiller-1: This function represents the bulk of the changes between this PR and the last one.

app/controllers/proposal/collection_draft_proposals_controller.rb
```diff
@@ -226,6 +230,44 @@ def submitter_from_resource
...
+    # Returns array of hashes for each user
+    def get_approver_emails
```
charleshuang80: This is a nice idea, getting the emails from the ACLs. We should make sure we are getting all the right emails -- in SIT and UAT we check 'SCIOPSTEST' for the Draft MMT ACLs as well.

spec/support/approved_proposals_helpers.rb
```diff
@@ -78,5 +78,20 @@ def mock_urs_get_users(count: 1, users: [])
       urs_response = cmr_success_response(urs_response_body.to_json)
       allow_any_instance_of(Cmr::UrsClient).to receive(:get_urs_users).and_return(urs_response)
     end
+
+    # Fake getting ACLs from CMR, then getting groups from CMR, then getting group members from CMR
+    def mock_get_approver_emails
```
charleshuang80: Mocking most of this is fine for some of the tests, but we should have at least one instance where we are running through as many of the endpoints we can talk to in test (i.e. CMR) as possible.

app/controllers/proposal/collection_draft_proposals_controller.rb
```diff
@@ -226,6 +230,54 @@ def submitter_from_resource
...
+      provider_targets = if Rails.env.production?
```
ryanmiller-1: Wrote MMT-2130 to move these to <environment>.rb files (needs to be done in other locations as well).

PR #475, opened 2020-01-27
MMT-2050-2 Upgrading Factory Girl => Factory Bot

ryanmiller-1:
I do not think that this PR addresses John's unease about factory_bot being outside the 'test and development' bundle.  To resolve that, we would need to rework the rake tasks in drafts.rake and testing.rake.  

For some reason, the initial update to factory bot causes faker to regress.  Updated faker to prevent unnecessary regression.


ryanmiller-1:
Made no changes; force pushed to capture rebase to master since this branch was a little stale

PR #476, opened 2020-01-27
MMT-2031 Removing Database Cleaner

ryanmiller-1:
Rails 5 added the necessary support to remove database cleaner.  Needed to lightly rewrite a pair of template tests to get them to pass due to the change.  Using database cleaner caused the ids to reset for each test (e.g. the first one always had an id of 1), but transactional fixtures still allows the record id to increment (e.g. each id is random based on the order the tests are run).  The tests were updated to be agnostic to this (like the rest of the tests of this nature in our codebase).


ryanmiller-1:
Made no changes; force pushed to capture rebase to master since this branch was a little stale

PR #479, opened 2020-01-29
MMT-1598 unwinding a change in hotfix-2

ryanmiller-1:


ryanmiller-1:
background-image does not respect the no-repeat left and displays the logo in MMT improperly as a result.  The other behaviors appear to have been fixed in the other hotfix.

PR #481, opened 2020-02-03
Feature/mmt 2056 3

jdf-eos:


app/assets/stylesheets/layout/_header.scss
```diff
@@ -6,7 +6,7 @@
 
 header.mmt-header {
   @include row();
-  @include linear-gradient($ocean-blue, darken($ocean-blue, 2%));
+  background-color: linear-gradient($ocean-blue, darken($ocean-blue, 2%));
```
ryanmiller-1: I do not think this is working correctly.  Locally, it looks like the header isn't being colored.

jdf-eos: Good catch! I didn't notice that one.    Its a weird one because the bourbon docs say that the default property is background-color, but what I see in master is background-image.   In any case, it is now working properly.

PR #491, opened 2020-02-10
MMT-2137 Adjusting tests for variable uniqueness changes in CMR

ryanmiller-1:
CMR made a change such that you cannot associate two collections to one variable.  Some of our tests tested for that specifically and needed to be changed as a consequence.

### DIFF
--- a/spec/features/variables/delete_variable_spec.rb
+++ b/spec/features/variables/delete_variable_spec.rb
@@ -5,11 +5,9 @@ describe 'Delete variable', js: true do
     @ingested_variable_with_associations, _concept_response = publish_variable_draft
 
     ingested_collection_1, _concept_response = publish_collection_draft
-    ingested_collection_2, _concept_response = publish_collection_draft
 
     create_variable_collection_association(@ingested_variable_with_associations['concept-id'],
-                                           ingested_collection_1['concept-id'],
-                                           ingested_collection_2['concept-id'])
+                                           ingested_collection_1['concept-id'])
 
     @ingested_variable_without_associations, _concept_response = publish_variable_draft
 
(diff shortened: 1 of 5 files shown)


spec/features/variables/delete_variable_spec.rb
```diff
@@ -41,7 +39,7 @@
 
         it 'informs the user of the number of collection associations that will also be deleted' do
           # 2 associated collections
-          expect(page).to have_content('This variable is associated with 2 collections. Deleting this variable will also delete the collection associations.')
+          expect(page).to have_content('This variable is associated with 1 collections. Deleting this variable will also delete the collection associations.')
```
william-valencia: Not a big deal but 'collections' should be singular 'collection'

spec/features/variables/variable_permissions_spec.rb
```diff
@@ -147,7 +145,7 @@
 
             it 'informs the user of the number of collection associations that will also be deleted' do
               # 2 associations created
-              expect(page).to have_content('This variable is associated with 2 collections. Deleting this variable will also delete the collection associations')
+              expect(page).to have_content('This variable is associated with 1 collections. Deleting this variable will also delete the collection associations')
```
william-valencia: Not a big deal but 'collections' should be singular 'collection'

spec/features/variables/delete_variable_spec.rb
ryanmiller-1: While I agree that the English is wrong, my goal with this ticket is to get master building.  The change in CMR should have a matching ticket in MMT to change our interface and our features, because a fair chunk more than the English is now obsolete.

PR #492, opened 2020-02-13
MMT-2126 Stop displaying/logging tokens returned by CMR in error messages

ryanmiller-1:
Added a helper method to replace the token in an error message with "Token beginning with <first 6 characters>...".
Applied helper method whenever the errors are being checked/set.
Made sure that clean inspect checks/sets the errors before returning a string containing them
Checked where we were using inspect vs clean_inspect and changed to clean_inspect when possible

### DIFF
--- a/lib/cmr/response.rb
+++ b/lib/cmr/response.rb
@@ -41,24 +41,27 @@ module Cmr
     end
 
     def errors
-      @errors ||= if body_is_html?
-                    Rails.logger.error "CMR Error Response Body is a HTML document: #{body}"
-                    ['There was an error with the operation you were trying to perform. There may be an issue with one of the services we depend on. Please contact your provider administrator or the CMR OPS team.']
-                  elsif body.is_a?(Hash)
-                    if body['errors']
-                      Array.wrap(body['errors'])
-                    elsif body['error']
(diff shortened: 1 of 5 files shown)


lib/cmr/response.rb
```diff
@@ -41,24 +41,27 @@ def body_is_html?
...
-                    if body['errors']
-                      Array.wrap(body['errors'])
-                    elsif body['error']
-                      Array.wrap(body['error'])
-                    end
-                  elsif body.is_a?(String)
-                    [body]
-                  else
-                    []
+      return @errors if @errors.present?
```
charleshuang80: what was wrong with the `||=` ?

ryanmiller-1: Nothing was wrong with it, but we can minimize the number of times we search the errors for tokens if we do it only when we assign them here.  
Using the ||= syntax would have meant each part of the if tree would have needed the call to scrub the tokens or the errors would be checked for tokens every time the errors method was called.
Either: 
`@errors ||= if a
                       do thing
                       check for tokens
                     elsif b
                       do thing
                       check for tokens...`
Or: 
`@errors ||= all of this unchanged
check for tokens
@errors` where the check for tokens occurs every time this is called...

PR #493, opened 2020-02-14
MMT-2009 Disabled variable name field when provider_id + native_id is found in CMR.  Updated flash for name uniqueness

ryanmiller-1:
Updated the flash message when a variable fails to ingest because we submitted a record which changed the name (not allowed per CMR). 
Added comments to make finding the conversion of variable/service_drafts to published variables/services more apparent for unknowing developers.  Collection drafts publish; Variable and Service drafts create new variables/services.  I feel that these should have been implemented the same way.
Added logic when making the forms for variables/services that allows for marking some fields as readonly based on the conditions at the time of rendering the forms (previously the schema was static)
Disabled the name field when querying the CMR finds a provider+native id to match the current record.

### DIFF
--- a/app/controllers/base_published_record_controller.rb
+++ b/app/controllers/base_published_record_controller.rb
@@ -63,11 +63,6 @@ class BasePublishedRecordController < ManageMetadataController
       Rails.logger.info("User #{current_user.urs_uid} attempted to ingest #{resource_name} draft #{draft.entry_title} in provider #{current_user.provider_id} but encountered an error.")
 
       @ingest_errors = generate_ingest_errors(ingested_response)
-      # Per CMR's documentation, 422 is the error code when the records fails
-      # non-structural validation e.g. the name of a variable does not match
-      # or a controlled keyword doesn't exist
-      # From observations: 400 is returned when it is structurally invalid
-      # e.g. missing a required field or formatted incorrectly
       flash[:error] = ingested_response.error_message(i18n: I18n.t("controllers.#{plural_resource_name}.create.flash.error"), force_i18n_preface: true)
       redirect_to send("#{resource_name}_draft_path", draft)
     end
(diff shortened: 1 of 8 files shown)


app/controllers/base_published_record_controller.rb
```diff
@@ -60,7 +63,12 @@ def create
       Rails.logger.info("User #{current_user.urs_uid} attempted to ingest #{resource_name} draft #{draft.entry_title} in provider #{current_user.provider_id} but encountered an error.")
 
       @ingest_errors = generate_ingest_errors(ingested_response)
-      redirect_to send("#{resource_name}_draft_path", draft), flash: { error: I18n.t("controllers.#{plural_resource_name}.create.flash.error") }
+      flash[:error] = if ingested_response.body['errors'].any? { |error| error.include?('does not match the existing variable name') }
```
ryanmiller-1: changes this to use .errors

ryanmiller-1: Investigate error code vs error message

spec/features/variable_drafts/create_variable_draft_from_variable_spec.rb
```diff
@@ -1,5 +1,5 @@
 describe 'Create new draft from variable', reset_provider: true do
-  context 'when editing a published variable' do
+  context 'when making a draft of a published variable' do
```
charleshuang80: super minor (and you can ignore if you disagree), but _if_ you end up making other changes, I think 'from' is more appropriate than 'of'

spec/features/variable_drafts/create_variable_draft_from_variable_spec.rb
```diff
@@ -17,4 +17,35 @@
...
+        @variable_draft = create(:full_variable_draft, native_id: @native_id)
+        visit edit_variable_draft_path(@variable_draft)
+      end
+
+      it 'cannot edit the name' do
+        expect(page).to have_field('variable_draft[draft][name]', readonly: true)
+      end
+    end
+
+    context 'when trying to publish/update a variable draft with a different name' do
```
charleshuang80: I feel a little ambivalent about this test. On one hand, it is good to know this is supposed to be blocked and what will happen. On the other hand, this is effectively a test in MMT's code that is testing that CMR is behaving correctly. But, since it is also testing your change to `error_message` then I think we should leave it.

PR #494, opened 2020-02-17
MMT-1831 Fix Element Issues for 508 Compliance

william-valencia:
Removed the 2nd main in the index page and modified all header elements to have a separate ender


app/views/welcome/index.html.erb
```diff
@@ -1,21 +1,19 @@
-<main class="welcome" role="main">
```
ryanmiller-1: It looks like there's CSS tied to this and no other uses of it.  Could you check if we can get rid of it?

PR #495, opened 2020-02-19
MMT-1831 Fix Element Issues Flagged for 508 Compliance

william-valencia:
Removed 2nd main and .welcome css. Also added a separate header ender for each header element.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -23,7 +23,7 @@
   <%= render partial: 'shared/google_tag_manager_body' %>
 
   <div class="th-wrapper">
-    <header id="earthdata-tophat2" />
+    <header id="earthdata-tophat2"></header>
   </div>
 
   <a href="#main-content" class="skip-nav">Skip to main content</a>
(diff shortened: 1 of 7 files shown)


william-valencia:
I recreated the branch since it seems there were issues picking up Ryan's changes for MMT-2137

app/views/layouts/application.html.erb
```diff
@@ -23,7 +23,7 @@
   <%= render partial: 'shared/google_tag_manager_body' %>
 
   <div class="th-wrapper">
-    <header id="earthdata-tophat2" />
+    <header id="earthdata-tophat2"></header>
```
jceaser: good catch, while this may have been fine under the xhtml days and is allowed in html5, I think this syntax is more clear.

app/views/welcome/index.html.erb
```diff
@@ -1,21 +1,19 @@
-<main class="welcome" role="main">
```
jceaser: what was wrong with "main"?

william-valencia: There are 2 of them. There is a <main class="internal" role="main"> just outside this main on the welcome page. I believe this is from application.html.erb.

PR #503, opened 2020-02-24
MMT-2147 Fixed empty warning banner message when an associated open d…

jdf-eos:
Added default warning banner message and updated existing test case to match.

### DIFF
--- a/app/views/services/show.html.erb
+++ b/app/views/services/show.html.erb
@@ -71,6 +71,9 @@
           <%= link_to 'This service has an open draft associated with it. Click here to view it.', @draft %>
         <% elsif available_provider?(@provider_id) %>
           <%= link_to 'This service has an open draft associated with it. Click here to view it.', '#not-current-provider-modal', class: 'display-modal not-current-provider', data: { 'provider': @provider_id, record_action: 'view-draft' } %>
+        <% else %>
+            It appears that you do not have access to an open draft already associated with this service.<br />
+            If you feel you should have access, please check with your provider manager or ensure that you are logged into the correct provider.
         <% end %>
       </p>
     </div>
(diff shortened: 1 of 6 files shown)


app/views/services/show.html.erb
```diff
@@ -71,6 +71,9 @@
           <%= link_to 'This service has an open draft associated with it. Click here to view it.', @draft %>
         <% elsif available_provider?(@provider_id) %>
           <%= link_to 'This service has an open draft associated with it. Click here to view it.', '#not-current-provider-modal', class: 'display-modal not-current-provider', data: { 'provider': @provider_id, record_action: 'view-draft' } %>
+        <% else %>
+            It appears that you do not have access to an open draft already associated with this service.<br />
+            If you feel you should have access, please check with your provider manager or ensure that you are logged into the correct provider.
```
ryanmiller-1: It does not make a lot of sense to me to display 'ensure that you are logged into the correct provider' to the user because they should not see this message if they have access to the provider.  I think it would make more sense to include something about refreshing their available providers.

charleshuang80: This is language that we are using in other places. If we want to change the language I think we should discuss with the PO and push off to another ticket to change it in all the other places that make sense to.

spec/features/services/service_with_draft_spec.rb
```diff
@@ -80,7 +80,7 @@
       end
     end
 
-    context 'when the services provider is not in the users available providers' do
```
charleshuang80: if you don't need to keep the `js: true` for the test to run and pass properly, then it should be removed.

app/views/services/show.html.erb
charleshuang80: However, I think that the first sentence seems a little confusing. I think it would be better worded more like 'This service has an open draft associated with it. However, it appears you do not have access to the open draft.', but I would make sure the PO is ok with it. If this is the language that she wanted then ignore my comment.

jdf-eos: I talked with PO about this and we agreed that Charles' last suggestion is the best option. I have made the suggested change in the latest commit.

spec/features/services/service_with_draft_spec.rb
jdf-eos: I have removed the "js: true" in the latest commit.

PR #506, opened 2020-02-26
MMT-1726: Update UMM-C to version 1.15

charleshuang80:
* Update UMM-C schemas to v1.15
* Change Collection Draft Forms
* update JS for validation errors, required icons, form selectors
* update preview gem
* update factories and tests
* create data migration and add test for it


app/assets/javascripts/required_fields.coffee
```diff
@@ -35,6 +35,8 @@ $(document).ready ->
 
       requiredFieldDataLevels.push topRequiredDataLevel + name + '_'
       requiredFieldDataLevels.push topRequiredDataLevel + name + '_0_'
+    console.log "requiredFieldDataLevels: #{JSON.stringify(requiredFieldDataLevels)}"
```
charleshuang80: comment out or remove

app/assets/javascripts/validation.coffee
```diff
@@ -328,6 +334,74 @@ $(document).ready ->
...
+            if $("##{otherDimId}").val() != ''
+              # if the other dimension has a value, this one is no longer required
+              # you shall not pass
+              error = null
+              return
+        # else
+        #   console.log "what??? another keyword `requred` with `anyOf` error??", error
+      else
+        # you shall not pass
+        console.log "supressing a keyword `required` relating to `anyOf` error", error
```
charleshuang80: comment out

spec/migrations/umm_1_15_migration_spec.rb
```diff
@@ -0,0 +1,701 @@
+# By the final PR, this file should be deleted because the function it depends on
```
charleshuang80: remove comment

app/assets/stylesheets/components/_metadata-preview.scss
```diff
@@ -53,6 +53,16 @@ ul.sub-field {
   }
 }
 
+.two-column-fields-left {
```
ryanmiller-1: I believe these are a remnant from the preview gem changes I made for v1.14.  I don't think you used them, so I think they can be removed.

app/assets/javascripts/draft_form_selectors.coffee
```diff
@@ -1,30 +1,53 @@
 $(document).ready ->
+  ###
```
ryanmiller-1: I am unsure that everything between here and the UMM-V stuff is all UMM-C.  Some of it may be common.  For instance, I suspect that the related URLs stuff is shared.

app/views/collection_drafts/forms/fields/_gridded_range_resolution_fields.html.erb
```diff
@@ -0,0 +1,78 @@
+<div class="row sub-fields">
```
ryanmiller-1: I think that these field sets can be done in partials that are shared between them.  I think you would only need to pass in the parent for the help text.
X/Y Dim
Units
Min/Max X/Y Dim
Viewing Angle/Scanning Direction

ryanmiller-1: We can spin it off as a tech debt ticket, but I suspect it won't get done anytime soon if it isn't done now.

charleshuang80: I will take a look, but I think it's possible the field html attributes (id, data-level, etc.) might not come out correctly

app/assets/stylesheets/components/_metadata-preview.scss
charleshuang80: removed

app/assets/javascripts/draft_form_selectors.coffee
charleshuang80: I removed the label. We can reorganize and add back the label correctly another time.

app/views/collection_drafts/forms/fields/_gridded_range_resolution_fields.html.erb
charleshuang80: I didn't think it would work, but I was able to move them into partials and it seems to be working.

README.md
```diff
@@ -96,7 +95,38 @@ After adding the line and saving the file, don't forget to source the file
 
     source ~/.bash_profile
 
-#### 3. Running the CMR rake tasks
+#### 3. Setting up local redis for CMR
+CMR comes with redis in the jar, but it is not compiled to run on Macs.  If you need to run the CMR on a Mac, download it from
```
cgokey: I don't think I've ever installed Redis for cmr, is needed?

app/assets/javascripts/validation.coffee
```diff
@@ -176,24 +176,21 @@ $(document).ready ->
...
-          "#{field} should have one type completed"
+      when 'oneOf'
+        # the stock message is: "should match exactly one schema in oneOf", which is not as useful to the user
+        "#{field} should have one option completed"
       when 'invalidPicklist' then "#{field} #{error.message}"
       when 'enum' then "#{field} value [#{$('#' + error.id + ' option:selected').text()}] does not match a valid selection option"
-      when 'anyOf' then "#{error.message}"
+      when 'anyOf'
+        # the stock message is: "should match some schema in anyOf", which is not as useful to the user
+        "#{field} should have one schema option completed"
```
cgokey: will the user know what a schema is?

app/assets/javascripts/validation.coffee
```diff
@@ -328,6 +334,74 @@ $(document).ready ->
     error.id = id
     error.element = $("##{id}")
 
+    # Hide individual required errors from an anyOf constraint
```
cgokey: Not saying this should be done now, but I think the below code should be refactored, the if/then/else nested structure is very deep makes it difficult to follow.

README.md
charleshuang80: I think that there are certain things that might not run properly if you don't install it. I forget what they are, Ryan probably knows more.

app/assets/javascripts/validation.coffee
charleshuang80: This is the message the PO approved. We have a ticket next sprint to improve the validation messages presented to the user.

app/assets/javascripts/validation.coffee
charleshuang80: I have tried to add comments to explain what decisions are being made and why. I thought about abstracting some of this into helper methods, but the need to return out of the function in some cases made it hard to figure out how to do that. If you have any suggestions I'm open to them. We can try to address them now or in the improvement ticket I mentioned.

README.md
ryanmiller-1: I did not compile a list of things that will not run properly without redis, but we recently had a discussion where Charles, John, and William could all reliably reproduce a series of test failures and I could not (I run redis with my uberjar now).  When they added redis, the test failures stopped.  I believe that full functionality of the uberjar is now dependent on it (circa December-ish).

spec/features/collection_drafts/forms/radio_button_selectors_spec.rb
```diff
@@ -141,33 +93,38 @@
...
-        expect(page).to have_field('Geographic Coordinate Units', with: '')
-        expect(page).to have_field('Latitude Resolution', with: '')
-        expect(page).to have_field('Longitude Resolution', with: '')
+        expect(page).to have_no_checked_field('Gridded Resolutions')
+
+        check 'Gridded Resolutions'
+
+        expect(page).to have_field('Unit', with: '')
+        expect(page).to have_field('X Dimension', with: '')
+        expect(page).to have_field('Y Dimension', with: '')
```
ryanmiller-1: Do we want a test for the data clearing with the checkboxes being toggled on and off?

charleshuang80: yes, good catch

README.md
cgokey: Ah that might be why some of my local builds have failed in the past.   Thanks!

spec/features/collection_drafts/forms/radio_button_selectors_spec.rb
charleshuang80: added it in the test for checkboxes

README.md
jceaser: this is one off the drivers for Docker

app/assets/javascripts/validation.coffee
jceaser: was using unless tried for readability on the cases where you just pass with ';'?

db/schema.rb
```diff
@@ -10,82 +10,82 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20200107163307) do
+ActiveRecord::Schema.define(version: 2020_02_06_201605) do
```
jceaser: why the format change, I like the new format, but why?

ryanmiller-1: We don't create the strings; they're a part of db:migrate.

app/assets/javascripts/validation.coffee
charleshuang80: It is a pretty nested set of if and else logic, and I think unless would make it much more complicated to understand.
Maybe I didn't explain it as well in the comment above, but I would have liked to abstract this logic into helper methods, but the need to return out of the function would necessitate adding more funkiness.
Also, because this is coffeescript, the `;` gets transpiled into `break;` to break out of the if/else logic. If it would be clearer, I can write it as `break;` instead.

db/schema.rb
jceaser: latter in the day I saw the same format in dashboard as well, so I retract my question.

PR #508, opened 2020-03-02
MMT-2151  Remove authentication token from SOAP response.   Create te…

jdf-eos:
Removed authentication token from SOAP response.  Create test case to verify that the token is removed.


spec/features/filters/echo_token_filter_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  
+  let(:echo_provider) { Echo::Provider.new('http://example.com', Rails.configuration.services['echo_soap']['services']) }
+  
+  
+  context 'When token is returned in SOAP Response' do
+    before do
+      allow_any_instance_of(Faraday::Connection).to receive(:post).and_return(Faraday::Response.new(body: token_response, status: 500))
+    end
+
+    it 'filters the token' do
```
ryanmiller-1: Should this succeed?  In this instance, shouldn't the response contain `<ns4:ErrorCode>NAMSAuidNotFoundFor**Token**</ns4:ErrorCode>` ?

jdf-eos: The code only removes tags named "Token" inside the AuthorizationFault section of the response.   The ErrorCode's value is not removed.  I could test that the full structure of the SOAP request is undisturbed, but that isn't really the point of the test.   

In fact,  the full test is to look at the log file and see what was written when this test completes.  That is what we will have to do to test it in production.

ryanmiller-1: My interpretation of line 17 is that it is supposed to be checking the string conversion of an echo response for the text 'Token'.  I do not have an intuitive sense of what the string conversion of an echo response is, but I would would expect it to contain most/all of the fields in the response.  Based on that assumption, the fact that this test passes struck me as odd, because I would expect at least one of the 'Token's in the response to match it.  
I threw a byebug in the test and discovered that `echo_provider.test_endpoint_connection("https://test.com", :non_token_response).to_s` evaluates to `"#<Echo::Response:0x00007fa4f4c5b4a8>"`.

jdf-eos: Note that the code in lines 40-44 of base.rb only affects the log message, not the SOAP response.  I used the test case to generate the log message to confirm that the logging did not include the <ns4:Token> tag.

I probably shouldn't have included the test case in my commit and I will remove it.

ryanmiller-1: Ah, I had overlooked that.

The alternative is that you could test the log message.  There are examples of that in spec/cmr/cmr_client_spec.rb

lib/echo/base.rb
```diff
@@ -37,10 +37,15 @@ def make_request(url, body)
...
       begin
-
-        Rails.logger.error "SOAP Response Error: #{echo_response.body.inspect}" if echo_response.error?
-
+        msg_hash = Hash.send('from_xml', echo_response.body)
+        if echo_response.error?
+          token_val = msg_hash.dig('Envelope', 'Body', 'Fault', 'detail','AuthorizationFault','Token')
+          msg_hash.dig('Envelope', 'Body', 'Fault', 'detail','AuthorizationFault').delete('Token') if token_val.present?
+        end
+        Rails.logger.error "SOAP Response Error: #{msg_hash.inspect}"
```
william-valencia: Seems like this line will print regardless if there is an error or not. Is that what you really want here? "SOAP Resonse Error:..." will always be printed.

jdf-eos: Good catch. I'll fix this.

PR #510, opened 2020-03-04
Feature/MMT-2160 Updating to jquery3

ryanmiller-1:
From Chris's work in MMT-2139, fixed existing datepicker, dropzone, and permissions issues that arose.

### DIFF
--- a/app/assets/javascripts/datepicker.coffee
+++ b/app/assets/javascripts/datepicker.coffee
@@ -1,3 +1,10 @@
+################################################################################
+# date pickers are used throughout our draft forms (e.g. temporal information) #
+# if you are looking for the UI objects controlling the selection of dates in  #
+# the CMR search related fields (e.g. bulk updates), you should look in the    #
+# datetimepicker files (e.g. datetimepicker.coffee)                            #
+################################################################################
+
 $(document).ready ->
   $('.metadata-form').on 'focus', 'input[type="datetime"]', ->
     pickerOpts =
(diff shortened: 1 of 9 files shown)


app/assets/stylesheets/components/_date-picker.scss
```diff
@@ -13,7 +13,7 @@
   direction: ltr;
 
   > div {
-    display: none;
+    display: inline-block;
```
ryanmiller-1: Ad comment to explain necessity

app/assets/javascripts/permissions.coffee
```diff
@@ -196,11 +196,11 @@ $(document).ready ->
         'search_groups[]':
           required: (element) ->
             # field is required if the other field has no value/selection
-            $('#search_and_order_groups_').val() == null
+            $('#search_and_order_groups_').val().length == 0 || $('#search_and_order_groups_').val() == null
```
jceaser: swap tests, null first

cgokey: yeah agree.

app/assets/javascripts/datepicker.coffee
```diff
@@ -1,3 +1,10 @@
+################################################################################
+# date pickers are used throughout our draft forms (e.g. temporal information) #
+# if you are looking for the UI objects controlling the selection of dates in  #
+# the CMR search related fields (e.g. bulk updates), you should look in the    #
+# datetimepicker files (e.g. datetimepicker.coffee)                            #
+################################################################################
+
```
cgokey: you sure?  search I though was Datepicker, forms (for selecting dates) was Datetimepicker?

ryanmiller-1: Yes.
My wild, not based on any real data hypothesis is that at some point we went to make the bulk updates and determined that we wanted a UI that offered maximum specificity for their query (maybe because it was a requirement) and discovered that our datepickers did not do the job.

PR #512, opened 2020-03-07
Update README.md

RubenArias1993:
Check documentetion


jceaser:
This word is spelled differently according to my spell checker.

PR #513, opened 2020-03-09
MMT-2178 Updating default from address to prevent gmail filtering

ryanmiller-1:
Updating MMT domain so that GMail should not filter it.


config/environments/development.rb
```diff
@@ -35,6 +35,7 @@
   config.action_mailer.default_url_options = { host: 'localhost', port: '3000' }
   config.action_mailer.perform_deliveries = false
   config.action_mailer.perform_caching = false
+  config.default_email_domain = 'earthdata.nasa.gov'
```
jceaser: you have a comment for the other environments, but not here

PR #514, opened 2020-03-09
MMT-2135 updated style sheet to fix div alignment problem.

jdf-eos:


jdf-eos:
@ryanmiller-1  What do you suggest?   Close this and open a new issue for the preview gem?

ryanmiller-1:
The preview gem just copies the precompiled version of MMT's assets.  There are instructions for it in its repo.  I think you'll need a PR here and over there.  Over here, you'll need to include the new preview gem commit ref.

PR #516, opened 2020-03-11
MMT-2175: fix SOAP calls to Legacy Services

charleshuang80:
* remove token_with_client_id method
* modify tokens used in Echo SOAP methods
* updates in Echo::Client libraries
* update tests
* move around some controller and test files


lib/echo/base.rb
```diff
@@ -65,6 +70,17 @@ def wrap_with_envelope(content)
       end
     end
 
+    def payload_token(token)
+      return nil if token.blank? # better to return '' or nil ???
```
charleshuang80: remove comment

PR #518, opened 2020-03-16
MMT-2179 Updating datetime inputs to be custom-datetimes and updating comment to make it clear that we are not relying on a deprecated html standard.

ryanmiller-1:


app/helpers/form_helper.rb
```diff
@@ -143,15 +143,17 @@ def mmt_select(options)
   end
 
   # RAILS5.1 -- datetime_field_tag returns a datetime_local instead of date_time
-  # datetime_local doesn't support UTC.  
-  ## def mmt_datetime_field_tag(name, dt, options)
+  # datetime_local doesn't support UTC and is supported in some, but not all
+  # browsers.  custom-datetimes are managed by datepicker.coffee and
+  # datepicker.scss and are unrelated to the deprecated html datetime input type
```
charleshuang80: In datapicker.scss (and in date-picker.scss) I see no reference to anything with 'datetime' as a selector. I'm not sure if that means this comment is misleading or not.

ryanmiller-1: The datepicker.coffee picks up the input tag and then all of the css to control the datepickers is in those scss files.  I can reword the comment

PR #523, opened 2020-03-20
MMT-2179-2 updating to use custom-datetime so that it is not confused with deprecated html input field

ryanmiller-1:
Lost a battle with git on the old branch, so I copied the changes into a new branch.  Same changes as other PR (https://github.com/nasa/mmt/pull/518/commits/85851e07072c8ec4a13f411c2089305a09109ed2), but with the comment and gems updated.


app/helpers/form_helper.rb
```diff
@@ -168,7 +171,7 @@ def mmt_datetime(options)
     )
     
     # TODO -- gsub reverts to the pre-Rails 5.1 behavior of using datetime,  instead of datetime-local See MMT-2060
```
charleshuang80: do we still need this comment?

PR #524, opened 2020-03-20
MMT-2175: fix for missed echo provider token passed to soap call

charleshuang80:


lib/echo/data_management.rb
```diff
@@ -168,11 +168,11 @@ def get_order_options(echo_provider_token, guids = nil)
     end
 
     # Deprecates an order so it can be deleted
-    def deprecate_order_options(token, guids)
+    def deprecate_order_options(echo_provider_token, guids)
       builder = Builder::XmlMarkup.new
 
       builder.ns2(:SetCatalogItemOptionDefinitionsDeprecated, 'xmlns:ns2': 'http://echo.nasa.gov/echo/v10', 'xmlns:ns3': 'http://echo.nasa.gov/echo/v10/types', 'xmlns:ns4': 'http://echo.nasa.gov/ingest/v10') do
-        builder.ns2(:token, payload_token(token))
```
charleshuang80: This was the cause of the bug that Francell found. We were passing the client id with the echo provider token, which caused problems.

PR #525, opened 2020-03-24
MMT-2115 adding modal partial to page.

ryanmiller-1:


spec/features/proposal/search/collection_search_spec.rb
```diff
@@ -64,8 +80,6 @@
       end
 
       it 'does not display the actions for collections in MMT proper' do
-        # TODO when we have the actions for collections added in Draft MMT
```
ryanmiller-1: These are being tested in create_from_collection_spec.rb, so they don't need to be added to another test here.  Nothing lost be removing this comment.

PR #526, opened 2020-03-24
Mmt 2054

jdf-eos:
Removed Neat Gem which is no longer supported. Provided a simpler replacement package to perform the essential calculations for the CSS grid.

### DIFF
--- a/app/assets/stylesheets/components/_callout-box.scss
+++ b/app/assets/stylesheets/components/_callout-box.scss
@@ -3,13 +3,21 @@
 \* ===================================================== */
 
 .eui-callout-box {
-  @include span-columns(4);
+  float: left;
+  display: block;
+  margin-right: neat-gutter-width();
+  width: neat-span-width(4);
   box-sizing: border-box;
   background-color: $white;
   margin-left: 0;
   margin-right: 5%;
   max-width: 294px; // Max width to handle rows with less than 3 callout boxes
   min-height: 400px;
+  
+}
+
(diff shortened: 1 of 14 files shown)


app/assets/stylesheets/layout/_grid.scss
```diff
@@ -21,27 +26,115 @@ main.dashboard {
...
-.col-1 { @include span-columns(1); }
-.col-2 { @include span-columns(2); }
-.col-3 { @include span-columns(3); }
-.col-4 { @include span-columns(4); }
-.col-5 { @include span-columns(5); }
-.col-6 { @include span-columns(6); }
-.col-7 { @include span-columns(7); }
-.col-8 { @include span-columns(8); }
-.col-9 { @include span-columns(9); }
+.col-1 {
```
ryanmiller-1: Can these be condensed?  Aside from the width, they're the same.

app/assets/stylesheets/components/_callout-box.scss
```diff
@@ -85,18 +96,31 @@
...
+  
+  .info:last_child {
+    margin-right: 0
   }
 
   .comments {
-    @include span-columns(2);
+    float: left;
+    display: block;
+    margin-right: neat-gutter-width();
```
charleshuang80: I am going back and forth on this a little bit. I think that if 12 is the default number of columns, we should be specific and use 12, just so it is clear in the future. However, it also looks like for this particular mixin, we are never passing a fewer number of columns.

PR #527, opened 2020-03-25
MMT-2091 turning on create endpoint for subscriptions

ryanmiller-1:


jceaser:
I get errors when I run the test file you updated

app/controllers/subscriptions_controller.rb
```diff
@@ -52,7 +50,7 @@ def create
   private
 
   def subscription_params
-    params.require(:subscription).permit(:description, :metadata, :user_id, :email_address) # :subscriber_email)
+    params.require(:subscription).permit(:name, :concept_id, :metadata, :user_id, :email_address) # :subscriber_email)
```
charleshuang80: I think we can remove the subscriber email from this for now

spec/features/manage_cmr/subscriptions/create_subscriptions_spec.rb
```diff
@@ -28,19 +28,21 @@
         end
 
         it 'displays validation errors within the form' do
-          expect(page).to have_content('Description is required.')
+          expect(page).to have_content('Name is required.')
```
charleshuang80: This should be 'Subscription Name', right?

spec/features/manage_cmr/subscriptions/create_subscriptions_spec.rb
```diff
@@ -72,25 +72,32 @@
...
-        context 'when submitting an subscription that fails' do
+        context 'when submitting a subscription that fails' do
+          # Generating a genuine failure by violating uniqueness constraints
+          # in the CMR.
+          let(:name2) { 'Exciting Subscription with Important Data4' }
           before do
-            error_response_body = '{"errors":["some error message"]}'
-            error_response = cmr_fail_response(JSON.parse(error_response_body))
-            allow_any_instance_of(Cmr::CmrClient).to receive(:create_subscription).and_return(error_response)
+            @ingest_response, _concept_response = cmr_client.ingest_subscription({ 'Name' => name2, 'Query' => query, 'CollectionConceptId' => collection_concept_id, 'SubscriberId' => 'fake', 'EmailAddress' => 'fake@fake.fake'}.to_json, 'MMT_2', 'test_native_id', 'token')
```
charleshuang80: We don't need to do this now as there aren't many, but we will probably want to create a helper method to ingest these, like we do for the metadata records

spec/features/manage_cmr/subscriptions/create_subscriptions_spec.rb
```diff
@@ -132,7 +139,7 @@
       end
 
       it 'does not display the new subscription form' do
-        expect(page).to have_no_field('Description')
+        expect(page).to have_no_field('Name')
```
charleshuang80: also should be Subscription name, right?

app/views/subscriptions/new.html.erb
```diff
@@ -5,10 +5,19 @@
...
+          <%= text_area_tag('subscription[name]', @subscription['name'], cols: 30, rows: 3, class: 'required textcounter', maxlength: 255) %>
+        </div>
+      </fieldset>
+      
+      <fieldset>
+        <div class="row">
+          <%= label_tag('subscription[concept_id]', 'Collection Concept ID', class: 'eui-required-o') %>
+        </div>
+        <div class="row">
+          <%= text_field_tag('subscription[concept_id]', @subscription['concept_id'], class: 'required full-width', placeholder: 'Enter a ConceptID for a single collection. (Example: C1234567-PODAAC)') %>
```
charleshuang80: I'm not a fan of the placeholder text (UI/UX team doesn't like it for design reasons), but I asked PO about it and we can make tweaks as we go along. So we can leave it, just wanted to register my thoughts.

spec/features/manage_cmr/subscriptions/create_subscriptions_spec.rb
ryanmiller-1: Yeah, I was going to make a few minor changes in this test file to align Name with 'Subscription Name' when I had other feedback.

ryanmiller-1:
Re: Errors when running the tests
I think your CMR uberjar is old (the endpoint this code uses is new).  If you update your uberjar and still get them, let me know.

spec/features/manage_cmr/subscriptions/create_subscriptions_spec.rb
ryanmiller-1: Adding to the list of nice to haves: rake command to set permissions

PR #529, opened 2020-03-25
MMT-2115 permitting download_xml in collections controller

ryanmiller-1:
The links were redirecting.


spec/features/proposal/search/collection_search_spec.rb
```diff
@@ -62,6 +62,23 @@
           expect(page).to have_link('DIF 9', href: download_collection_xml_path(@ingest_response['concept-id'], 'dif'))
           expect(page).to have_link('DIF 10', href: download_collection_xml_path(@ingest_response['concept-id'], 'dif10'))
         end
+
+        context 'when trying to download data' do
```
charleshuang80: interesting that it worked this way, but I like that we are able to actually test it somehow.

PR #532, opened 2020-03-30
MMT-2114: Only proposal owners and approvers can access proposals

charleshuang80:
* update model associations
* update calls for proposal lists
* update policies to reflect change
* remove redundant proposal mode check in policies
* add policy
* add policy exception messages
* add and update tests


app/controllers/proposal/manage_collection_proposals_controller.rb
```diff
@@ -1,11 +1,12 @@
 module Proposal
+
   class ManageCollectionProposalsController < ManageMetadataController
     skip_before_action :provider_set?
 
     before_action :ensure_non_nasa_draft_permissions
 
     def show
-      authorize CollectionDraftProposal, policy_class: CollectionDraftProposalPolicy
+      authorize CollectionDraftProposal, policy_class: ManageCollectionProposalPolicy
```
ryanmiller-1: does :manage_collection_proposal or something like how subscriptions does this work here?

spec/features/proposal/collection_draft_proposals/collection_draft_proposal_list_spec.rb
```diff
@@ -189,12 +189,12 @@
 
   context 'when logged in as an approver' do
     before do
-      login
+      real_login(method: 'urs')
       mock_urs_get_users
```
ryanmiller-1: This was removed above.  Can it be removed here?

app/controllers/proposal/manage_collection_proposals_controller.rb
charleshuang80: good call, made the change

spec/features/proposal/collection_draft_proposals/collection_draft_proposal_list_spec.rb
charleshuang80: I actually didn't remove it above, it just got moved down to line 11. I tried removing them and the tests failed so we need to keep the calls.

PR #533, opened 2020-03-31
MMT-2197 - New Rake Task

jceaser:
adding a rake task to download the lattest CMR jar from the maven repository

### DIFF
--- a/lib/tasks/local_cmr.rake
+++ b/lib/tasks/local_cmr.rake
@@ -6,21 +6,35 @@ namespace :cmr do
   jar_name = 'cmr-dev-system-0.1.0-SNAPSHOT-standalone.jar'
   url_source = "https://maven.earthdata.nasa.gov/repository/mmt/#{jar_name}"
 
-  desc 'Fetch lattest jar'
+  desc 'Fetch lattest CMR jar from the maven repository'
   task :fetch do
-    # make a backup copy the old jar if it exists 
+    # make a backup copy of the old jar if it exists, and a backup of the backup 
     cmd_backup = 'cd cmr ; '\
       "if [ -a \"#{jar_name}\" ] ; then "\
         'echo Backup jar... ; '\
+        "mv -fv #{jar_name}.last #{jar_name}.oldest ; "\
         "mv -fv #{jar_name} #{jar_name}.last ; "\
       'fi'
     puts %x( #{cmd_backup} )
     
-    # download a new jar
+    # download a new jar to a temp location and rename it if successfull
     puts 'Download jar...'
     cmd_fetch = 'cd cmr ; '\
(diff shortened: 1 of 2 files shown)


lib/tasks/local_cmr.rake
```diff
@@ -3,6 +3,41 @@ require 'test_cmr/load_data.rb'
...
+    # restore if error, as defined by there not being a jar
+    cmd_restore = 'cd cmr ; '\
+      "if [ -a \"#{jar_name}\" ] ; then "\
+        "rm #{jar_name}.oldest ; "\
+      'else '\
+        "mv -f #{jar_name}.last #{jar_name} ;"\
+        "mv -f #{jar_name}.oldest #{jar_name}.last ;"\
+        "rm #{jar_name}.tmp ;"\
+      'fi'
+    puts %x( #{cmd_restore} )
```
charleshuang80: I don't know what these outputs look like, are they sufficiently informative?

lib/tasks/local_cmr.rake
```diff
@@ -3,6 +3,41 @@ require 'test_cmr/load_data.rb'
 namespace :cmr do
   include Cmr
 
+  jar_name = 'cmr-dev-system-0.1.0-SNAPSHOT-standalone.jar'
+  url_source = "https://maven.earthdata.nasa.gov/repository/mmt/#{jar_name}"
+
+  desc 'Fetch lattest CMR jar from the maven repository'
```
charleshuang80: wondering if we want to add anything to the readme for this task. not quite certain either way.

ryanmiller-1: I think it should be documented somewhere that we would point a new developer to.  Also, latest is misspelled in the description of the task

jceaser: I can add a section in the readme

lib/tasks/local_cmr.rake
jceaser: I'll put a comment in on why I'm using puts %() and not Process.spawn()

lib/tasks/local_cmr.rake
```diff
@@ -3,6 +3,45 @@ require 'test_cmr/load_data.rb'
 namespace :cmr do
   include Cmr
 
+  jar_name = 'cmr-dev-system-0.1.0-SNAPSHOT-standalone.jar'
+  url_source = "https://maven.earthdata.nasa.gov/repository/mmt/#{jar_name}"
+
+  desc 'Fetch latest CMR jar from the maven repository'
+  task :fetch do
+    # use puts %x() to run commands and not Process.spawn(), which returns
+    # imediatly, because curl can take a while and we want to see the output
```
charleshuang80, line 12: typo on immediately (not important though).

PR #534, opened 2020-04-01
MMT-2093 Update Subscriptions

ryanmiller-1:


lib/tasks/acls.rake
```diff
@@ -71,6 +71,29 @@ namespace :acls do
     end
   end
 
+  namespace :subscriptions do
+    desc 'Creates a group and grants permission for managing subscriptions'
+    task :full_acls, [:username] => :environment do |_task, args|
+      # Create the group
+      Rake::Task['acls:groups:create'].invoke('Subscription Management', 'Group Created for Subscription Management', args.username, 'MMT_2')
```
charleshuang80, line 78: if we need to be able to ingest subscriptions with this, you might need to also add `testuser` to the group, as the token we pass in development and test is for that user

spec/features/manage_cmr/subscriptions/edit_subscriptions_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+
+  context 'when visiting the edit page' do
+    before do
+      # go to show page
+      VCR.use_cassette('urs/rarxd5taqea', record: :none) do
+        visit edit_subscription_path(@ingest_response['concept_id'], subscription: @subscription, native_id: @native_id)
+      end
+    end
+
+    it 'has the right fields' do
```
charleshuang80, line 38: this seems like a duplicate of what's being tested on line 21, not sure it is necessary

spec/features/manage_cmr/subscriptions/edit_subscriptions_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      end
+    end
+
+    context 'when making an invalid modification to a subscription' do
+      before do
+        @new_name = 'A Different Query'
+        fill_in 'Subscription Name', with: @new_name
+        @second_native_id = 'test_edit_id_2'
+
+        @ingest_response, @subscription = publish_new_subscription(name: @new_name, native_id: @second_native_id)
```
charleshuang80, line 78: might want to add a comment as to what you're doing here, as it takes a bit to fully grasp

spec/features/manage_cmr/subscriptions/edit_subscriptions_spec.rb
```diff
@@ -0,0 +1,93 @@
+describe 'Edit/Updating Subscriptions' do
+  before do
+    login
+    allow_any_instance_of(SubscriptionPolicy).to receive(:create?).and_return(true)
+    allow_any_instance_of(SubscriptionPolicy).to receive(:show?).and_return(true)
+    allow_any_instance_of(SubscriptionPolicy).to receive(:destroy?).and_return(true)
```
charleshuang80, line 6: do we actually need destroy here?

PR #538, opened 2020-04-06
MMT-2093 updating name length for subscriptions

ryanmiller-1:


app/views/subscriptions/forms/_subscription_fields.html.erb
```diff
@@ -3,7 +3,7 @@
     <%= label_tag('subscription[Name]', 'Subscription Name', class: 'eui-required-o') %>
   </div>
   <div class="row">
-    <%= text_area_tag('subscription[Name]', @subscription['Name'], cols: 30, rows: 3, class: 'required textcounter', maxlength: 255) %>
+    <%= text_field_tag('subscription[Name]', @subscription['Name'], cols: 30, rows: 3, class: 'required full-width textcounter', maxlength: 80) %>
```
jceaser, line 6: is there not a test for this? 3 humans have now had to deal with it.

ryanmiller-1: The best I can do now is update the test to use a name that is 80 characters long.

jceaser: well, then lets just think about this then, and you go forward as is.

PR #539, opened 2020-04-06
MMT-2187 Validating paired fields

ryanmiller-1:


app/controllers/collection_drafts_controller.rb
```diff
@@ -254,6 +260,68 @@ def validate_parameter_ranges(errors, metadata)
     errors
   end
 
+  def validate_paired_fields(errors, metadata)
+    errors = validate_parameter_ranges(errors, metadata)
+    errors = validate_project_paired_dates(errors, metadata)
+    errors = validate_temporal_paired_dates(errors, metadata)
+    validate_tiling_identification_systems_paired_fields(errors, metadata)
```
charleshuang80, line 267: even though it is probably unnecessary, it seems better to make this line match the others by starting with `errors = `

charleshuang80: if you didn't have to fix anything else I would say ignore this comment for now. but looks like you have merge conflicts to fix also.

ryanmiller-1: John merged his preview gem changes before mine, so it's probably that

PR #540, opened 2020-04-08
MMT-2193 added ability to clean_inspect for echo soap responses and modified logs to use it.

cgokey:


ryanmiller-1:
I would appreciate tests for this.  See spec/cmr/cmr_client_spec.rb for an example on how we do this for the CMR client.

charleshuang80:
I like the changes, but I agree with Ryan, we should add tests

PR #541, opened 2020-04-08
MMT-2077 Add env info to proposal notification email

htranho:


app/views/proposal_mailer/proposal_submitted_approvers_notification.html.erb
```diff
@@ -1,12 +1,12 @@
-<h1 class="styled-heading"><%= "#{@short_name}_#{@version} Submitted" %></h1>
+<h1 class="styled-heading"><%= "#{@short_name}_#{@version} Submitted in #{Rails.configuration.cmr_env.upcase}" %></h1>
```
ryanmiller-1, line 1: Have you checking with Alicia on this? I was sort of expecting there to be no change to the e-mails sent from prod

htranho: Checked, now updated

app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,11 @@
 class ApplicationMailer < ActionMailer::Base
   default from: "no-reply@#{Rails.configuration.default_email_domain}"
   layout 'mailer'
+
+  def add_env_if_needed(email_subject)
+    env = Rails.configuration.cmr_env.upcase
+    if env != 'OPS'
+      email_subject = email_subject + " (#{env})"
```
charleshuang80, line 8: 1. I'm not sure why we this is checking against the `cmr_env` instead of what MMT is running.
2. I don't think the assignment on line 8 is necessary, and I think the preferred syntax is to put it all in quotes and use interpolation
So I think lines 6 - 8 can be replaced with
```
unless Rails.env.production?
  "#{email_subject} (#{Rails.env})"
```

app/views/draft_mailer/draft_published_notification.html.erb
```diff
@@ -9,9 +9,16 @@
   <p>
     <%= @user[:name] %>,
   </p>
+<% env = Rails.configuration.cmr_env.upcase %>
+<% if env != 'OPS'%>
+  <p>
+    <%= "Your collection metadata record #{@short_name}_#{@version} has been successfully #{ @is_update ? 'updated' : "published to the CMR (#{env})"}. Your collection's concept ID is #{@concept_id}." %>
```
charleshuang80, line 15: same comments about `cmr_env` vs the `Rails.env`
also, i think you can do this without the if/else blocks for the content. I would do it like:
```
<% add_env_if_needed = Rails.env.production? nil : " (#{Rails.env})" %>
<p>
  <%= "same string message as before but add #{add_env_if_needed} rest of the message" %>
</p>
```

app/views/draft_mailer/draft_published_notification.text.erb
```diff
@@ -1,5 +1,8 @@
 <%= @user[:name] %>,
-
+<% env = Rails.configuration.cmr_env.upcase %>
```
charleshuang80, line 2: I think you can use the change I suggested above and not need the if/else for the main content. But if it is necessary, I think it would be better to indent what is after the if and else.

app/views/draft_mailer/draft_published_notification.html.erb
charleshuang80: I think this change can be applied to many of the files below.

app/mailers/application_mailer.rb
htranho: Because we don't want any change for production env

jceaser: is your comment about the "unless" code (question 2) or question 1?

charleshuang80:
We have tests in for emails and email content, right? If so, we probably need to update them to have '(test)' in them

app/views/draft_mailer/draft_published_notification.html.erb
jceaser: I see, because the content is mostly the same, only somewhat different?

app/mailers/application_mailer.rb
htranho: It was about 2, but it is clear now, nil will not hurt

htranho:
> We have tests in for emails and email content, right? If so, we probably need to update them to have '(test)' in them

(test) added

ryanmiller-1:
https://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-callbacks
I think that you can compute and set env_if_needed once as a before action.  Never tried it with mailers, but this says it works.

htranho:
@ryanmiller-1 : Great, it works! Code change is now shorter. Thanks.

app/views/draft_mailer/draft_published_notification.html.erb
charleshuang80: yes and if it's `nil` it won't add anything

app/mailers/draft_mailer.rb
```diff
@@ -13,6 +13,8 @@ def draft_published_notification(user, concept_id, revision_id, short_name, vers
       email_subject = 'New Record Published in Metadata Management Tool'
```
htranho, line 13: Add to 11 and 13

app/views/draft_mailer/draft_published_notification.html.erb
```diff
@@ -10,7 +10,7 @@
     <%= @user[:name] %>,
   </p>
   <p>
-    <%= "Your collection metadata record #{@short_name}_#{@version} has been successfully #{ @is_update ? 'updated' : 'published to the CMR'}. Your collection's concept ID is #{@concept_id}." %>
+    <%= "Your collection metadata record #{@short_name}_#{@version} has been successfully #{ @is_update ? 'updated' : "published to the CMR#{@email_env_note}"}. Your collection's concept ID is #{@concept_id}." %>
```
htranho, line 13: <%= "Your collection metadata record #{@short_name}_#{@version} has been successfully #{ @is_update ? 'updated in' : 'published to'} the CMR#{@email_env_note}. Your collection's concept ID is #{@concept_id}." %>

PR #542, opened 2020-04-09
MMT-2198 Reading Subscriptions from CMR

ryanmiller-1:
Updated Index controller and view actions to read from CMR
Updated Show/Edit/Update/Destroy to only require a (concept_)id to be passed
Created a before action for Show/Edit/Update/Destroy to find the rest of the information they need from CMR
Updated Edit/Delete buttons where necessary
Updated get_subscriptions
Updated tests - note: the after block deletes could be removed if we could correctly use reset_provider, but it does not work correctly because there is a bug in CMR (ticket written).

### DIFF
--- a/app/concerns/cmr_subscriptions.rb
+++ b/app/concerns/cmr_subscriptions.rb
@@ -1,0 +1,26 @@
+module CMRSubscriptions
+  extend ActiveSupport::Concern
+
+  def get_latest_revision(concept_id, revision_id)
+    attempts = 0
+    while attempts < 20
+      revisions_response = cmr_client.get_subscriptions({ concept_id: concept_id, all_revisions: true }, token)
+
+      revisions = if revisions_response.success?
+                    revisions_response.body.fetch('items', [])
+                  else
+                    []
+                  end
+      revisions.select! { |revision| revision['meta']['deleted'] == false }
+      revisions.sort! { |a, b| b['meta']['revision-id'] <=> a['meta']['revision-id'] }
+
+      latest = revisions.first
+
+      return latest if latest.present? && latest['meta']['revision-id'].to_s >= revision_id.to_s
+
+      attempts += 1
+      sleep 0.05
+    end
+    nil
+  end
+end
--- a/app/concerns/cmr_subscriptions.rb
+++ b/app/concerns/cmr_subscriptions.rb
@@ -1,6 +1,9 @@
 module CMRSubscriptions
   extend ActiveSupport::Concern
 
+  # This is derived from a similar method found in the cmr_collections_helper
+  # Subscriptions face a similar challenge where without something like this,
+  # MMT can try to load the page without finding the user's new content.
   def get_latest_revision(concept_id, revision_id)
     attempts = 0
     while attempts < 20
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -1,5 +1,6 @@
 class SubscriptionsController < ManageCmrController
   include UrsUserEndpoints
+  include CMRSubscriptions
 
   RESULTS_PER_PAGE = 25
   before_action :subscriptions_enabled?
@@ -55,11 +56,7 @@ class SubscriptionsController < ManageCmrController
     subscription_response = cmr_client.ingest_subscription(@subscription.to_json, current_user.provider_id, native_id, token)
     if subscription_response.success?
       flash[:success] = I18n.t('controllers.subscriptions.create.flash.success')
-
-      # Hitting CMR too quickly in dev environment.
-      sleep(1) if Rails.env.development? || Rails.env.test?
-
-      redirect_to subscription_path(subscription_response.parsed_body['result']['concept_id'])
+      redirect_to subscription_path(subscription_response.parsed_body['result']['concept_id'], revision_id: subscription_response.parsed_body['result']['revision_id'])
     else
       Rails.logger.error("Creating a subscription failed with CMR Response: #{subscription_response.clean_inspect}")
       flash[:error] = subscription_response.error_message(i18n: I18n.t('controllers.subscriptions.create.flash.error'))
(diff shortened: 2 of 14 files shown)


app/controllers/subscriptions_controller.rb
```diff
@@ -68,36 +71,42 @@ def create
...
-    native_id = params.permit(:native_id)['native_id']
 
-    subscription_response = cmr_client.ingest_subscription(@subscription.to_json, current_user.provider_id, native_id, token)
+    subscription_response = cmr_client.ingest_subscription(@subscription.to_json, current_user.provider_id, @native_id, token)
     if subscription_response.success?
       flash[:success] = I18n.t('controllers.subscriptions.update.flash.success')
-      redirect_to subscription_path(subscription_response.parsed_body['result']['concept_id'], subscription: @subscription, native_id: native_id)
+
+      # Hitting CMR too quickly in dev environment.
+      sleep(1) if Rails.env.development? || Rails.env.test?
```
jceaser, line 83: how did this come up? Why did we need to do this?

lib/cmr/cmr_client.rb
```diff
@@ -549,21 +549,13 @@ def check_user_permissions(options, token)
...
-      # so they will change
-      success_response_body =  '{"items":[{"description":"It is a subscription", "user_id":"fake_user_id", "email_address":"fake@fake.fake", "collection_concept_id":"C1200000000-FAKE", "metadata":"thing=stuff&&stuff_id=stuff&&stuff_color=more_stuff"},{"description":"It is a second subscription", "user_id":"fake_user_id", "email_address":"fake2@fake.fake", "collection_concept_id":"C1200000001-FAKE", "metadata":"thing=stuff&&stuff_id=stuff2&&stuff_color=more_stuff2"}]}'
-      Cmr::Response.new(Faraday::Response.new(status: 200, body: JSON.parse(success_response_body)))
-
-      # This error response is provided for viewing and testing a failure
-      # error_response_body = '{"errors":["some error message"]}'
-      # Cmr::Response.new(Faraday::Response.new(status: 400, body: JSON.parse(error_response_body)))
+      url = if Rails.env.development? || Rails.env.test?
+              'http://localhost:3003/subscriptions.umm_json'
+            else
```
jceaser, line 555: Why the different formats?

ryanmiller-1: Error on my part

app/controllers/subscriptions_controller.rb
```diff
@@ -135,4 +144,19 @@ def get_subscriber_email(subscriber_id)
   def add_top_level_breadcrumbs
     add_breadcrumb 'Subscriptions', subscriptions_path
   end
+
+  def fetch_subscription
+    @concept_id = params.permit(:id)['id']
+    options = { 'ConceptId' => @concept_id }
```
ryanmiller-1, line 150: condense

app/controllers/subscriptions_controller.rb
jceaser: I see from the demo why you are doing this, but I think we should find some other solution if possible. If no other solution exists then this idea is not horrible.

spec/features/manage_cmr/subscriptions/create_subscriptions_spec.rb
```diff
@@ -68,6 +68,12 @@
             end
           end
 
+          after do
```
ryanmiller-1, line 71: add comment for all these afters that they should be superfluous when we can reset_provider

lib/cmr/cmr_client.rb
ryanmiller-1: addressed

app/controllers/subscriptions_controller.rb
ryanmiller-1: addressed, will comment to explain below

ryanmiller-1:
I removed the waits in development/test by adding a 'get_latest_revision' function.  It will try 20 times to get the revision specified or later.  This mimics what we do for publishing collections.  If it fails to find it, it will log and display an error warning (parallel to the one we show with collections).

app/controllers/subscriptions_controller.rb
```diff
@@ -135,4 +133,22 @@ def get_subscriber_email(subscriber_id)
...
     add_breadcrumb 'Subscriptions', subscriptions_path
   end
+
+  def fetch_subscription
+    permitted_params = params.permit(:id, :revision_id)
+    @concept_id = permitted_params['id']
+
+    subscription = get_latest_revision(@concept_id, permitted_params['revision_id'] || 0)
+
+    if subscription
```
charleshuang80, line 143: i would suggest using `fetch` in here as well

app/views/subscriptions/index.html.erb
```diff
@@ -10,10 +10,15 @@
       <p><%= page_entries_info @subscriptions, entry_name: 'Subscription' %></p>
     <% end %>
 
+    <p>
+      <em><span class="fa fa-info-circle"></span> Subscription operations may take some time. If you are not seeing what you expect below, please <%= link_to 'refresh the page', 'javascript:history.go(0)', class: 'eui-btn--link link-lift' %>.</em>
```
charleshuang80, line 14: if we are doing the 'try 20 times', do we still need this?

app/views/subscriptions/index.html.erb
```diff
@@ -27,26 +32,33 @@
...
             <tr>
-              <td>
-                <%= subscription['description'] %>
+              <td class='wrapping-td'>
+                <div class='wrapping-div'>
+                  <%= link_to subscription['Name'], subscription_path(subscription['ConceptId']) %>
+                </div>
               </td>
+              <td><%= subscription['CollectionConceptId'] %></td>
               <!-- TODO: need to put a limit or char cutoff for query? -->
```
charleshuang80, line 41: still need comment?

spec/features/manage_cmr/subscriptions/index_subscriptions_spec.rb
```diff
@@ -17,54 +30,73 @@
...
+    context 'when viewing the index page' do
+      before do
+        visit subscriptions_path
+      end
+
+      # These tests can be improved when we can reset_provider. They should have
+      # only 2 subscriptions on them, but if the subscription tests are run
+      # together, CMR does not always finish deleting subscriptions from other
+      # tests before starting this one, so it fails at the commented out line.
+      it 'displays sample subscriptions' do
```
charleshuang80, line 44: 1. I don't know if 'sample' is the right descriptor
2. I think you want to be more explicit here that it is showing the subscriptions but not the links for which they don't have access

app/concerns/cmr_subscriptions.rb
```diff
@@ -0,0 +1,26 @@
+module CMRSubscriptions
+  extend ActiveSupport::Concern
+
+  def get_latest_revision(concept_id, revision_id)
+    attempts = 0
```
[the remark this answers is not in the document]
ryanmiller-1: done

app/controllers/subscriptions_controller.rb
ryanmiller-1: done

app/views/subscriptions/index.html.erb
ryanmiller-1: Spoke with PO, she wanted a new message in its place.

PR #544, opened 2020-04-10
MMT-2117: validate subscription query

charleshuang80:


app/assets/javascripts/subscriptions.coffee
```diff
@@ -27,3 +74,23 @@ $(document).ready ->
           element.closest('.subscriber-group').append(error)
         else
           error.insertAfter(element)
+
+
+    # method for validating query
+    $.validator.addMethod 'isValidQuery', (value, elem, params) ->
```
charleshuang80, line 80: add comment about event and clicking submit

spec/features/manage_cmr/subscriptions/subscription_form_validation_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+          expect(page).to have_content('Query is required.')
+          expect(page).to have_content('Subscriber is required.')
+        end
+      end
+
+      context 'when submitting a subscription query with an invalid structure' do
+        let(:invalid_query_structure) { 'thisisabadquery123456$#*(@notlookingforanything' }
+
+        before do
+          fill_in 'Query', with: invalid_query_structure
```
charleshuang80, line 49: comment about order

app/assets/javascripts/subscriptions.coffee
ryanmiller-1: The validation triggers the same way for the required fields on SIT.  If you click on the text box, enter nothing, then try to submit it will only validate the one you were in.  I hate that behavior, but I suspect it is hard to fix.

charleshuang80: comments added

charleshuang80:
@jceaser you wanted me to ask Chris (D) and Leo if there were any unlisted search parameters we would need to support. They both said no, that we only need to support what's in the docs.

PR #548, opened 2020-04-15
MMT-2117: add missed query params, allow spaces in values, add test with comment

charleshuang80:


spec/features/manage_cmr/subscriptions/subscription_form_validation_spec.rb
```diff
@@ -113,6 +113,32 @@
...
+          fill_in 'Query', with: valid_query
+          fill_in 'Name', with: name
+          fill_in 'Collection Concept ID', with: collection_concept_id
+
+          within '.subscription-form' do
+            click_on 'Submit'
+          end
+        end
+
+        it 'displays the proper validation errors withing the form' do
```
jdf-eos, line 134: within

charleshuang80: hah, that typo is literally copy/pasted through the whole file. If I have to re-run the build, I will make the change, otherwise I will leave it for another time.

PR #549, opened 2020-04-15
MMT-2195 Perform validation before publish and submit

ryanmiller-1:
Renamed show_view_setup to collection_validation_setup because it needs to be run in all of the places we run validate_metadata.
Added verify_valid_metadata as a before action for publish action to repeat the validation that was run when the show page was loaded.
Added before actions for the submit action (proposals) to load the schema, run collection_validation_setup, and run the validation


ryanmiller-1:
Force push to capture rebase to master

PR #550, opened 2020-04-16
MMT-2133 Removing dif 9 from available formats for download

ryanmiller-1:


ryanmiller-1:
Screen Shot 2020-04-16 at 11 19 08 AM

PR #551, opened 2020-04-17
MMT-2188 updating ingest errors to display parent field when the field with the error is a multiple item. Prevented string operations on integer.

ryanmiller-1:
We were trying to display the field with an error, but in this case the field CMR returned to us was the entire G-polygon.  So we were trying to perform string operations on the number 0.  Displaying simply 0 as a link wasn't informative to the user, so when the 'field' is a number, I updated ingest_errors to return the field as `"<parent field> <number>"`.  I also changed the indexing from 0 index to 1 index (like the forms).

Removed an earlier version of the subscriptions controller that wasn't deleted. I suspect it is a remnant from a merge conflict.


ryanmiller-1:
Screen Shot 2020-04-17 at 11 57 41 AM

PR #554, opened 2020-04-27
MMT-2154: allow temporal filters for collection permissions

charleshuang80:
* add temporal filter fields to collection permission form
* hook up datetime picker and radio buttons
* add validation of fields
* add or modify methods used for validation
* update helpers for show page explaining filters
* update tests to include new fields and validation

### DIFF
--- a/app/assets/javascripts/datetimepicker.coffee
+++ b/app/assets/javascripts/datetimepicker.coffee
@@ -1,8 +1,9 @@
 ################################################################################
-# date time pickers are used throughout our CMR search related forms (e.g. bulk#
-# updates. If you are looking for the UI objects controlling the selection of  #
-# dates in the draft forms, you should be look in the datepicker files (e.g.   #
-# datepicker.coffee or vendor/assets/bootrap-datepicker-1.6.0                  #
+# date time pickers are used throughout our CMR search related forms (e.g.     #
+# bulk updates) and some Manage Cmr forms. If you are looking for the UI       #
+# objects controlling the selection of dates in the draft forms, you should be #
+# look in the datepicker files (e.g. datepicker.coffee or                      #
+# vendor/assets/bootrap-datepicker-1.6.0)                                      #
 ################################################################################
 
 $(document).ready ->
--- a/app/assets/javascripts/datetimepicker.coffee
+++ b/app/assets/javascripts/datetimepicker.coffee
@@ -1,7 +1,7 @@
 ################################################################################
 # date time pickers are used throughout our CMR search related forms (e.g.     #
 # bulk updates) and some Manage Cmr forms. If you are looking for the UI       #
-# objects controlling the selection of dates in the draft forms, you should be #
+# objects controlling the selection of dates in the draft forms, you should    #
 # look in the datepicker files (e.g. datepicker.coffee or                      #
 # vendor/assets/bootrap-datepicker-1.6.0)                                      #
 ################################################################################
--- a/app/assets/javascripts/permissions.coffee
+++ b/app/assets/javascripts/permissions.coffee
@@ -1,12 +1,14 @@
 # Handle the enabling and disabling of the granule
-# access constraints form fields
+# access constraints and temporal filter form fields
 handleGranuleConstraintState = (granulesEnabled) ->
-  $('#granule-access-constraints-container :input').attr('disabled', !granulesEnabled)
+  $('#granule-access-constraints-container :input').prop('disabled', !granulesEnabled)
+  $('#granule-temporal-filter-container :input').prop('disabled', !granulesEnabled)
   $('#granule-access-constraints-container').toggleClass('disabled', !granulesEnabled)
+  $('#granule-temporal-filter-container').toggleClass('disabled', !granulesEnabled)
 
 # Handle the hiding and showing of the collection chooser
 handleCollectionOptions = (selectedCollections) ->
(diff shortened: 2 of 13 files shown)


app/controllers/manage_cmr/permissions_controller.rb
```diff
@@ -148,12 +146,28 @@ def destroy
 
   private
 
+  def collection_permission_params
+    # because there are nested params, when they go through strong parameters
+    # the nested params become a ActionController Parameters object
+    # so when grabbing the nested params we need to call to_h on them or some
+    # hash methods will cause an error
```
charleshuang80, line 153: do the `to_h` here instead of needing to call it in all the other places.

charleshuang80: @ryanmiller-1 , I could not get this to work. Even if I returned a param with the nested params turned into a hash out of this method, somehow the nested hashes we expected to grab out of it still turned into a parameters object. I can show you if you are interested, but I don't think we can make the fix.

app/assets/javascripts/permissions.coffee
```diff
@@ -242,7 +311,13 @@ $(document).ready ->
 
       return true if value == '' || minimum == ''
       parseFloat(value) >= parseFloat(minimum)
-    , 'Maximum value must be greater than or equal to Minimum value.' # default message
+
+    $.validator.addMethod 'startBeforeStop', (value, elem, arg) ->
```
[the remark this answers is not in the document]
charleshuang80: do you mean like startDate? the parameter is supposed to be the value in the field, so I am inclined to keep it, but am open to hearing more about why it should be changed and to what exactly.

app/helpers/permissions_helper.rb
```diff
@@ -25,6 +25,43 @@ def display_access_constraints(access_value)
...
+                 'contained within'
+               when 'intersect'
+                 'having some overlap with'
+               when 'disjoint'
+                 'outside of'
+               end
+
+    filters << "the date range #{temporal['start_date']} to #{temporal['stop_date']}"
+
+    filters.join(' ')
```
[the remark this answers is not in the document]
charleshuang80: normally I would as well, but I just mimicked how the method above it was constructed, I don't think it makes sense to have them operate differently.

app/controllers/manage_cmr/permissions_controller.rb
jdf-eos: Rails really doesn't want you to do this.  They did provide a 'to_unsafe_h' which still doesn't do exactly what you want.      We have a ticket to avoid these problems by having a method that creates a hash from params using a passed in structure of the keys that you want to permit.  It is probably time to fix that ticket so all the controllers work the same way.

ryanmiller-1: The suggestion was to get the permitted parameters as a hash because the parameter class that they are does not implement all the hash methods (I assume Charles can provide a specific example as he found it necessary to .to_h in some places). I don't think that is counter to the spirit of the strong params...

app/controllers/manage_cmr/permissions_controller.rb
```diff
@@ -148,12 +146,28 @@ def destroy
 
   private
 
+  def collection_permission_params
```
ryanmiller-1, line 149: Do we understand how the permit system works? If it is rerun, I assume it is recomputing.  In which case, I think an ivar would save compute in construct_request_object, for instance.

app/controllers/manage_cmr/permissions_controller.rb
charleshuang80: @jdf-eos is right, there is some Rails magic happening that won't allow us to make nested params into a straight hash. Tried doing that in the strong params method here, as well as assigning to an ivar in the compute method. In both cases, the nested params after manipulation still came out as a Parameters object. So basically, we can't avoid having to call `to_h` in the places we are grabbing the nested params.

app/assets/javascripts/datetimepicker.coffee
```diff
@@ -1,8 +1,9 @@
 ################################################################################
-# date time pickers are used throughout our CMR search related forms (e.g. bulk#
-# updates. If you are looking for the UI objects controlling the selection of  #
-# dates in the draft forms, you should be look in the datepicker files (e.g.   #
-# datepicker.coffee or vendor/assets/bootrap-datepicker-1.6.0                  #
+# date time pickers are used throughout our CMR search related forms (e.g.     #
+# bulk updates) and some Manage Cmr forms. If you are looking for the UI       #
+# objects controlling the selection of dates in the draft forms, you should be #
+# look in the datepicker files (e.g. datepicker.coffee or                      #
```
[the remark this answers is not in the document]
charleshuang80: I think this is fixed, let me know if there's something else

app/assets/javascripts/required_fields.coffee
```diff
@@ -70,6 +70,10 @@ $(document).ready ->
     if $fields.length > 0
       # get any values from the fields at the current data-level
       fieldsWithValues = $fields.filter ->
+        # TODO: I wanted to change this filtering to use jQuery to be more
```
charleshuang80, line 73: @ryanmiller-1 , @jdf-eos , just wanted to make you aware that I had to revert the change here and added comments, in case that affects your approval. There was bad behavior/a failing test I was not able to resolve quickly and didn't want to hold up the ticket.

PR #558, opened 2020-05-05
hotfix/MMT-2093-2 disabling collection concept id and subscriber id from changing during update.

ryanmiller-1:


app/views/subscriptions/_subscription_fields.html.erb
```diff
@@ -13,7 +13,11 @@
   </div>
   <div class="row">
     <p class="form-description">Enter a Concept ID for a single collection. (Example: C1234567-PODAAC)</p>
-    <%= text_field_tag('subscription[CollectionConceptId]', @subscription['CollectionConceptId'], class: 'required full-width') %>
+    <% if edit %>
```
charleshuang80, line 16: according to the [text_field_tag](https://apidock.com/rails/v5.2.3/ActionView/Helpers/FormTagHelper/text_field_tag) and [select_tag](https://apidock.com/rails/ActionView/Helpers/FormTagHelper/select_tag) docs, you can use true or false for the disabled option. So I think you can set the disabled fields without the if/else blocks.
So you could have `class: "required full-width #{'disabled' if edit}", disabled: edit`.
if you're going to use it this way, maybe `edit` should be a bit more descriptive, i.e. `edit_form` maybe. On the groups form we used `new_form`, that's a possibility too.

app/controllers/subscriptions_controller.rb
```diff
@@ -70,8 +70,9 @@ def create
   def update
     authorize :subscription
     # Overwrite the old subscription with the new values
-    @subscription = subscription_params
-    @subscription['EmailAddress'] = get_subscriber_email(@subscription['SubscriberId'])
+    subscription_from_form = subscription_params
```
charleshuang80, line 73: maybe write a comment about which fields can be edited

app/views/subscriptions/_subscription_fields.html.erb
```diff
@@ -13,7 +13,11 @@
...
+      <%= text_field_tag('subscription[CollectionConceptId]', @subscription['CollectionConceptId'], class: 'required full-width disabled', disabled: 'disabled') %>
+    <% else %>
+      <%= text_field_tag('subscription[CollectionConceptId]', @subscription['CollectionConceptId'], class: 'required full-width') %>
+    <% end %>
   </div>
```
charleshuang80, line 21: I'm also wondering if we should have a comment as we do on the group form about certain fields not being editable

app/views/subscriptions/edit.html.erb
```diff
@@ -2,8 +2,8 @@
   <section>
     <h2>Edit <span class="provider-context"><%= current_user.provider_id %></span> Subscription</h2>
 
-    <%= form_tag(subscription_path(@concept_id), method: :put, class: 'subscription-form') do %>
-      <%= render partial: 'subscription_fields' %>
+    <%= form_tag(subscription_path(@concept_id), method: :put, class: 'subscription-form edit-subscription-form') do %>
+      <%= render partial: 'subscription_fields', locals: { edit: true } %>
```
charleshuang80, line 6: see my other comment about the fields, maybe `edit` local var should be more descriptive

app/assets/stylesheets/components/_metadata-fields.scss
```diff
@@ -100,3 +100,13 @@
     width: 70%;
   }
 }
+
+.edit-subscription-form {
+  
+  .disabled {
```
charleshuang80, line 106: just want to make sure you verified this is consistent with where we may have other disabled fields

app/views/subscriptions/_subscription_fields.html.erb
jceaser: is there any other forms documentation, such as test that checks for this? If so, then we can just move on, otherwise I am more in favor of documentation than not.

ryanmiller-1: the tests were updated to expect the fields to be disabled.

ryanmiller-1: and it was documented in the controller

PR #559, opened 2020-05-07
MMT-2219: add ability to create a new tools draft record

charleshuang80:
* add new controllers and only necessary new actions and methods
* add model
* add necessary views
* add T schema
* create T form schema for first form
* update helpers and validation only as needed
* add factory
* add tests


app/controllers/manage_tools_controller.rb
```diff
@@ -0,0 +1,13 @@
+# :nodoc:
+class ManageToolsController < ManageMetadataController
+  before_action :umm_t_enabled?
+
+  def show
+    # If you change this number you must also change it in the corresponding test file - features/manage_tools/open_drafts_spec.rb.
+    @draft_display_max_count = 5
+
+    @drafts = ToolDraft.where(provider_id: current_user.provider_id)
```
ryanmiller-1, line 9: Not important to make a change on it's own, but if other things are changed: is the alignment here correct in the file? It doesn't seem to be in the web diff

config/environments/uat.rb
```diff
@@ -111,6 +111,9 @@
   # Feature toggle for subscriptions in MMT
   config.subscriptions_enabled = true
 
+  # Feature toggle for UMM-T
```
ryanmiller-1, line 114: I think we should talk with PO to determine when we want to flip this.  I'd like to get a ticket number and put it in a comment here (not necessarily in this ticket, but an action for a future PR)

spec/features/header/header_spec.rb
```diff
@@ -1,21 +1,24 @@
...
-    Rails.configuration.umm_s_version = "vnd.nasa.cmr.umm+json; version=2.23"
-    @origin_var_config = Rails.configuration.umm_var_version
-    Rails.configuration.umm_var_version = "vnd.nasa.cmr.umm+json; version=1.90"
+    @original_c_config = Rails.configuration.umm_c_version
+    Rails.configuration.umm_c_version = "vnd.nasa.cmr.umm+json; version=2.34"
+    @original_var_config = Rails.configuration.umm_var_version
+    Rails.configuration.umm_var_version = "vnd.nasa.cmr.umm+json; version=3.45"
+    @original_s_config = Rails.configuration.umm_s_version
+    Rails.configuration.umm_s_version = "vnd.nasa.cmr.umm+json; version=4.56"
+    @original_s_config = Rails.configuration.umm_t_version
```
ryanmiller-1, line 9: Overwriting `@original_s_config`

app/controllers/tool_drafts_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    # Allow for completely empty forms to be saved
+    return {} unless params.key?(:tool_draft)
+
+    # If the form isn't empty, only permit whitelisted attributes
+    permitted = params.require(:tool_draft).permit(:draft_type).tap do |whitelisted|
+      # Allows for any nested key within the draft hash
+      whitelisted[:draft] = params[:tool_draft][:draft]
+    end
+
+    permitted.to_unsafe_h # need to understand what this is doing more, think related to nested parameters not permitted.
```
ryanmiller-1, line 41: Does rails balk if this is just a to_h? These are the params we have already explicitly permitted, after all.

charleshuang80: I'll give it a try. But if it doesn't work, I plan on looking into the params a bit in the next ticket.

charleshuang80: actually I'm just going to write a comment to look at it in the next ticket. I don't think I can fully test if it will work in this one as we don't have all the forms.

PR #560, opened 2020-05-08
MMT-2096 E-mail notification estimation

ryanmiller-1:
Added environment variable to control iteration time in CMR
Added button to edit/new pages.
Added modal and javascript to manipulate modal
Added an extra validation trigger for the urs id because it wasn't validating when the user added a valid id.
Added controller action for test subscription
Added helper methods to controller for query processing and email estimation
Added granule ingest to cmr_client with only a local cmr endpoint
Added rake task to bulk ingest the same granule multiple times 
Added helper method for tests to bulk ingest the same granule multiple times
Altered collection ingest helper method for tests such that it can suppress search errors when ingesting in providers where it cannot search for a collection due to permissions restrictions


app/assets/javascripts/subscriptions.coffee
```diff
@@ -100,3 +100,27 @@ $(document).ready ->
         return false if validGranuleQueryParameters.indexOf(parameter) == -1
 
       true
+
+  $('#subscriber').on 'change', ->
+    if $('#subscriber-error').length > 0
+      $('#subscriber').valid()
+
+  $('.test-subscription').on 'click', ->
+    $('.status-text').text('Creating an estimate from available data...')
```
ryanmiller-1, line 109: May need spinner for this behavior too

app/assets/javascripts/subscriptions.coffee
```diff
@@ -100,3 +100,27 @@ $(document).ready ->
         return false if validGranuleQueryParameters.indexOf(parameter) == -1
 
       true
+
+  $('#subscriber').on 'change', ->
+    if $('#subscriber-error').length > 0
+      $('#subscriber').valid()
```
ryanmiller-1, line 106: Take a look at collection permissions for alternates because it's a select2; search groups/search and order

app/controllers/subscriptions_controller.rb
```diff
@@ -167,4 +184,29 @@ def user_not_authorized(exception)
...
+    # restrictive, but if two sets of revision dates are passed, both are
+    # permitted.
+    query.slice!(0) if query[0] == '?'
+    terms = query.split('&')
+    terms.select! do |term|
+      subterms = term.split('=')
+      !subterms[0].include?('updated_since') && !subterms[0].include?('revision_date')
+    end
+
+    prepend = "collection_concept_id=#{concept_id}&sort_key[]=revision_date&updated_since=#{DateTime.parse(30.days.ago.to_s).strftime('%FT%TZ')}"
```
ryanmiller-1, line 200: remove sort_by

app/controllers/subscriptions_controller.rb
```diff
@@ -167,4 +184,29 @@ def user_not_authorized(exception)
...
+    "#{prepend}&#{terms.join('&')}"
+  end
+
+  def emails_per_day(granules)
```
ryanmiller-1, line 204: Update to round to whole numbers

lib/cmr/cmr_client.rb
```diff
@@ -74,6 +74,16 @@ def search_collections(options, token)
       get(url, options, token_header(token))
     end
 
+    def test_query(query, token, headers = {})
```
ryanmiller-1, line 77: This needs to be a post

spec/features/manage_cmr/subscriptions/test_query_spec.rb
```diff
@@ -0,0 +1,256 @@
...
+          all('.select2-container .select2-selection').first.click
+          # page.find('.select2-selection--single').click
+        end
+        page.find('.select2-search__field').native.send_keys('rarxd5taqea')
+
+        page.find('ul#select2-subscriber-results li.select2-results__option--highlighted').click
+      end
+    end
+
+    context 'when the query starts with a "?"' do
```
ryanmiller-1, line 44: give this granules

lib/cmr/cmr_client.rb
ryanmiller-1: Turns out the CMR search API accepts up to ~500k characters, but advises the post for anything over 6k.

ryanmiller-1: After discussion with Siwei, I am inclined to leave this as a get until CMR changes their processing to a post, at which point we should change to the post.

jceaser: meaning they don't accept a POST on this interface?

ryanmiller-1: They accept a POST, but when CMR processes queries to actually send e-mails it currently uses the get end point. I think our estimate should match the endpoint CMR is using internally.

config/environments/uat.rb
```diff
@@ -121,4 +121,6 @@
 
   # Google Tag Manager ID for EOSDIS usage stats
   config.tag_manager_id = 'GTM-WNP7MLF'
+
+  config.cmr_email_frequency = ENV['cmr_email_frequency'].to_i
```
jceaser, line 125: this variable needs to be documented in the OPS manual. Point it out to rosy please.

config/environments/development.rb
```diff
@@ -101,4 +101,5 @@
 
   config.tophat_url = 'https://cdn.sit.earthdata.nasa.gov/tophat2/tophat2.js'
 
+  config.cmr_email_frequency = ENV['cmr_email_frequency'].to_i
```
jceaser: this value should default to something reasonable and not require a value to be set. This same comment is for all environments.

In general all configuration should default to something reasonable if possible and allow for OPS to override when needed. In this way configuration files can hold only the required deviations from base making it easy to understand and read or contain settings that must be their like passwords. In general I don't want our config files to be anything like an apache config file

config/environments/development.rb
```diff
@@ -101,4 +101,9 @@
 
   config.tophat_url = 'https://cdn.sit.earthdata.nasa.gov/tophat2/tophat2.js'
 
+  config.cmr_email_frequency = if ENV['cmr_email_frequency']
```
charleshuang80, line 104: I think you can us the safe nav operator here: `ENV['cmr_email_frequency']&.to_i || 3600`

lib/cmr/cmr_client.rb
charleshuang80: I think you should add a comment with your conclusion

lib/cmr/cmr_client.rb
```diff
@@ -306,6 +316,19 @@ def delete_subscription(provider_id, native_id, token)
       delete(url, {}, nil, headers.merge(token_header(token)))
     end
 
+    # MMT does not need to ingest granules in any environment that is not dev or
+    # test.
+    def ingest_granule(metadata, provider_id, native_id)
+      url = "http://localhost:3002/providers/#{provider_id}/granules/#{encode_if_needed(native_id)}"
```
charleshuang80, line 322: I think it might be better to add a guard clause here: `return unless Rails.env.development? || Rails.env.test?`

spec/features/manage_cmr/subscriptions/test_query_spec.rb
```diff
@@ -0,0 +1,306 @@
...
+  context 'when the form is complete' do
+    before do
+      fill_in 'Name', with: 'valid_test_name'
+      fill_in 'Query', with: 'day_night_flag=day&updated_since=2020-05-06T16:23:09Z&production_date[]=2000-01-01T10:00:00Z,2021-03-10T12:00:00Z'
+      fill_in 'Collection Concept ID', with: @ingest_response['concept-id']
+
+      VCR.use_cassette('urs/search/rarxd5taqea', record: :none) do
+        within '.subscriber-group' do
+          all('.select2-container .select2-selection').first.click
+          # page.find('.select2-selection--single').click
```
charleshuang80, line 39: need to keep?

spec/features/manage_cmr/subscriptions/test_query_spec.rb
```diff
@@ -0,0 +1,306 @@
...
+        allow_any_instance_of(Cmr::CmrClient).to receive(:check_user_permissions).and_return(cmr_response)
+        click_on 'Estimate Notification Frequency'
+      end
+
+      it 'displays an error message on the modal' do
+        expect(page).to have_content("An error occurred while checking the user's permissions: One of parameters [user_type] or [user_id] are required.")
+      end
+    end
+
+    context 'when the test succeeds' do
```
charleshuang80, line 231: You had mentioned that the estimation tests for update were somewhat duplicative. I agree that you should have some tests, but I don't think you need all of the duplicative ones. Happy to discuss this further but I think you should keep:
- 'when the form is incomplete'
- _one_ of the tests under 'when the test succeeds'
I don't think you need to duplicate the others for the update form estimation.

spec/support/subscriptions_helpers.rb
```diff
@@ -29,5 +29,16 @@ def prepare_subscription_permissions(permissions, provider = 'MMT_2')
 
       [group_response['concept_id'], permission_response['concept_id']]
     end
+
+    def ingest_granules(collection_entry_title, count, provider = 'MMT_2')
```
charleshuang80, line 33: all the other ingest helper methods use keyword arguments, you might want to make this one consistent.
Since all the other helpers for ingest are in the drafts_helpers and now you have a separate one for granules for subscriptions, maybe we can consolidate all the ingest related helper methods to an ingest_helpers file?

app/controllers/subscriptions_controller.rb
```diff
@@ -104,6 +105,30 @@ def destroy
...
+
+    granule_search_response = cmr_client.test_query(prepare_query_for_test(subscription['CollectionConceptId'], subscription['Query']), token)
+    render plain: "An error occurred while searching for granules: #{granule_search_response.error_message}", status: :internal_server_error and return unless granule_search_response.success?
+    granules_found = granule_search_response.body['hits']
+    frequency = (Rails.configuration.cmr_email_frequency / 3600.0).ceil(2)
+    frequency = frequency.to_i if frequency.to_i == frequency.to_f
+    return_json = {
+      'granules' => granules_found,
+      # in hours
+      'frequency' => "#{frequency} #{frequency > 1 ? 'hours' : 'hour'}",
```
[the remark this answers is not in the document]
ryanmiller-1: pluralize causes x < 1 to be plural, e.g. 0.5 hours.

PR #562, opened 2020-05-11
MMT-2216 Turned on Rails 5  CSRF improvements (per-form-csrf-tokens & forgery-…

jdf-eos:
Turned on Rails 5  CSRF improvements (per-form-csrf-tokens & forgery-protection-with-origin-check).

Removed old framework initializers and replaced with a single config param, config.load_defaults 5.2


config/application.rb
```diff
@@ -19,7 +19,7 @@
 module Mmt
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 5.1
+    config.load_defaults 5.2
```
jceaser, line 22: that looks like something we missed, should we request a full regression test because of this?

jdf-eos: Yes, I think so.

jceaser: https://apidock.com/rails/v5.2.3/Rails/Application/Configuration/load_defaults not that many changes. 5 maybe.

ryanmiller-1, line 22: I believe that updating this value is equivalent to flipping all of the values in related initializers to true. So, by updating to 5.2 those initializers no longer control anything. If that's accurate, setting them in the environments is only to specifically show that they are on, correct?

jdf-eos: The initializers were previously removed for this PR,  and only added back last night to see the result on the CI build.    As I have untangled the initialization path, I've been removing the initializer files, one by one,  to ensure that things are operating as before.  I'll finish that process shortly  and will let you know when this is "reviewable".

config/application.rb
```diff
@@ -66,6 +66,16 @@ class Application < Rails::Application
 
     # Caches user information in orders for the specified period of time
     config.orders_user_cache_expiration = 15.minutes
+    
+    # Make `form_with` generate non-remote forms.
+    config.action_view.form_with_generates_remote_forms = false
+    
+    # Store boolean values are in sqlite3 databases as 1 and 0 instead of 't' and
+    # 'f' after migrating old data.
+    config.active_record.sqlite3.represent_boolean_as_integer = true
```
ryanmiller-1, line 75: Is true the default value?

jdf-eos: yes, but there is a note in the code that indicates that it may be removed from the defaults list.  No explanation, so I thought it best to retain the current behavior.

PR #564, opened 2020-05-13
MMT-2254 updating proposal publish logging and adding comment to facilitate development on the manage proposals page

ryanmiller-1:


app/controllers/users_controller.rb
```diff
@@ -126,6 +126,7 @@ def finish_successful_login(profile)
     # Stores additional information in the session pertaining to the user
     store_profile(profile)
     Rails.logger.debug "Successful URS Login by user #{authenticated_urs_uid}" if session[:login_method] == 'urs'
```
charleshuang80, line 128: is this not true for proposal mode?
If the new line is necessary, then I would combine it to the end of this line: `|| Rails.configuration.proposal_mode`

ryanmiller-1: Different logging levels. I assumed we don't have debug on in SIT+

charleshuang80: ok

charleshuang80:
I'm good with the new changes also.

PR #567, opened 2020-05-14
MMT-2240 Rake task to run sloc report

jceaser:
add a rake command to run a sloc report. Plan for this to be executed in the future by a cron or some other trigger.

### DIFF
--- a/lib/tasks/sloc.rake
+++ b/lib/tasks/sloc.rake
@@ -16,6 +16,11 @@ namespace :sloc do
     if cloc.empty?
       puts 'run brew cloc before using this feature'
     else
+      # exclude Markdown, Dockerfiles and Shell scripts because they are not 
+      # contributing maintance of code and not code itself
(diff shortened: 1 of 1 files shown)


lib/tasks/sloc.rake
```diff
@@ -0,0 +1,41 @@
...
+
+namespace :sloc do
+  desc 'use the cloc command to analyze the code, warn if not installed'
+  task :cloc do
+    cloc=`which cloc`
+    if cloc.empty?
+      puts 'run brew cloc before using this feature'
+    else
+   	  shell_cmd = 'cloc ' \
+        '--exclude-lang=Markdown,Dockerfile,"Bourne Shell" ' \
```
ryanmiller-1, line 20: I think comments explaining why these things are being excluded should be added. I think we have a grasp of why these are being done now, future us might not remember.

jceaser: will do

PR #568, opened 2020-05-20
Mmt 2217 Enable Content Security Policy (CSP) Report Only logging

jdf-eos:


app/controllers/users_controller.rb
```diff
@@ -1,4 +1,6 @@
 # :nodoc:
+require 'browser'
```
ryanmiller-1, line 2: Necessary? Would have expected rails to handle this.
Is there a reason to put this in the users controller?

app/controllers/users_controller.rb
```diff
@@ -114,6 +118,13 @@ def associate_urs_and_launchpad_ids
     end
   end
 
+  def crpt
```
ryanmiller-1, line 121: Is there a reason to put this in the users controller?

app/controllers/users_controller.rb
jdf-eos: No, I couldn't find it. The community recommends browser gem.

app/controllers/users_controller.rb
jdf-eos: This is a request that is coming directly from the user's browser, not from any MMT code.  The user controller seemed like the right place. I thought about a new controller, but that seemed like overkill.

app/controllers/users_controller.rb
ryanmiller-1: I meant the require. If I remove it, I still get log messages like: 
`[3bf5d184-be54-4750-912e-85adfc51c6a2] CSP violation_report | violated_directive=script-src|effective-directive=|disposition=|blocked-uri=inline|document-uri=https://mmt.localtest.earthdata.nasa.gov/collection_drafts/39|source-file=https://mmt.localtest.earthdata.nasa.gov/collection_drafts/39|line-number=864|column-number=1|status-code=|referrer=https://mmt.localtest.earthdata.nasa.gov/collection_drafts/new|browser=Firefox|browser_version=68.0|MMT_version=2a319eae`

app/controllers/users_controller.rb
ryanmiller-1: Hmm. It feels out of place here to me. This controller is all related to logging in and establishing the user context for MMT.  I think I'm inclined to put this in its own controller, but I'd like another opinion.

charleshuang80: I agree with Ryan, I think this seems out of place in this controller. I don't think it's a problem to have a separate controller to deal with this. I suppose an argument could be made that this could just go into ApplicationController, but I think a separate controller makes more sense.

Also, I think the method name should be more descriptive - I think I know but I'm not totally certain what this stands for.

jceaser:
I was suspicious of seeing a new gem in Gemfile, but then I see how it is used. Parsing the user agent can be funny so I see how much was saved by using this gem.

PR #571, opened 2020-05-26
MMT-2257 Show the query on the estimate notification.

jdf-eos:


app/views/subscriptions/_estimate_notifications_modal.html.erb
```diff
@@ -5,6 +5,7 @@
     <p class="status-text"></p>
   </div>
   <div class="results-text">
+    <p class="query-text"></p>
```
ryanmiller-1, line 8: This can be 40k characters long; I think an option to collapse the text would be a good inclusion.

jdf-eos: Does MMT have a UX model for the collapse?

ryanmiller-1: We use eui accordions in the forms. I'm still thinking if there's some place we use it on modals

app/assets/javascripts/subscriptions.coffee
```diff
@@ -122,7 +122,17 @@ $(document).ready ->
...
+          if (query_text.length > 300)  # more than 300 chars gets a text box
+            $('.query-text-large').text(query_text)
+            $('.query-text-small').hide()
+            $('.query-title').show()
+            $('.query-text-large').show()
+          else
+            $('.query-text-small').text("Query: #{query_text}")
+            $('.query-text-small').show()
+            $('.query-title').hide()
+            $('.query-text-large').hide()
```
ryanmiller-1, line 135: What happens when the user enters a valid query to get an estimation and then generates an error response (simple test case would be an invalid collection concept)?  I think the query text needs to be managed outside of the success/error check.

jdf-eos: Seems to be ok.   The error modal looks like this:
Screen Shot 2020-05-28 at 2 29 48 PM

ryanmiller-1: Apologies, I missed something.

PR #577, opened 2020-05-29
MMT-2247: Feature branch for Creating and Updating UMM-T Drafts

charleshuang80:


app/concerns/controlled_keywords.rb
```diff
@@ -59,6 +59,21 @@ def set_service_keywords
     @service_keywords = keywords
   end
 
+  # for UMM-T v 1.0, Tool Keywords are using the same set of Service Keywords
+  # this is supposed to change in the future
+  def set_tool_keywords
+    keywords = fetch_science_keywords
+    if keywords.key? 'category'
+      keywords['category'].each do |category|
+        if category['value'] == 'EARTH SCIENCE'
```
jceaser, line 68: If the original person who created this keyword only knew how many times it would have been filtered out...

charleshuang80: was that person you ;)

PR #587, opened 2020-06-09
Mmt 1571

ChristianTrummer99:
Removed the Directory Names forms and created tests for this.


spec/features/collection_drafts/forms/hidden_fields_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    visit edit_collection_draft_path(draft)
+
+    within '.nav-top' do
+      click_on "Save"
+    end
+  end
+
+  it 'collection draft should still contain DirectoryNames' do
+    draft = CollectionDraft.first
+    expect(draft.draft.keys).to include('DirectoryNames')
```
ryanmiller-1, line 16: This test is only checking that the key still exists. I think verifying the value is unchanged would be better.

app/views/collection_drafts/forms/_metadata_information.html.erb
```diff
@@ -41,7 +41,7 @@
     ) %>
     <div class="eui-accordion__icon" tabindex="0">
       <i class="eui-icon eui-fa-chevron-circle-down"></i>
-      <span class="eui-sr-only">Toggle Metadata Dates</span>
+      <h3><span class="eui-sr-only">Toggle Metadata Dates</span></h3>
```
ryanmiller-1, line 44: Why is this being changed?

spec/features/collection_drafts/forms/hidden_fields_spec.rb
```diff
@@ -0,0 +1,16 @@
+# MMT-1571
```
charleshuang80, line 1: we don't need the ticket number, this is something we've actually been trying to remove from our tests

spec/features/collection_drafts/forms/hidden_fields_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+describe 'Removed field', js: true do
```
charleshuang80, line 2: Does this test work without `js: true`?
Where possible we should run our tests without js. Loading it adds extra time to our tests.

spec/features/collection_drafts/forms/metadata_information_form_spec.rb
```diff
@@ -1,7 +1,5 @@
 # MMT-288 MMT-295
```
charleshuang80, line 1: since you're in this test, can you remove this comment?

PR #595, opened 2020-06-16
MMT-2036 Delete Collections with Granules

ryanmiller-1:


app/controllers/collections_controller.rb
```diff
@@ -5,6 +5,8 @@ class CollectionsController < ManageCollectionsController
   before_action :set_collection
   before_action :ensure_correct_collection_provider, only: [:edit, :clone, :revert, :destroy]
 
+  CONFIRMATION_TEXT = 'I want to delete this collection and all associated granules.'.freeze
```
ryanmiller-1, line 8: remove period

ryanmiller-1:
what do we do when we change the provider and they don't want to proceed with the action?

app/views/collections/show.html.erb
```diff
@@ -70,7 +70,7 @@
             <% end %>
 
             <% if current_provider?(@provider_id) %>
-              <%= link_to 'Delete Collection Record', "#delete-record-modal", class: 'display-modal delete-collection eui-btn--link bar-after' %>
+              <%= link_to 'Delete Collection Record', "#delete-record-modal", class: 'display-modal delete-collection delete-collection-in-current-provider eui-btn--link bar-after' %>
```
ryanmiller-1, line 73: check if delete-collection does anything else or if it can be removed

lib/json_schema_form/umm_preview_elements/umm_preview_contacts.rb
```diff
@@ -181,13 +181,11 @@ def render_related_urls(contact_info)
                      end)
 
               concat(content_tag(:ul, class: 'arrow-tag-group-list') do
-                if ['GET SERVICE', 'GET DATA'].include? related_url['Type']
+                if related_url['Type']
```
ryanmiller-1, line 184: remove

ryanmiller-1:
Updated modal to refresh the page if they change provider and then use the cancel or close buttons.

app/views/collections/show.html.erb
```diff
@@ -115,11 +115,17 @@
...
           <div id="granules-modal" class="eui-modal-content">
-            <a href="javascript:void(0);" class="modal-close float-r"><i class="fa fa-times"></i><span class="is-invisible">Close</span></a>
-            <p>This collection cannot be deleted using the MMT because it has associated granules.  Use the CMR API to delete the collection and its granules.</p>
-            <p>
-              <a href="javascript:void(0)" class="eui-btn modal-close">Ok</a>
-            </p>
+            <a href=<%=current_provider?(@provider_id) ? "javascript:void(0);" : collection_path %> class="modal-close float-r"><i class="fa fa-times"></i><span class="is-invisible">Close</span></a>
+            <p>This collection has <%= @num_granules %> associated <%= 'granule'.pluralize(@num_granules) %>.</p> 
+            <p id="danger-warning">Deleting this collection will delete all associated granules.</p> 
+            <p>Please confirm that you wish to continue by entering 'I want to delete this collection and all associated granules' below.</p>
```
charleshuang80: I think the CONFIRMATION_TEXT constant should be used here

PR #598, opened 2020-06-16
MMT-2282 UMM-C update to 1.15.3

ChristianTrummer99:
- Updated schema to 1.15.3
- Added format description fields to each (File Archive and Distribution Information) accordion
- collection_factory_data: added the format descriptions for factories
- draft_helpers.rb: added Format descriptions to test forms 
- In preview gem:  _archive_and_distribution_information.html.erb (L47-54,121-128) added blocks for format description
- Changed MMT Gemfile (added branch: “MMT-2282”) -> changed back before opening PR
- Added tests to preview gem
- archive_and_distribution_information_preview_spec.rb (L29,37,48) added the appropriate tests
- Updated the “umm-merged.."  file with 1.15.3


app/helpers/search_helper.rb
```diff
@@ -63,14 +63,25 @@ def display_last_modified(record)
...
+    # TODO: Make this a permanent change when the UMM-T toggle is removed.
+    # search dropdown is wider if tools radio button is enabled
+    ' search-dropdown-wide' if Rails.configuration.umm_t_enabled
+  end
+
+  # TODO: Make this a permanent change when the UMM-T toggle is removed.
+  def search_field_size_modification
+    return if Rails.configuration.proposal_mode
+
+    'search-field-wide' if Rails.configuration.umm_t_enabled
```
jceaser, line 80: why does line 80 have no space at the beginning while line 73 and 69 do?

ryanmiller-1: This is a change from my PR, not his.  The answer is that the dropdown size modification is adding to a list of classes and the search field size modification is the only class.

ryanmiller-1:
Github is showing an empty file named 'git' being added in this PR. Please double check if it is in your repo.

app/helpers/search_helper.rb
jceaser: the function does not know how it is being used, should the space not be in the calling code so that these methods 'could' be reused for any condition?

ryanmiller-1: The goal is to get specific css applied, so I would argue that 80 is the one that should be changed (to have a space) rather than removing the spaces from the first function if you're trying to make these more reusable (though, by the nature of where this is used, I am skeptical we will find a place to reuse it).

ryanmiller-1: How should it be reusable? Should it be reusable to apply the css attached to those classes wherever this is called? If so, I think that 80 needs a space and 73 and 69 should remain the same. If the goal is to apply the css, I think you want the class added and not appended to some other class.  If it should be reusable such that the text is appendable to an existing class to create some new class that does not have any currently attached css, then 73 and 69 should drop their spaces.  I think the first case is more likely to be reused, but I expect that neither would be.

ryanmiller-1: I will make these more uniform in MMT-2267.  Discussed offline, all should have a space.

PR #599, opened 2020-06-22
MMT-2225, MMT-2226, MMT-2238: Add ability to publish UMM-T records and view published records

charleshuang80:
* add ability to publish UMM-T drafts to CMR
* add Mailers for publishing UMM-T records
* add show page for published Tool records
* make Tools search results show links
* add tests
* add disabled links for action items on published Tool preview page


app/mailers/draft_mailer.rb
```diff
@@ -47,4 +47,20 @@ def service_draft_published_notification(user, concept_id, revision_id, short_na
...
   end
+
+  def tool_draft_published_notification(user, concept_id, revision_id, short_name)
+    @user = user
+    @concept_id = concept_id
+    @short_name = short_name
+
+    @revision_id = revision_id.to_i
+    @is_update = @revision_id > 1
+    if @is_update
```
ryanmiller-1, line 58: I feel like all of these should be:
```
email_subject = if@is update
...
```

app/views/draft_mailer/tool_draft_published_notification.html.erb
```diff
@@ -0,0 +1,22 @@
...
+<% else %>
+  <h1 class="styled-heading"><%= "#{@short_name} Published" %></h1>
+<% end %>
+
+<div class="email-message">
+  <p>
+    <%= @user[:name] %>,
+  </p>
+  <p>
+    <%= "Your tool metadata record #{@short_name} has been successfully #{ @is_update ? 'updated in' : 'published to'} the CMR#{@email_env_note}. Your tool's concept ID is #{@concept_id}." %>
```
ryanmiller-1, line 13: If this text is the same in both html and text, do we want to generate it in the mailer?

spec/features/tool_drafts/publish_tool_draft_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    let(:tool_draft) { create(:empty_tool_draft, user: User.where(urs_uid: 'testuser').first) }
+    let(:incomplete_message) { 'This tool draft is not ready to be published. Please use the progress indicators on the draft preview page to address incomplete or invalid fields.' }
+
+    before do
+      login
+      visit tool_draft_path(tool_draft)
+      click_on 'Publish Tool Draft'
+    end
+
+    it 'displaysa message to the user' do
```
ryanmiller-1, line 25: typo

spec/features/tools/preview/compatability_and_usability_preview_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+  end
+
+  # TODO: remove after CMR-6332
+  after :all do
+    delete_response = cmr_client.delete_tool('MMT_2', @native_id, 'token')
+
+    raise unless delete_response.success?
+  end
+
+  context 'when examing the Compatibility and Usability sections' do
```
ryanmiller-1, line 18: examing typo

app/controllers/manage_metadata_controller.rb
```diff
@@ -191,6 +191,56 @@ def set_service_information
     end
   end
 
+  def set_tool
```
ryanmiller-1, line 194: I think my biggest concern with metaprogramming to write these functions is figuring out where they are to make modifications.  If the debugging tool says the error is in 'set_tool', we should be able to search 'set_tool' and find a thread to pull on.  I feel the complexity of the V/S/T ecosystem is already high (already has metaprogramming + the schema code).  I would definitely like to hear other opinions on this though.

app/views/tools/show.html.erb
```diff
@@ -0,0 +1,151 @@
...
+        data: { "provider": @provider_id, action_link: "change-provider-tool-#{@record_action}" }) %>
+    </p>
+    <%# TODO: this method does not exist yet. It should be created and used when %>
+    <%# additional actions are added to published Tool records %>
+    <%#= render_change_provider_tool_action_link(@record_action, @concept_id, @revision_id) %>
+  </div>
+<% end %>
+
+<% if @unsupported_version %>
+  <%# with MMT-1559 see if this can be refactored into a partial %>
```
ryanmiller-1, line 33: I'm pretty interested in partialing out as much of this as we can at some point.  These erbs are pretty redundant across V/S/T.

app/controllers/manage_metadata_controller.rb
charleshuang80: I don't think we need to resolve the metaprogramming issue in this PR, but we should figure it out before the end of the published Tool records functionality. I understand your concern, and these methods are actually in a somewhat confusing place because they are used by both drafts and published records so ended up way up here in the chain.

I leaning slightly towards refactoring these methods with metaprogramming similar to some of the base drafts methods, and adding more comments explaining these methods and also in the v/s/t controllers to say what method they are using and where.

app/views/draft_mailer/tool_draft_published_notification.html.erb
charleshuang80: I will take a look into both your mailer comments

app/controllers/manage_metadata_controller.rb
jceaser: I do feel that readability will take a hit by doing this, however, I think this should be weighed against the result of reduced lines of code (if that is in fact the result) which _should_ result in less code to support. Metaprogramming needs a good working example before it can be adopted across the board by other developers. I'm in favor of seeing an example, seeing documentation, and if the solution is radically different code, a wiki write up on it the practice. Having these things in place will increase the chances that the code is readable and increase the chances that we can take advantage of the technique.  As for this ticket, if you can get the change done in time for sprint end, the lets see it, otherwise a new ticket. Maybe we just leave a todo note for now?

app/controllers/manage_metadata_controller.rb
```diff
@@ -191,6 +191,56 @@ def set_service_information
...
+      @revisions = tool_data
+      latest = tool_data.first
+      meta = latest.nil? ? {} : latest.fetch('meta', {})
+
+      @old_revision = !@revision_id.nil? && meta['revision-id'].to_s != @revision_id.to_s ? true : false
+
+      break if latest && !@revision_id
+      break if latest && meta.fetch('revision-id', 0) >= @revision_id.to_i && meta['concept-id'] == @concept_id
+      attempts += 1
+      sleep 0.05
```
jceaser, line 233: I assume this is a guess. it would be nice if there was a long that could tell us how  many waits were needed and then after a time in production we could set this value to the average of that time? Then we spam CMR less and end up waiting mostly the correct amount of time? -- comment only made because I remember writing code that frequently called CMR only to be asked not to do that as we were hurting the performance.

app/views/draft_mailer/tool_draft_published_notification.html.erb
jceaser: that's a good point unless you plan to make the concept_id or short_name linkable in the HTML version

app/controllers/manage_metadata_controller.rb
charleshuang80: I don't think we'll forget about it, but am happy to add a todo. I don't think we need a new ticket, all the other Tools tickets to enable those action links should use these methods, plus at least one will need to tweak these methods. I suggest we tackle it in one of those upcoming tickets.

app/controllers/manage_metadata_controller.rb
charleshuang80: I believe that doing it this way was originally suggested by CMR

charleshuang80: Ah, I didn't understand you were suggesting we log the attempts. I'll look into that.

PR #600, opened 2020-06-23
MMT-2293 - updating the ACLs to match changes in CMR

jceaser:


lib/tasks/acls.rake
```diff
@@ -88,10 +92,12 @@ namespace :acls do
           'permissions' => ['create', 'read', 'update', 'delete']
```
jceaser, line 92: change the perms to only read and update, also clear cache before just to be safe

app/policies/subscription_policy.rb
```diff
@@ -9,24 +18,23 @@ def edit?
...
-    @read.nil? ? fetch_granted_permissions(action: 'read', user: user.user, type: 'provider', target: 'EMAIL_SUBSCRIPTION_MANAGEMENT', token: user.token) : @read
+    @read.nil? ? fetch_granted_permissions(action: 'read', user: user.user, type: 'provider', target: 'SUBSCRIPTION_MANAGEMENT', token: user.token) : @read
   end
 
   def destroy?
-    @delete.nil? ? fetch_granted_permissions(action: 'delete', user: user.user, type: 'provider', target: 'EMAIL_SUBSCRIPTION_MANAGEMENT', token: user.token) : @delete
+    update?
   end
 
   def estimate_notifications?
```
ryanmiller-1, line 40: I think this can just be `update?` now.

spec/features/manage_cmr/subscriptions/create_subscriptions_spec.rb
```diff
@@ -1,4 +1,4 @@
-describe 'Creating Subscriptions' do
+allow_any_instance_ofdescribe 'Creating Subscriptions' do
```
ryanmiller-1, line 1: ?

spec/features/manage_cmr/subscriptions/create_subscriptions_spec.rb
```diff
@@ -52,7 +49,6 @@
             # This is necessary to load the delete button for the subscription
             # clean up.
             # TODO: Remove when we can reset_provider
-            allow_any_instance_of(SubscriptionPolicy).to receive(:destroy?).and_return(true)
```
ryanmiller-1, line 55: Can remove the top comment here, the TODO should be updated with the CMR ticket number (6332)

spec/features/manage_cmr/subscriptions/index_subscriptions_spec.rb
```diff
@@ -85,6 +84,7 @@
     context 'when viewing the index page with full permissions' do
       before do
         allow_any_instance_of(SubscriptionPolicy).to receive(:edit?).and_return(true)
+        allow_any_instance_of(SubscriptionPolicy).to receive(:update?).and_return(true)
```
ryanmiller-1, line 87: I think that this should either be index, edit, destroy, and create or index, and update.

app/policies/subscription_policy.rb
charleshuang80: agree

spec/features/manage_cmr/subscriptions/index_subscriptions_spec.rb
charleshuang80: I think this context block should be pulled out to be at the same indentation level as line 34. Then you don't have to mock any of the policy methods, because they are being granted in the first before :all block.

lib/tasks/acls.rake
charleshuang80: this still needs to be addressed

spec/features/manage_cmr/subscriptions/test_query_spec.rb
```diff
@@ -17,7 +17,7 @@
   before do
     login
 
-    allow_any_instance_of(SubscriptionPolicy).to receive(:create?).and_return(true)
+    allow_any_instance_of(SubscriptionPolicy).to receive(:update?).and_return(true)
     allow_any_instance_of(SubscriptionPolicy).to receive(:show?).and_return(true)
```
charleshuang80, line 21: I think both of these mocked permissions will be provided by the before all block.

spec/features/manage_cmr/subscriptions/test_query_spec.rb
```diff
@@ -191,9 +191,7 @@
 
   before do
     login
-    allow_any_instance_of(SubscriptionPolicy).to receive(:create?).and_return(false)
     allow_any_instance_of(SubscriptionPolicy).to receive(:update?).and_return(true)
```
charleshuang80, line 194: I think this will also be provided by the before all block.

PR #601, opened 2020-06-23
MMT-2267 Updating to UMM-S 1.3.3

ryanmiller-1:
Schema Changes:
Erich 'flipped' the schema so the definitions are first. This has no impact on us, but it part of the diff for that file.
Updated Supported reformattings to be 1 => many
Added formats to enum list
Removed fields from URL top level field
Added required field to URL top level field
Replaced Related URLs in Service Contacts with Online Resources
Updated Url => URL

Changes in MMT:
Updated forms and preview to accommodate schema changes
Data migration
Added new preview element type to schema generated preview
Removed a lot of now unnecessary UMM-S relatedurl code.


app/helpers/online_resource_helper.rb
```diff
@@ -0,0 +1,26 @@
...
+              concat content_tag(:p, online_resource['Name']) if online_resource['Name']
+              concat content_tag(:p, online_resource['Description']) if online_resource['Description']
+
+              concat(if online_resource['Linkage']
+                       link_to online_resource['Linkage'], online_resource['Linkage'], title: online_resource.fetch('Name', 'Online Resource Linkage')
+                     else
+                       'Not provided'
+                     end)
+
+              concat content_tag(:p, "Protocol: #{online_resource['Protocol']}")
```
ryanmiller-1, line 17: Make a change for if the fields aren't present

lib/json_schema_form/umm_preview_elements/umm_preview_link.rb
```diff
@@ -0,0 +1,10 @@
+# render basic data as a link
+# Used in UMM-S for URL::URLValue
+
+# :nodoc:
+class UmmPreviewLink < UmmPreviewElement
+  def render
+    title = 'USE SERVICE API' if schema_type == 'service'
```
ryanmiller-1, line 7: check with Erich

PR #604, opened 2020-06-26
Hotfix/MMT-2267 Fix validation bug and make more consistent with UMM-T form appearance

ryanmiller-1:


spec/features/service_drafts/forms/validation_spec.rb
```diff
@@ -68,6 +68,10 @@
             fill_in 'Version', with: '1.0'
             fill_in 'service_draft_draft_description', with: 'Service Description'
           end
+          
+          within '#url' do
+            fill_in 'URL Value', with: 'an URL'
```
william-valencia, line 73: This is probably being nitpicky and not a big deal, but I think the proper article here to use is 'a URL'. The U in URL is pronounced as 'you' which would be a consonant which would make it ideally use 'a'

PR #606, opened 2020-07-01
Mmt 1421 Strange error when deleting collection records in MMT

jdf-eos:


app/controllers/collections_controller.rb
```diff
@@ -47,7 +47,7 @@ def destroy
     if delete_response.success?
       flash[:success] = I18n.t('controllers.collections.destroy.flash.success')
       Rails.logger.info("Audit Log: Collection with native_id #{@native_id} was deleted for #{provider_id} by #{session[:urs_uid]}")
-      redirect_to collection_revisions_path(id: delete_response.body['concept-id'], revision_id: delete_response.body['revision-id'])
+      redirect_to collection_revisions_path(id: delete_response.body['concept-id'])
```
ryanmiller-1, line 50: We use get_revisions to poll CMR because we ask for the new revision back faster than they can deal with all the repercussions.  Without a revision_id being supplied here, it looks like the get_revisions method (in set_collection) is accepting the first round of revisions and returning rather than trying to wait for CMR to finish updating.  With enough attempts, this should result in deleting a collection and loading the revisions page without the revision that is deleted. 
Screen Shot 2020-07-01 at 10 37 09 AM

ryanmiller-1: I think we will need a different solution to avoid introducing this race.

jdf-eos: Ah, so the race condition has been there forever.  I think that we should leave this fix in; it does fix the bug identified in the ticket.  In addition,  we should call get_revisions before the redirect at line 50.  This will ensure that the delete operation is complete in CMR before we get to the revisions page.

ryanmiller-1: This solution also allows users to view collections which have been deleted rather than being informed that the record is deleted because navigating to a concept_id will cause MMT to load the most recent non-deleted record via the select revision method.

jdf-eos: What scenario would produce that?   I didn't see these errors in the test cases.

ryanmiller-1: If a user navigates back in the browser or navigates directly to a page such as:
https://mmt.sit.earthdata.nasa.gov/collections/C1200371242-MMT_2

In SIT it displays a message about that collection being deleted.  In MMT-1421 it loads metadata as though it is not deleted.

jdf-eos: OK, I reworked the solution so that everything  works the same as before.  The only difference is in the handling of the delete collection case.   The redirect will go to the specific version of the delete. An additional parameter is passed in the redirect to enable set_collection and select_revision to work correctly for that specific case.       This eliminated the race condition, so no changes to spec files are required.

PR #607, opened 2020-07-02
MMT-2234 Adding download json for tools

ryanmiller-1:
adding download json for tools

### DIFF
--- a/spec/features/collections/downloading_xml_spec.rb
+++ b/spec/features/collections/downloading_xml_spec.rb
@@ -26,16 +26,17 @@ describe 'Downloading Collection XML', js: true do
         before do
           @file = "#{Rails.root}/#{@concept_id}.echo10"
           click_on 'ECHO 10'
+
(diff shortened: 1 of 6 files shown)


spec/features/collections/downloading_xml_spec.rb
```diff
@@ -26,16 +26,17 @@
...
           sleep(0.1)
+        end
+
+        after do
           FileUtils.rm @file if File.exist?(@file)
         end
 
         it 'downloads the file' do
-          expect(File.exist?(@file))
+          expect(File.exist?(@file)).to eq(true)
```
william-valencia, line 39: Why the extra .to eq(true) call?

ryanmiller-1: Without a matcher it isn't actually testing anything.  `expect(false)` passes.  It is unclear to me why it does not issue a syntax warning for this or what the use case there is for this behavior.

PR #608, opened 2020-07-06
MMT-2230 Adding tool cloning

ryanmiller-1:


app/views/tools/show.html.erb
```diff
@@ -94,14 +92,11 @@
           <% end %>
         <% end %>
 
-        <%# TODO: All links commented out and disabled links added for MMT-2238 %>
```
charleshuang80, line 97: If not all the links are being enabled, should we keep the comments and possibly move them down to be near links that are still disabled?

ryanmiller-1: Moving it around was going to cause more merge conflicts, so I decided to remove it.  Especially since I expect to have all of these links working by the end of the week.

PR #610, opened 2020-07-07
MMT-2231 Add revisions page to published UMM-T page

ryanmiller-1:


app/views/tools/revisions.html.erb
```diff
@@ -0,0 +1,119 @@
...
+  <section class="errors">
+    <div class="eui-banner--danger">
+      <div class="eui-banner__message">
+        <ul class="no-bullet">
+          <% @errors.each do |error| %>
+            <li>
+              <%= "#{error[:field]}, " if error[:field] %>
+              <%= error[:error] %>
+              <% if error[:request_id] %>
+                <a href="javascript:feedback.showForm({details: '\nFill in details above this line. Please try to be as specific as possible.\n--------------------\n\nRequest ID: <%= error[:request_id] %>'});">Click here to submit feedback</a>
```
charleshuang80, line 16: I did not know we had forms on some of our pages go through the Feedback Module. I think we should see if Rosy and Alicia know that this is being done for revisions and some drafts, and see if they would like to use it in other locaitons.

PR #613, opened 2020-07-09
MMT-2311: Add endpoint for CMR translation to CMR client

ChristianTrummer99:
- I've added a rake task called translate collections which accepts a collection record that is saved locally as an xml file
- the CMR client is then used to translate the record to UMM json then back to it's native format; the original record and it's translated version are compared against one another, generating a lossiness report
- nokogiri-diff was added to the gemfile


ChristianTrummer99:
- added a parameter to the rake task so that individual (nokogiri) changes can be selectively displayed

lib/tasks/translate_collections.rake
```diff
@@ -0,0 +1,277 @@
...
+    native_converted_hash = Hash.from_xml(back_to_native.body)
+    native_converted_xml = back_to_native.body
+    back_to_native.success? ? puts("successful translation to native format \n\n") : puts("Native format translation failure \n\n")
+
+    # nokogiri output
+    o = Nokogiri::XML(native_original_xml) { |config| config.strict.noblanks }
+    c = Nokogiri::XML(native_converted_xml) { |config| config.strict.noblanks }
+    # puts args.display, args.display.class
+      o.diff(c, {:added => true, :removed => true}) do |change,node|
+        next if path_leads_to_list?(node.parent.path, native_original_hash, native_converted_hash)
```
ryanmiller-1, line 33: Aren't all the things that trigger this list items?  Is something missing here?  It seems like you could create your list of lists to check from this data.

lib/tasks/translate_collections.rake
```diff
@@ -0,0 +1,277 @@
...
+          puts("#{change}: #{node.to_xml}".ljust(60) + node.parent.path) if args.disp.eql? 'show'
```
ryanmiller-1, line 34: Indentation is off for these.

lib/tasks/translate_collections.rake
```diff
@@ -0,0 +1,277 @@
+require 'libxml_to_hash'
+
+namespace :collection do
+  desc 'Translate a collection from native format to UMM JSON and back to native format'
+  task :translate, [:file, :format, :disp, :version] => :environment do |_task, args|
+    args.with_defaults(:version => '1.15.3')
+    args.with_defaults(:disp => 'show')
+
+    puts 'FORMAT INVALID' unless args.format == ('echo10' || 'dif10' || 'iso19115')
```
ryanmiller-1, line 9: Should fail gracefully here if the format is invalid o/w results text can push the error out of the window and the user may not know that it was invalid.

lib/tasks/translate_collections.rake
```diff
@@ -0,0 +1,277 @@
...
+          end
+        end
+      end
+    elsif iterable.is_a? Array
+      iterable.each do |item|
+        if item.is_a?(Hash) && hash_to_list_of_values(item).include?(value)
+          dir += get_dir(value,item)
+          return dir
+        elsif item.is_a?(Array) && array_to_list_of_values(item).include?(value)
+          puts "\n\n\n\n\n\n\n\n USED THIS SECTION \n\n\n\n\n\n\n\n\n\n\n\n"
```
ryanmiller-1, line 237: rm

lib/tasks/translate_collections.rake
```diff
@@ -0,0 +1,277 @@
+require 'libxml_to_hash'
```
ryanmiller-1, line 1: General comment for this file: there's a lot of custom code involving the manipulation of data structures without a lot of comments.  In particular, comments about the expected shapes of the inputs and outputs would be helpful throughout.

lib/tasks/translate_collections.rake
```diff
@@ -0,0 +1,275 @@
...
+    puts 'INVALID FORMAT' if !args.format.eql?('echo10' && 'dif10' && 'iso19115')
+
+    filename = args.file.split('/')[-1]
+    puts "\nTranslating #{filename} to UMM JSON..."
+
+    native_original_xml = File.read(args.file)
+    native_original_hash = Hash.from_xml(native_original_xml)
+
+    #translate to UMM
+    umm = cmr_client.translate_collection(native_original_xml, "application/#{args.format}+xml", "application/vnd.nasa.cmr.umm+json;version=#{args.version}", skip_validation=true )
```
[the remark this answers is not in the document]
ChristianTrummer99: Got it, will do

lib/tasks/translate_collections.rake
```diff
@@ -0,0 +1,275 @@
...
+    umm_json = umm.body.to_json
+    umm.success? ? puts("\nsuccessful translation to UMM") : puts("\nUMM translation failure")
+
```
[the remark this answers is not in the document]
ChristianTrummer99: Good point, I'll use abort(msg)

lib/tasks/translate_collections.rake
```diff
@@ -0,0 +1,275 @@
...
+  def compare_arrays(diff_hash, original_hash, converted_hash)
+    dif_hash = diff_hash.clone
+    original = original_hash.clone
+    converted = converted_hash.clone
+    paths = get_list_paths(dif_hash, original, converted)
+
+    paths.each do |path|
+      org_array = hash_navigation(path, original)
+      org_arr = org_array.clone
```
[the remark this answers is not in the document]
ChristianTrummer99: Removing the clones erroneously changes the output. I do agree that this is inelegant though. The reason Im cloning the arrays used in the for loops that follow is that using the .delete method on the same array that is used as the enumerable object in the for loop throws off the comparison that occurs between the original and converted arrays. By cloning org_array and conv_array, I now have copies of the enumerable that can be edited while the looping occurs. Those edited arrays are then used in the following lines to list the items that were added/removed. 

This explanation is pretty confusing for me to write which tells me that there must be a way to make this more readable or, better, another - clearer - way to achieve the same output. If you have suggestions, or questions, I'm happy to hear them.

lib/tasks/translate_collections.rake
```diff
@@ -0,0 +1,275 @@
...
+  end
+
+  def find_difference_bt_hashes(org, conv)
+    missing = Hash.new
+    if org.eql? conv
+      return missing
+    else
+      org.each do |org_key,org_value|
+        conv_value = conv[org_key]
+        if conv.key? org_key
```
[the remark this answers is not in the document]
ChristianTrummer99: Thanks, good catch

lib/tasks/translate_collections.rake
```diff
@@ -0,0 +1,277 @@
...
+
+      org_arr = Array.wrap(org_arr) unless org_arr.is_a?(Array)
+      org_array = Array.wrap(org_array) unless org_array.is_a?(Array)
+      conv_arr = Array.wrap(conv_arr) unless conv_arr.is_a?(Array)
+      conv_array = Array.wrap(conv_array) unless conv_array.is_a?(Array)
+
+      for conv_item in conv_array
+        for org_item in org_array
+          if org_item.eql? conv_item
+            org_arr.delete(org_item)
```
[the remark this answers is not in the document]
ChristianTrummer99: very true!

lib/tasks/translate_collections.rake
```diff
@@ -0,0 +1,277 @@
...
+      end
+    end
+    missing
+  end
+
+  def get_dir(value, hash_or_arr)
+    iterable = hash_or_arr.clone
+    dir = String.new
+    if iterable.is_a? Hash
+      unless iterable.key(value).nil?
```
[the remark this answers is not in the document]
ChristianTrummer99: Got it, thanks.

lib/tasks/translate_collections.rake
ChristianTrummer99: Will do! Thanks for parsing through all of this so carefully - I'm taking note of each comment for my future reference

lib/tasks/translate_collections.rake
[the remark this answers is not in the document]
ChristianTrummer99: Good point, clones removed. Org_arr is the array that is being edited (with the .delete method) and org_array is the array being iterated through. The same is true for conv_arr and conv_array. If I only used org_arr, and continued to delete items as the loop iterated, the inner for loop would be changing size after each iteration of the outer for loop.

lib/tasks/translate_collections.rake
ChristianTrummer99: I'm a little unsure of your question. If I understand correctly, you're asking 'why can't you use the nokogiri output to find all the paths that lead to array changes instead of writing custom code to find those paths again?' 

If that is your question, I think you make a great point, however, I believe I looked into that originally, but found that nokogiri wasn't recognizing some changes at all or, further, wasn't generating useful paths for some list additions/removals. But I will certainly look at it again to see if there is optimization potential. **Please confirm if I understood/misunderstood your question**

lib/tasks/translate_collections.rake
```diff
@@ -0,0 +1,275 @@
...
+      end
+
+      org_arr.each do |item|
+        path_with_index = path + "[#{org_array.index(item)}]"
+        puts "-: ". + path_with_index
+      end
+
+      conv_arr.each do |item|
+        path_with_index = path + "[#{conv_array.index(item)}]"
+        puts "+: " + path_with_index #THIS INDEX DOESN'T MAKE SENSE
```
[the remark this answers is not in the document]
ChristianTrummer99: this is a subjective UX related question - just sent a long slack message explanation/question

lib/tasks/translate_collections.rake
ryanmiller-1: You understood the question correctly.

lib/tasks/compare_xml_collections.rake
```diff
@@ -0,0 +1,121 @@
...
+    back_to_native.success? ? puts("successful translation to native format \n\n") : abort("Native format translation failure \n\n")
+
+    # nokogiri output
+    nokogiri_original = Nokogiri::XML(native_original_xml) { |config| config.strict.noblanks }
+    nokogiri_converted = Nokogiri::XML(native_converted_xml) { |config| config.strict.noblanks }
+
+    ignored_paths = Array.new
+
+    nokogiri_original.diff(nokogiri_converted, {:added => true, :removed => true}) do |change,node|
+      if node.parent.path.include?('[') && !ignored_paths.include?(node.parent.path.split('[')[0])
```
ryanmiller-1, line 35: I would memoize node.parent.path.split('[') given that you use it several times.  Ruby probably manages this under the hood for you, but if it doesn't, this will save a bunch of string operations.

lib/tasks/compare_xml_collections.rake
```diff
@@ -0,0 +1,121 @@
...
+        end
+      end
+    else
+      hash = hash[dir]
+    end
+    hash
+  end
+
+  def array_comparison(path, original_hash, converted_hash)
+    # this is a 'less iterative' version of compare_arrays. Args: a single path, the original hash, and the converted hash.
```
ryanmiller-1, line 85: I would remove the history from these comments. While helpful to watch the PR as it changes, they reference something that future developers won't be able to look up.

lib/tasks/compare_xml_collections.rake
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  def path_leads_to_list?(path, org_hash, conv_hash)
+    # this method takes a path string (and the full original and converted hashes) and outputs true if the path string contains a list; else false
+    org_hash_path = hash_navigation(path, org_hash)
+    conv_hash_path = hash_navigation(path, conv_hash)
+
+    if path.include?("[") && path.include?("]")
+      bool = true
+    elsif org_hash_path.is_a?(Hash) && conv_hash_path.is_a?(Hash)
```
ryanmiller-1, line 55: I would add a comment about why the keys length is valuable/necessary.

lib/tasks/compare_xml_collections.rake
```diff
@@ -0,0 +1,121 @@
...
+      bool = false
+    end
+    bool
+  end
+
+  def hash_navigation(dir, hash)
+    # Passed a path string and the hash being navigated. This method parses the path string and
+    # returns the hash at the end of the path
+    dir = dir.split '/'
+    if dir.is_a? Array
```
ryanmiller-1, line 70: I think this is an unnecessary type check

lib/tasks/compare_xml_collections.rake
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  def array_comparison(path, original_hash, converted_hash)
+    # this is a 'less iterative' version of compare_arrays. Args: a single path, the original hash, and the converted hash.
+    # Rather than finding all the array paths and using those to find the array differences, the array paths are individually
+    # supplied by the nokogiri gem; this reduces redundancy
+    org_array = hash_navigation(path, original_hash)
+    conv_array = hash_navigation(path, converted_hash)
+
+    org_array.is_a?(Array) ? org_arr = org_array.clone : org_arr = Array.wrap(org_array)
```
ryanmiller-1, line 91: I think that the names of x_arr and x_array should clearly highlight their intended use to increase readability.

ChristianTrummer99: Ohh, I see. Will do

lib/tasks/compare_xml_collections.rake
```diff
@@ -0,0 +1,124 @@
...
+    end
+    bool
+  end
+
+  def hash_navigation(dir, hash)
+    # Passed a path string and the hash being navigated. This method parses the path string and
+    # returns the hash at the end of the path
+    dir = dir.split '/'
+    dir.each do |key|
+      if !key.empty? && hash.is_a?(Hash)
```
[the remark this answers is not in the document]
ChristianTrummer99: I decided to go with `hash.key?(key)`, (more readable than `!key.empty?)`. That check does need to be there since the first character of each path is '/' which causes an empty key to be at the front of `each path.split('/').each` enumerable and therefore the options are to remove that first character (oddly, this caused issues - maybe I did it wrong) or add the extra if condition, in this case I think `hash.key?(key)` is optimal.

lib/tasks/compare_xml_collections.rake
```diff
@@ -0,0 +1,121 @@
...
+    bool
+  end
+
+  def hash_navigation(path, hash)
+    # Passed a path string and the hash being navigated. This method parses the path string and
+    # returns the hash at the end of the path
+    path.split('/').each do |key|
+      if hash.key?(key) && hash.is_a?(Hash)
+        hash = hash[key]
+      elsif hash.is_a? Array
```
charleshuang80, line 77: If `hash` could be an array, I think the comment needs to be updated, and also the parameter might need to be renamed?
Also, above and below, you're using `is_a?` with parens, but not here. They should all be consistent.

lib/tasks/compare_xml_collections.rake
```diff
@@ -0,0 +1,121 @@
...
+        return hash
+      end
+    end
+    hash
+  end
+
+  def array_comparison(path, original_hash, converted_hash)
+
+    pretranslation_array = hash_navigation(path, original_hash)
+    post_translation_array = hash_navigation(path, converted_hash)
```
charleshuang80, line 87: seems inconsistent to have pre without an underscore but post with

lib/tasks/compare_xml_collections.rake
```diff
@@ -0,0 +1,121 @@
...
+
+    pretranslation_array.is_a?(Array) ? lost_items_arr = pretranslation_array.clone : lost_items_arr = Array.wrap(pretranslation_array)
+    pretranslation_array = Array.wrap(pretranslation_array) unless pretranslation_array.is_a?(Array)
+    post_translation_array.is_a?(Array) ? added_itmes_arr = post_translation_array.clone : added_itmes_arr = Array.wrap(post_translation_array)
+    post_translation_array = Array.wrap(post_translation_array) unless post_translation_array.is_a?(Array)
```
charleshuang80, line 92: I think these lines could use some comments to explain what is going on. Also I think you can drop the `unless` conditionals after `Array.wrap`

lib/tasks/compare_xml_collections.rake
```diff
@@ -0,0 +1,121 @@
...
+
+    # org_arr and conv_arr are copies of org_array and conv_array, respectively.
```
charleshuang80, line 94: In ruby we want to try and write things out to be more descriptive (even though it can take up more space). I like this explanation but I think you should use longer more descriptive names (including typing out what org and conv stand for), as it gets a bit confusing to read through the code below.

lib/tasks/compare_xml_collections.rake
```diff
@@ -0,0 +1,121 @@
...
+    # The *_arr values are edited during the comparison between the org_array and conv_array arrays
+    # and so the *_array arrays are used to maintained a full version of each array for indexing the items in the following lines.
+
+    for conv_item in post_translation_array
+      for org_item in pretranslation_array
+        if org_item.eql? conv_item
```
charleshuang80, line 100: minor thing, but since we are using equal signs for comparisons elsewhere, you should probably also use them here

lib/tasks/compare_xml_collections.rake
```diff
@@ -0,0 +1,121 @@
+require 'libxml_to_hash'
+
+namespace :collection do
+  desc 'Translate a collection from native format to UMM JSON and back to native format'
```
charleshuang80, line 4: I would be concerned about this somewhat complicated task not having tests. But since you are going to be adding the functionality into the app I expect there to be tests added then.

lib/tasks/compare_xml_collections.rake
```diff
@@ -0,0 +1,121 @@
...
+  task :loss, [:file, :format, :disp, :version] => :environment do |_task, args|
+    args.with_defaults(:version => '1.15.3')
+    args.with_defaults(:disp => 'show')
+
+    abort 'FORMAT INVALID' unless args.format.eql?('echo10') || args.format.eql?('dif10') || args.format.eql?('iso19115')
```
charleshuang80, line 9: made a comment about using equal signs before. still think we should be consistent. [this post](https://stackoverflow.com/questions/7156955/whats-the-difference-between-equal-eql-and) and the links might help.
Also, what's the reason these are the only formats we can use? A comment might help.

lib/tasks/compare_xml_collections.rake
ChristianTrummer99: I forgot to update the comments after I changed the names per a change request Ryan made a few days ago - I'll fix it right up

lib/tasks/compare_xml_collections.rake
ChristianTrummer99: correct.

lib/tasks/compare_xml_collections.rake
ChristianTrummer99: I appreciate the link - just read through it. As for the reason for only using those formats, I am not sure what the other formats are (that cmr supports for collections). This is from the [cmr data partner's guide](https://wiki.earthdata.nasa.gov/display/CMR/CMR+Data+Partner+User+Guide) so I just assumed these were the only ones: 
Screen Shot 2020-07-20 at 9 56 53 AM

PR #614, opened 2020-07-13
MMT-2229 Delete published tools

william-valencia:


spec/features/tools/delete_tool_spec.rb
```diff
@@ -0,0 +1,96 @@
+require 'rails_helper'
```
charleshuang80, line 1: don't need this we're actually trying to remove from any files we see this in

spec/features/tools/delete_tool_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+
+describe 'Delete tool', reset_provider: true, js: true do
+  before :all do
+    @ingested_tool, _concept_response, _native_id_1 = publish_tool_draft
+
+    @ingested_tool_for_delete_messages, _concept_response, @native_id_2 = publish_tool_draft
+  end
+
+  after :all do
+    delete_response = cmr_client.delete_tool('MMT_2', @native_id_2, 'token')
```
charleshuang80, line 11: I think we should add a comment to remove this after the reset_provider ticket is resolved (can't recall the ticket number right now)

ryanmiller-1: CMR-6332

spec/features/tools/delete_tool_spec.rb
william-valencia: Will remove. Was copied in from the similar delete_service_spec.rb. Do we have a ticket to remove it from all the other places it is already in?

spec/features/tools/delete_tool_spec.rb
william-valencia: Comment added

spec/features/tools/tool_permissions_spec.rb
```diff
@@ -84,46 +82,38 @@
...
+
+          click_on 'Delete Tool Record'
+        end
+
+        it 'displays a modal informing the user they need to switch providers' do
+          expect(page).to have_content("Deleting this tool #{modal_text}")
+
+          # Remove this section after CMR-6332 is resolved
+          delete_response = cmr_client.delete_tool('MMT_2', @native_id_2, 'token')
+          raise unless delete_response.success?
```
william-valencia, line 99: Noticed that this test was still making searching tools have the error from CMR-6332. Should be able to remove this part after CMR-6332 is resolved.

PR #616, opened 2020-07-15
MMT-2232/2233

ryanmiller-1:
Also updated a subscriptions test which was causing intermittent build failures.


ryanmiller-1:
github isn't updating with new commit, will reopen

PR #620, opened 2020-07-23
MMT-2301 Refactoring V/S/T to use more partials and shared methods

ryanmiller-1:
Added a set_metadata method to replace set_service, set_tool, and set_variable when they all did roughly the same thing
Used partials on show pages when the show pages used the same code for V/S/T published records
Reactivated some disabled tests
Removed some manual deletions in tool and subscription tests associated with a CMR bug


app/controllers/manage_metadata_controller.rb
```diff
@@ -87,148 +87,61 @@ def compare_resource_umm_version(concept_id = nil)
...
+    headers = { 'Accept' => "application/#{metadata_version}; charset=utf-8" }
+    concept_response = cmr_client.get_concept(@concept_id, token, headers, @revision_id)
 
-    @service = if service_concept_response.success?
-                 service_concept_response.body
+    metadata = if concept_response.success?
+                 concept_response.body
                else
-                 Rails.logger.error("Error retrieving concept for Service #{@concept_id} in `set_service`: #{service_concept_response.clean_inspect}")
+                 Rails.logger.error("Error retrieving concept for #{@metadata_type} #{@concept_id} in `set_tool`: #{concept_response.clean_inspect}")
```
charleshuang80, line 119: '`set_metadata`'

app/controllers/manage_metadata_controller.rb
```diff
@@ -238,14 +151,15 @@ def set_tool_information
       attempts += 1
       sleep 0.05
       # TODO when/if we refactor these V/S/T methods, make sure the logging is accurate for record type being retrieved
```
charleshuang80, line 153: comment still needed?

app/controllers/manage_metadata_controller.rb
```diff
@@ -87,148 +87,61 @@ def compare_resource_umm_version(concept_id = nil)
...
-            
-    set_tool_information
+    set_metadata_information
   end
 
-  def set_tool_information
-    # search for tool by concept id to get the native_id and provider_id
-    # if the tool is not found, try again because CMR might be a little slow to index if it is a newly published record
+  def set_metadata_information
+    # search for metadata record by concept id to get the native_id and provider_id
```
charleshuang80, line 130: for future devs, do we want to say here that the retrying was recommended by CMR?

app/helpers/tools_helper.rb
```diff
@@ -9,4 +9,9 @@ def render_change_provider_tool_action_link(tool_action, concept_id, revision_id
       link_to('Delete Service', tool_path(concept_id), method: :delete, class: 'is-invisible', id: 'change-provider-tool-delete')
     end
   end
+
+  # Simplifies shared code between V/S/T
```
charleshuang80, line 13: I'm not sure I agree with what the comment is saying, can you elaborate?

app/helpers/tools_helper.rb
```diff
@@ -9,4 +9,9 @@ def render_change_provider_tool_action_link(tool_action, concept_id, revision_id
       link_to('Delete Service', tool_path(concept_id), method: :delete, class: 'is-invisible', id: 'change-provider-tool-delete')
```
charleshuang80, line 9: So these links say 'Service', is that correct?

app/views/services/show.html.erb
```diff
@@ -1,134 +1,30 @@
...
-  </div>
-  <div class="no-access">
-    <h3 class="access-title">Unconfirmed UMM Version</h3>
-    <p class="access-message">
-      <%= "It appears we currently cannot confirm the UMM version of this #{@published_resource_name}. In order to prevent unintentional data loss, editing this #{@published_resource_name} is currently unavailable." %>
-      Please try again or contact <%= mail_to('support@earthdata.nasa.gov', 'Earthdata Support') %>
-    </p>
-  </div>
-<% end %>
+<%= render partial: 'base_published_record/direct_access_without_permissions' %>
```
charleshuang80, line 6: I really like how many lines we're replacing with the partials

ChristianTrummer99: I concur

spec/features/variables/empty_response_when_setting_variable_spec.rb
```diff
@@ -41,26 +39,15 @@
     end
 
     context 'when visiting the show page of a non-existent variable' do
-      # This test should be revisited with MMT-2301. MMT-1557 made a change that if a concept cannot be retrieved to compare UMM Versions, editing is not allowed so we are only showing a page notifying the user the version cannot be confirmed
       before do
-        visit variable_path('fake-concept-id')
+        visit variable_path('Vfake-concept-id')
```
ChristianTrummer99, line 43: I think if this `Vfake-concept-id` was changed to `var-fake-concept-id` and the `Sfake-concept-id` was changed to `service-fake-concept-id` it may increase readability, because at first I wasn't sure if it was a typo or intentional. A comment may help too

ryanmiller-1: I will add comments.  They need to contain the 'signature' letter (V or S) for the CMR concept to be parsed correctly in set_metadata.

PR #621, opened 2020-07-29
MMT-2313: Make crude lossiness report available in web browser from MMT

ChristianTrummer99:
- added route to make loss report viewable in browser
- added method to collections_controller to view report as either text or json
- created helper which performs the comparison and generates the report
- created helper spec file and feature spec file
- created a file in spec/support to hold the sample reports being used in the helper spec


app/helpers/loss_report_helper.rb
```diff
@@ -0,0 +1,211 @@
...
+  end
+
+  def token
+    if session[:login_method] == 'launchpad'
+      session[:launchpad_cookie]
+    elsif session[:login_method] == 'urs'
+      session[:access_token]
+    end
+  end
+
```
ChristianTrummer99, line 14: the `cmr_client` and `token` methods are only here to allow the loss_report_helper_spec.rb to pass because without these two defined here, the spec file cannot access the definitions of `cmr_client` and `token` elsewhere. Keep in mind, the method definitions (`cmr_client` and `token`) aren't needed when being run in the app (as opposed to being run as a spec).

spec/helpers/loss_report_helper_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'rails_helper'
```
ryanmiller-1, line 1: see above

spec/helpers/loss_report_helper_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+      it 'successfully produces a json loss report' do
+        expect(helper.loss_report_output(echo_id, hide_items=false, disp='json')).to eql(echo_json_report)
+      end
+    end
+    context 'when processing an iso collection' do
+      it 'successfully produces a text loss report' do
+        expect(helper.loss_report_output(iso_id).gsub(/\s+/, "")).to eql(iso_text_report.gsub(/\s+/, ""))
+      end
+      it 'successfully produces a json loss report' do
```
ryanmiller-1, line 40: I think I would prefer this to have a larger fingerprint for the record.

spec/helpers/loss_report_helper_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        expect(helper.loss_report_output(iso_id, hide_items=false, disp='json')).to have_key("31. +: /DS_Series/seriesMetadata/MI_Metadata/dateStamp/Date")
+      end
+    end
+    # the reason the iso example only checks the last key (instead of verifying the full report) is that cmr adds/updates an 'id' attribute
+    # in the actual collection (every time it is translated) and therefore the the comparison report will always include this change
+    # except with a different value for the 'id' attribute. This would cause the equality between the hashes to evaluate false and fail the
+    # test every time. Checking the last change is a comparible solution because it
```
ryanmiller-1, line 47: because it?

spec/support/loss_report_samples_helper.rb
```diff
@@ -0,0 +1,834 @@
+module Helpers
+  module LossReportSamplesHelper
+    def dif_id
```
ryanmiller-1, line 3: I would manage these ids with lets in the test files

app/helpers/loss_report_helper.rb
```diff
@@ -0,0 +1,211 @@
...
+    buckets
+  end
+
+  def prepare_collections(concept_id, umm_c_version)
+    original_collection_native_xml = cmr_client.get_concept(concept_id,token, {})
+    return false if !original_collection_native_xml.success?
+
+    content_type = original_collection_native_xml.headers.fetch('content-type').split(';')[0]
+    original_collection_native_hash = Hash.from_xml(original_collection_native_xml.body)
+    translated_collection_umm_json = cmr_client.translate_collection(original_collection_native_xml.body, content_type, "application/vnd.nasa.cmr.umm+json;version=#{umm_c_version}", skip_validation=true)
```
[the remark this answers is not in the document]
ChristianTrummer99: The `params[format:]` yields something like `text` or `json`.  Is there another `params[<symbol>]` that will yield the info needed?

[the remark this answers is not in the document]
ryanmiller-1: `before_action :set_collection`
means that the set_collection method is being called before each action.  Anything set there should be accessible in the helper when it is called, including @revisions.  I believe that it should be an array of revisions where the first is the most recent, but you should verify that before using it.

app/helpers/loss_report_helper.rb
```diff
@@ -0,0 +1,211 @@
...
+      end
+    end
+    return hash, current_path
+  end
+
+  def array_comparison(path, original_hash, converted_hash)
+    pre_translation_array = hash_navigation(path, original_hash)[0]
+    post_translation_array = hash_navigation(path, converted_hash)[0]
+
+    pre_translation_array == false ? pre_translation_array = Array.new : pre_translation_array = Array.wrap(pre_translation_array)
```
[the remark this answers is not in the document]
ChristianTrummer99: I should probably make a comment to better explain this in the code since the purpose of these conditionals has evolved since the 2311 PR. The current purpose is now to notice a `false` output from `hash_navigation`. This is because `hash_navigation` will return `false` if the path passed to `hash_navigation` doesn't exist. A path may not exist if, for instance, an array that was in the original collection was completely removed during translation and therefore doesn't exist in the translated collection. So in order for the code following the conditionals in question to function properly, array-wrapping a `false` will cause a `false` line item to be output to the loss-report. Therefore, the `pre_translation_array` needs to be redefined as an empty array.

spec/features/collections/loss_report_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
```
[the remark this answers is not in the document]
ChristianTrummer99: They do. Good catch.

spec/features/collections/loss_report_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+describe 'Displaying the comparison report in browser', js: true do
```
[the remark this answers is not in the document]
ChristianTrummer99: they do not.

app/views/collections/loss_report.html.erb
```diff
@@ -0,0 +1,51 @@
+<table id='search-results'>
```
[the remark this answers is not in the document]
ChristianTrummer99: must have forgotten to remove this.

spec/helpers/loss_report_helper_spec.rb
ChristianTrummer99: Yes, will do.

spec/support/loss_report_samples_helper.rb
ChristianTrummer99: I originally defined it using `let`'s but I thought centralizing the definitions would eliminate the possibility that different collections could be tested in the loss_report_helper_spec.rb and the loss_report_spec.rb files.

app/helpers/loss_report_helper.rb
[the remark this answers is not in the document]
ryanmiller-1: Ah, then the last bit can be struck.

app/helpers/loss_report_helper.rb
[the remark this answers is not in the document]
ChristianTrummer99: Ohh, I thought you were talking about the `content_type`, I see you're talking about the whole collection- will fix.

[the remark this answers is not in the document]
ChristianTrummer99: Ah, so it seems that the concept obtained within `set_collection` (the variable is `@collection`) is of the umm-c format so I would still have to `get_concept` in the `prepare_collections` method. Please let me know if I'm misunderstanding something.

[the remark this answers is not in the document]
ryanmiller-1: You need to fetch the native (before) and get the back-translation (after). but you don't need to translate the before to the umm-json because you already have it.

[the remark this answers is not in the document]
ChristianTrummer99: I'm not exactly sure what the difference would be between asking cmr for 1. the UMM version of a collection or 2. the native version of a collection and then using the translate endpoint to convert to UMM. That being said, is it possible that the UMM collection that is retrieved by `set_collections` could be different to the one that is produced by retrieving the native collection and then translating it to UMM? I tested a few collections and haven't found any differences in the loss reports but I imagine it _might_ be possible for those outputs to be different. You may be able to speak more to this..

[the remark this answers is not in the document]
ryanmiller-1: `That being said, is it possible that the UMM collection that is retrieved by set_collections could be different to the one that is produced by retrieving the native collection and then translating it to UMM?`
Erich says that the translation endpoint uses the same code as their internal translation from native to umm-json.  Since the set collections is already being run, you can save on that translation.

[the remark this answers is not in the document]
ChristianTrummer99: perfect. Sounds good.

app/helpers/loss_report_helper.rb
```diff
@@ -0,0 +1,189 @@
...
+
+  def token
+    if session[:login_method] == 'launchpad'
+      session[:launchpad_cookie]
+    elsif session[:login_method] == 'urs'
+      session[:access_token]
+    end
+  end
+
+  def loss_report_output(hide_items=true, disp='text')
```
charleshuang80, line 15: I have sat with this for a while, and the usage of all the instance variables feels wrong to me. I can't cite a specific rails or ruby best practice but here are a few reasons:
- as this is a helper file (and not in the controller), using instance variables directly in the methods makes this confusing. I think normally you would call helper methods in the view file and pass in the instance variables as parameters
- as it is (maybe this will change but based on my understand of how the endpoint will work I don't think it will) this isn't actually being used by a view, so this might be better in a different file, in which case the instance variables might not be available (I am thinking of a concern or service object, but that is not something you need to worry about for this PR)
- along with the previous point of this potentially moving elsewhere in the future, if you want to use these methods in the rake task (which I think we should seriously consider as to not have duplicate code), the instance variables would not work and that information would need to be passed in

Bottom line, I think you should be passing in all the method arguments instead of using instance variables here.

app/helpers/loss_report_helper.rb
```diff
@@ -0,0 +1,189 @@
...
+    # depending on the input selection (json or text) a comparison string/hash is created and displayed in-browser
+
+    # @collection_error is true if there is an error in the translation that is performed by prepare_collections in the collections_controller
```
charleshuang80, line 18: I think you need to add that there error is not something that the method is doing wrong, rather there was something wrong with what CMR gave us back

app/helpers/loss_report_helper.rb
```diff
@@ -0,0 +1,189 @@
...
+    if @content_type.include?('iso') || @content_type.include?('dif')
+      orig = Nokogiri::XML(orig_xml) { |config| config.strict.noblanks } .remove_namespaces!
+      conv = Nokogiri::XML(conv_xml) { |config| config.strict.noblanks } .remove_namespaces!
+    else
+      orig = Nokogiri::XML(orig_xml) { |config| config.strict.noblanks }
+      conv = Nokogiri::XML(conv_xml) { |config| config.strict.noblanks }
+    end
+
+    arr_paths = Array.new # This array is used to keep track of the paths that lead to arrays that have already been mapped
+    text_output = String.new if disp == 'text'
```
charleshuang80, line 40: 1. Can you add a comment here that the text and json outputs seem to be a good candidate for using dependency injection?
2. I think since lines 40-44 are all checking what `disp` is they can be combined.
3. It seems that whatever `disp` is not, that variable will be `nil`, and we will be passing that around the other methods, right? If that's the case I think it would be clearer to just declare it as `nil`
4. This is super minor and you don't have to change it, but I think `display` would be a bit more clear than `disp` and not a lot more characters

app/helpers/loss_report_helper.rb
```diff
@@ -0,0 +1,189 @@
...
+    json_output = Hash.new if disp == 'json'
+
+    json_output['format'] = @content_type if disp == 'json'
+    text_output += (@content_type + "\n\n") if disp == 'text'
+
+    orig.diff(conv, {:added => true, :removed => true}) do |change,node|
```
charleshuang80, line 46: Overall I think that the helper methods have good comments explaining what they are doing. However, this block is fairly confusing so if you can I think it would be helpful to add comments explaining what/how/why you are parsing things

config/routes.rb
```diff
@@ -69,6 +69,7 @@
   get '/collections/:id/download_xml/:format(/:revision_id)' => 'collections#download_xml', as: 'download_collection_xml'
   get '/collections/:id/create_delete_proposal' => 'collections#create_delete_proposal', as: 'create_delete_proposal_collection'
   get '/collections/:id/create_update_proposal' => 'collections#create_update_proposal', as: 'create_update_proposal_collection'
+  get '/collections/:id/loss' => 'collections#loss_report', as: 'loss_report_collections'
```
charleshuang80, line 72: I think we should make the url `loss_report` as well, that will probably be more user friendly/less scary

spec/helpers/loss_report_helper_spec.rb
```diff
@@ -0,0 +1,54 @@
+
+describe 'Loss Report Helper' do
+  let(:umm_c_version) { '1.15.3' }
+
+  context '#prepare_collections' do
```
charleshuang80, line 5: this is no longer in the helper file so shouldn't be here, but that doesn't mean you can't test it in a controller test

spec/support/loss_report_samples_helper.rb
```diff
@@ -0,0 +1,840 @@
...
+      'C1200000031-SEDAC'
+    end
+
+    def iso_id
```
charleshuang80, line 7: 1. As mentioned previously, I think since this is static information, it should be moved to the fixtures folder
2. I think that hard coding in a concept id makes the tests brittle. While it is unlikely that they will change unless our cmr start up script changes, it is still possible. And if that script changes, someone would need to know to make the change here. Sometimes that can't be avoided, but if we can we should try and either try and use a method to grab the concept id or if we are in feature tests, we should do a CMR search and use that information (I am adding a comment in the feature test)
3. if we still need to use hard coded concept ids, I think having a tad more information in the method name would be helpful (e.g., loss_report_iso_id or something with more information)

spec/features/collections/loss_report_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    context 'when displaying json' do
+      it 'properly displays the echo json report' do
+        visit loss_report_collections_path(echo_id, format:'json')
+        expect(page).to have_content('application/echo')
+      end
+      it 'properly displays the iso json report' do
+        visit loss_report_collections_path(iso_id, format:'json')
+        expect(page).to have_content('application/iso')
```
charleshuang80, line 17: So it is good that you have the helper tests to test those methods, but when we are doing feature tests/integration tests, we want to test to make sure that everything is working together. So I think that just checking that it has the right format is not sufficient, you should also test the report result as you are doing in the helper tests.

spec/features/collections/loss_report_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  context 'when accessing the comparison report' do
+
+    before do
+      login
+    end
+
+    context 'when displaying json' do
+      it 'properly displays the echo json report' do
+        visit loss_report_collections_path(echo_id, format:'json')
```
charleshuang80, line 12: As I mentioned in the helper/fixture file, you it is better if we don't use a hard coded concept id. I think for this you might want to do a search for this collection based on short name or something, and then go to the show page. Because there is no button yet, you might want to either grab the concept id from the url (I can provide an example if you need) or grab the url to construct the path you need. This may mean that it makes more sense to restructure this file a bit (i.e., grouped by collection native format vs output format).

app/helpers/loss_report_helper.rb
ChristianTrummer99: Just made changes 1,2, and 4. I'm not entirely sure what you mean for #3. If you could expand a little more, that'd be great.

app/helpers/loss_report_helper.rb
```diff
@@ -1,93 +1,89 @@
...
+    return json_output if display == 'json'
   end
 
-  def is_xml?(node)
-    # checks if the node being passed is xml
+  def is_xml?(element)
+    # checks if the element being passed is xml
     # may be beneficial to add more checks
-    node.to_xml.include?('<' && '</' && '>') ? true : false
+    element.include?('<' && '</' && '>') ? true : false
```
ryanmiller-1, line 78: The ternary is redundant if the first clause returns true or false.

app/controllers/collections_controller.rb
```diff
@@ -127,6 +142,25 @@ def ensure_correct_collection_provider
     render :show
   end
 
+  def prepare_translated_collections
+    original_collection_native_xml = cmr_client.get_concept(params[:id],token, {})
+    return false if !original_collection_native_xml.success?
```
charleshuang80, line 147: In the case of these errors, instead of returning `false`, I think it might be better if you returned something like `{  error: 'this thing failed' }`. That way you can check for something more specific (an error key), you can provide a more specific error message since there are 3 places you do this, and it is a little more similar to how we deal with cmr responses (not necessarily the best reason).

app/controllers/collections_controller.rb
```diff
@@ -115,6 +117,19 @@ def create_update_proposal
...
 
+  def loss_report
+    # When a user wants to use MMT to edit metadata that currently exists in a non-UMM form,
+    # it's important that they're able to see if any data loss occurs in the translation to umm.
+    # This method is needed to reference the appropriate helper and view for the lossiness report.
+    # If compared_collections is false, the error message will appear. Note that compared_collections
+    # is false when the cmr calls aren't successful.
+    compared_collections = prepare_translated_collections
+    respond_to do |format|
+      format.text { render plain: JSON.pretty_generate(@collection) + (compared_collections ? loss_report_output(compared_collections, hide_items: true, display: 'text') : 'Failure to get_concept or translate_collection' )}
```
charleshuang80, line 128: I think that these blocks are quite long one liners, and they are important/complicated enough that it is worth splitting them out into `if/else` statements

app/helpers/loss_report_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+    true
+  end
+
+  def top_level_arr_path(path, orig_h, conv_h)
+    # if an array is passed that passes through an array ie. /Contacts/Contact[0]/Role/Name
+    # this method would return /Contacts/Contact because Contact is the outermost array (or false if the path doesn't contain an array)
+    pre_translation_array, pre_translation_path = hash_navigation(path, orig_h)
+    post_translation_array, post_translation_path = hash_navigation(path, conv_h)
+
+    return false if pre_translation_array == false && post_translation_array == false
```
charleshuang80, line 106: we talked about this, might be worth a comment explaining what scenario this is informing us about

app/helpers/loss_report_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+    else
+      orig = Nokogiri::XML(compared_collections[:original_collection_native_xml]) { |config| config.strict.noblanks }
+      conv = Nokogiri::XML(compared_collections[:translated_collection_native_xml]) { |config| config.strict.noblanks }
+    end
+
+    arr_paths = Array.new # This array is used to keep track of the paths that lead to arrays that have already been mapped
+
+    if display == 'text'
+      text_output = String.new
+      text_output += (compared_collections[:native_format] + "\n\n")
```
charleshuang80, line 26: To your question, I meant that in this case you are declaring what `text_output` is, but in subsequent method calls, you are passing both this variable as well as `json_output` which will be nil. So I think it would be helpful to just be explicit and also declare `json_output = nil` here. And vice versa in the elsif below.

ChristianTrummer99: Ah, I see. will do

app/controllers/collections_controller.rb
ChristianTrummer99: Great point. This makes the error handling much cleaner too

app/helpers/loss_report_helper.rb
charleshuang80: I missed the added declarations on my first pass. I think that's why we avoid multiple assignment even though ruby allows it

app/helpers/loss_report_helper.rb
```diff
@@ -0,0 +1,189 @@
+module LossReportHelper
+
+  def loss_report_output(compared_collections: compared_collections, hide_items: true, display: 'text')
```
charleshuang80, line 3: Sorry if I was unclear. You are not providing a default value here for compared_collections (I think the "default value" here actually doesn't mean anything and if you tried calling the method without passing in any arguments it would throw an error), so this should look like `def loss_report_output(compared_collections:, hide_items: true, display: 'text')`

ChristianTrummer99: okay I see, Is there any benefit to using symbols for plain arguments?

app/helpers/loss_report_helper.rb
ChristianTrummer99: Okay just changed it.

app/helpers/loss_report_helper.rb
ryanmiller-1: Consistency with your other arguments and argument order agnosticicity (agnosticness?)

app/controllers/collections_controller.rb
```diff
@@ -129,7 +129,7 @@ def loss_report
         format.json { render json: JSON.pretty_generate(compared_collections) }
       else
         format.text { render plain: loss_report_output(compared_collections: compared_collections, hide_items: true, display: 'text') }
-        format.json { render json: JSON.pretty_generate(loss_report_output(compared_collections: compared_collections, hide_items: false, display: 'json')) }
```
ryanmiller-1: Doesn't removing this mean that all paths are using the default behavior of true? If so, why have the option?

ChristianTrummer99: The option is only for `text_output`, it doesn't affect `json_output`

ChristianTrummer99: I suppose then, to answer your question, the purpose of having the option (which only affects `text_output`) is for debugging if a dev is using the `text_output` option.

ryanmiller-1: I think I would prefer it to be pulled out of the params with a default value then.  E.g. url/loss_report?hide_items=false should work

app/controllers/collections_controller.rb
```diff
@@ -156,12 +156,25 @@ def prepare_translated_collections
     translated_collection_native_xml = cmr_client.translate_collection(JSON.pretty_generate(@collection), "application/#{Rails.configuration.umm_c_version}; charset=utf-8", content_type,  skip_validation=true)
     return { error: 'Failed to translate collection from UMM back to native format' } unless translated_collection_native_xml.success?
 
+    # this checks the 'hide_items' url parameter that is can be manually added. Its primary use is for developers
```
ryanmiller-1, line 159: If you're going to provide an error for an invalid parameter, I think this should happen before you contact CMR.  Additionally, I don't think this belongs in this function.  

You can do something like 
```
compared collections = if hide_items.is_a?(Hash)
  hide_items
else
  prepare_translated_collections
end
```
compared_collections isn't accurate is it?  They haven't been compared yet?

PR #625, opened 2020-08-12
MMT-2346: Notify user of potential data loss when attempting to edit a non-UMM collection

ChristianTrummer99:
- Added extra logic to the collections show.html.erb page that pops up the loss_report_modal
- created _loss_report_modal.html.erb that offers users the option to view a loss report before editing a non-UMM collection
- because the _not_current_provider_modal.html.erb will be displayed if a user tries to edit a collection that requires a provider context that is not present, coffeescript was added to change_current_provider.coffee
- wrote tests to ensure the modals display correctly


app/views/shared/_loss_report_modal.html.erb
```diff
@@ -0,0 +1,19 @@
+<div id="loss-report-modal" class="eui-modal-content" >
```
ChristianTrummer99, line 1: I was having trouble with scss to get the margins looking right on the buttons so I put the `margin-right` value directly on the rails tag. See the comment on the scss file below..

app/views/shared/_loss_report_modal.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  <a href="javascript:void(0);" class="modal-close float-r" id="close-loss-report-modal"><i class="fa fa-times"></i><span class="is-invisible">Close</span></a>
+  <div>
+    <h4>Conversion to UMM Format Required</h4>
+    <p>The native format of this collection is <%= options[:native_format].split('/').last.gsub('+xml','').upcase %>. Editing this record using the MMT will convert it to UMM-JSON, which may result in data loss.</p>
+    <p style="display: inline-flex">
```
ChristianTrummer99, line 6: for some reason one or two buttons were not aligning horizontally and this `display: inline-flex` solved that issue. Let me know if there's a better way to do this.

app/assets/stylesheets/components/_buttons.scss
```diff
@@ -26,7 +26,19 @@
   &:hover {
     background-color: darken($red, 6%);
   }
-}  
+}
+
+.eui-btn--light-blue {
```
ChristianTrummer99, line 31: Alicia suggested I make the "View Loss Report" button a different blue color so I made this new button style. Just want to make sure adding this is okay

app/assets/stylesheets/components/_buttons.scss
```diff
@@ -26,7 +26,19 @@
   &:hover {
     background-color: darken($red, 6%);
   }
-}  
+}
+
```
ChristianTrummer99, line 30: Alicia suggested I use a different blue for the "View Loss Report" button so I added this button style, let me know if I should do this differently

app/views/shared/_loss_report_modal.html.erb
ryanmiller-1: For both of these issues, attach the css to a class and the class to the html.

spec/features/collections/loss_report_modal_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+          expect(page).to have_content('The native format of this collection is ECHO10. Editing this record using the MMT will convert it to UMM-JSON, which may result in data loss.')
+          expect(page).to have_link('View Loss Report', href: loss_report_collections_path(concept_id, format: 'json'))
+          expect(page).to have_link('Edit Collection', href: edit_collection_path(id: concept_id))
+          expect(page).to have_link('Cancel', href: 'javascript:void(0);')
+        end
+      end
+
+      context 'when the "Edit Collection" button is clicked' do
+        it 'does not display the not-current-provider-modal' do
+          within '#loss-report-modal' do
```
ryanmiller-1, line 27: Move 27-29 to a before

spec/features/collections/loss_report_modal_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+            click_on 'Edit Collection'
+          end
+
+          expect(page).to have_no_css('#not-current-provider-modal')
```
ryanmiller-1, line 31: Include an expectation about what you do expect to see

spec/features/collections/loss_report_modal_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+        end
+      end
+
+      context 'when the "View Loss Report" button is clicked' do
+        it 'does not display display the not-current-provider-modal and does not close the loss-report-modal' do
```
ryanmiller-1, line 36: extra displays in it text (and below)

spec/features/collections/loss_report_modal_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+      end
+
+      context 'when the "Cancel" button is clicked' do
+        it 'does not display display the not-current-provider-modal and closes loss-report-modal' do
+          within '#loss-report-modal' do
+            click_on 'Cancel'
+          end
+
+          expect(page).to have_no_css('#not-current-provider-modal')
+          expect(page).to have_no_css('#loss-report-modal')
```
ryanmiller-1, line 53: Verify no page transition

spec/features/collections/loss_report_modal_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+describe 'loss report modal', js: true do
+  # this is an echo collection (SEDAC provider)
+  let(:cmr_response) { cmr_client.get_collections({'EntryTitle': 'Anthropogenic Biomes of the World, Version 2: 1700'}) }
+  let(:concept_id) { cmr_response.body.dig('items',0,'meta','concept-id') }
+
+  context 'when user clicks Edit Collection Record for a non-UMM collection' do
+    context 'when provider context does not need to be changed' do
+
+      before do
+        login(provider: 'SEDAC', providers: %w[SEDAC LARC])
```
ryanmiller-1, line 11: Don't need the larg provider for these tests

spec/features/collections/loss_report_modal_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+        within '#loss-report-modal' do
+          expect(page).to have_content('The native format of this collection is ECHO10. Editing this record using the MMT will convert it to UMM-JSON, which may result in data loss.')
+          expect(page).to have_link('View Loss Report', href: loss_report_collections_path(concept_id, format: 'json'))
+          expect(page).to have_link('Edit Collection', href: '#')
+          expect(page).to have_link('Cancel', href: 'javascript:void(0);')
+        end
+      end
+
+      context 'when the "Edit Collection" button is clicked' do
+        it 'displays the not-current-provider-modal' do
```
ryanmiller-1, line 76: Comment above applies to these tests; move the clicking/actions into before blocks, as much as possible.

spec/features/collections/loss_report_modal_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+            click_on 'View Loss Report'
+          end
+
+          expect(page).to have_css('#loss-report-modal')
+          expect(page).to have_no_css('#not-current-provider-modal')
+        end
+      end
+
+      context 'when the "Cancel" button is clicked' do
+        it 'does not display display the not-current-provider-modal and closes loss-report-modal' do
```
ryanmiller-1, line 102: more it text duplicate 'display'

app/assets/javascripts/change_current_provider.coffee
```diff
@@ -106,7 +113,7 @@ $(document).ready ->
         # Click the link that the user needs
         if linkType == 'delete-collection' && $('.collection-granule-count').text() != 'Granules (0)'
           $('#display-granules-modal').click()
-        else 
+        else
```
jdf-eos, line 116: Fixing these trailing blanks makes the PR harder to review. I'm not sure that the benefit justifies the confusion.  Just a thought.

app/views/shared/_loss_report_modal.html.erb
charleshuang80: I think we talked about this modal and modifying some of the classes, so this style is no longer necessary, but leaving the above comment as it is something to keep in mind.

app/views/shared/_loss_report_modal.html.erb
```diff
@@ -0,0 +1,19 @@
...
+      <% if current_provider?(options[:provider_id]) %>
+        <%= link_to 'View Loss Report', loss_report_collections_path(options[:concept_id], format: 'json'), class: 'button-spacing eui-btn--steel-blue', target: :_blank %>
+        <%= link_to 'Edit Collection', edit_collection_path(revision_id: options[:revision_id]), class: 'eui-btn--blue modal-close-link button-spacing' %>
+        <%= link_to 'Cancel', 'javascript:void(0);', class: 'eui-btn modal-close-link' %>
+      <% elsif available_provider?(options[:provider_id]) %>
+        <%= link_to 'View Loss Report', loss_report_collections_path(options[:concept_id], format: 'json'), class: 'button-spacing eui-btn--steel-blue', target: :_blank %>
+        <%= link_to 'Edit Collection', '#', class: 'eui-btn--blue loss-report button-spacing' %>
+        <%= link_to 'Cancel', 'javascript:void(0);', class: 'eui-btn modal-close-link' %>
+      <% end %>
+      <%= link_to 'Invisible Link', '#not-current-provider-modal', class: 'display-modal not-current-provider eui-btn is-invisible', id: 'loss-report-to-edit-collection', data: { provider: options[:provider_id], record_action: 'edit-collection' } %>
```
charleshuang80, line 16: I think we figured out how to not need this

app/views/shared/_loss_report_modal.html.erb
```diff
@@ -0,0 +1,19 @@
+<div id="loss-report-modal" class="eui-modal-content" >
+  <a href="javascript:void(0);" class="modal-close float-r" id="close-loss-report-modal"><i class="fa fa-times"></i><span class="is-invisible">Close</span></a>
+  <div>
+    <h4>Conversion to UMM Format Required</h4>
+    <p>The native format of this collection is <%= options[:native_format].split('/').last.gsub('+xml','').upcase %>. Editing this record using the MMT will convert it to UMM-JSON, which may result in data loss. Select 'View Loss Report' to see how the conversion will affect this record.</p>
+    <p style="display: inline-flex">
+      <% if current_provider?(options[:provider_id]) %>
```
charleshuang80, line 7: The view loss report and cancel buttons are the same in both of these scenarios and so are repeated.
1. Can the order of the buttons be changed, or is this the order the PO wanted? If it can be changed it would be easier to 'dry' this up.
2. If the order is fixed, I'm debating if it makes more sense to do something like
```
if current_provider || available_provider
  loss report button
  if current_provider
    edit button
  else
    edit button
  end
  cancel button
end
```
That doesn't seem too ideal either. Other opinions would be helpful here.

app/assets/stylesheets/components/_modal.scss
```diff
@@ -15,6 +15,7 @@
   display: none;
 }
 
+#loss-report-modal,
```
charleshuang80, line 18: I think we decided this wasn't necessary

app/assets/javascripts/change_current_provider.coffee
```diff
@@ -4,6 +4,13 @@ $(document).ready ->
     overlay: 0.6
     closeButton: '.modal-close'
 
+  $('a.modal-close-link').on 'click', (element) ->
```
ryanmiller-1, line 7: This is necessary because using modal-close causes the button to float right because of the CSS on lines 33-36 of _modal.scss  I think I would favor a solution where you overwrite the CSS and use modal-close rather than this JS.

app/views/shared/_loss_report_modal.html.erb
ryanmiller-1: Depending on how you solve the modal-close comment above, this could easily be added to that.  Either way, I think you should remove the inline styling.

app/views/shared/_loss_report_modal.html.erb
ryanmiller-1: The modal is not visible unless `current_provider || available_provider` is true.  I think it can just be:
```loss report button
  if current_provider
    edit button
  else
    edit button
  end
  cancel button```

File: app/views/shared/_loss_report_modal.html.erb (hunk 13 above)
Comment (ChristianTrummer99) [reply app/views/shared/_loss_report_modal.html.erb:16]: This instance is different than the `modal-close-link`. this causes the first modal to close and the next to open. reference coffeescript

Comment (ryanmiller-1) [reply app/views/shared/_loss_report_modal.html.erb]: Couldn't the wrong provider modal move into the available_provider edit collection link?

File: app/views/shared/_loss_report_modal.html.erb [hunk 17]
```diff
@@ -0,0 +1,19 @@
...
+      <%= link_to 'View Loss Report', loss_report_collections_path(options[:concept_id], format: 'json'), class: 'eui-btn eui-btn--steel-blue', target: :_blank %>
+
+      <% if current_provider?(options[:provider_id]) %>
+        <%= link_to 'Edit Collection', edit_collection_path(revision_id: options[:revision_id]), class: 'eui-btn--blue modal-close button-spacing' %>
+      <% elsif available_provider?(options[:provider_id]) %>
+        <%= link_to 'Edit Collection', '#', class: 'eui-btn--blue loss-report button-spacing' %>
+      <% end %>
+
+      <%= link_to 'Cancel', 'javascript:void(0);', class: 'eui-btn modal-close' %>
+      <%= link_to 'Invisible Link', '#not-current-provider-modal', class: 'display-modal not-current-provider is-invisible', id: 'loss-report-to-edit-collection', data: { provider: options[:provider_id], record_action: 'edit-collection' } %>
```
charleshuang80, line 16: Is this invisible link necessary? Could you add all this information to the Edit link directly?

app/views/collections/show.html.erb
```diff
@@ -57,10 +57,15 @@
...
-            <% elsif available_provider?(@provider_id) %>
-              <%= link_to 'Edit Collection Record', '#not-current-provider-modal', class: 'display-modal not-current-provider eui-btn--link bar-after', data: { provider: @provider_id, record_action: 'edit-collection' } %>
+            <% native_format = @revisions.first.fetch('meta', {})['format'] %>
+            <% if !native_format.include?('umm') && (current_provider?(@provider_id) || available_provider?(@provider_id)) %>
+              <%= link_to 'Edit Collection Record', '#loss-report-modal', class: 'display-modal eui-btn--link bar-after' %>
+            <% else %>
+              <% if current_provider?(@provider_id) %>
+                <%= link_to 'Edit Collection Record', edit_collection_path(revision_id: @revision_id), class: 'eui-btn--link bar-after' %>
+              <% elsif available_provider?(@provider_id) %>
+                <%= link_to 'Edit Collection Record', '#not-current-provider-modal', class: 'display-modal not-current-provider eui-btn--link bar-after', data: { provider: @provider_id, record_action: 'edit-collection' } %>
```
charleshuang80, line 67: could the information given to this link be added to the corresponding link in the loss report modal?

app/views/shared/_loss_report_modal.html.erb
ChristianTrummer99: That was my first thought, but closing one modal (the `loss-report-modal`) then opening another (the `not-current-provider-modal`) requires coffeescript as far as I can tell

app/views/shared/_loss_report_modal.html.erb
ChristianTrummer99: There were 2 things I considered when I was making this invisible link
1. closing one modal (the `loss-report-modal`) then opening another (the `not-current-provider-modal`) requires coffeescript so unfortunately I can just link directly to the next modal. There has to be a "close modal" action and then an "open modal" action in between. On top of that, the `not-current-provider-modal` is a complex modal to open (reference change_current_provider.coffee)
2. That being said, the `not-current-provider-modal` does a bit of event handling on its own, so I decided against copying all the code that is used to produce the `not-current-provider-modal` and pasting it into the handler that handles clicks on the `loss-report` class. From what I could tell the next easiest option was to have a hidden link (similar to the _not_current_provider_modal.html.erb partial)

ChristianTrummer99: Also simply adding the classes from 'invisible link' to the Edit link unfortunately doesn't work

app/views/shared/_loss_report_modal.html.erb
ChristianTrummer99: Also simply adding the classes from 'invisible link' to the Edit link unfortunately doesn't work

app/views/collections/show.html.erb
ChristianTrummer99: I would say yes if the loss-report-modal didn't have to close first but that seems to add complexity unfortunately

app/views/shared/_loss_report_modal.html.erb
ryanmiller-1: `    <%= link_to 'Edit Collection', '#not-current-provider-modal', class: 'eui-btn--blue loss-report button-spacing display-modal not-current-provider', data: { provider: options[:provider_id], record_action: 'edit-collection' } %>
    `
and removing the invisible link seems to work for me.  Should check if all of that is necessary since my check was really coarse.  Also lets you remove the click from the coffee, the hide is sufficient.

ChristianTrummer99: I stand corrected, I did not properly add the classes from the invisible link to the Edit link. Thank you both for this suggestion

app/views/shared/_loss_report_modal.html.erb
ChristianTrummer99: I stand corrected, I did not properly add the classes from the invisible link to the Edit link. Thank you both for this suggestion

app/views/collections/show.html.erb
ChristianTrummer99: I stand corrected, I did not properly add the classes from the invisible link to the Edit link. Thank you both for this suggestion

ChristianTrummer99:
Thank you both for helping me clean up the modal close/open action. If you have any suggestions for the css let me know. I think I implemented 80% of the things Charles and I talked about for the css. I couldn't figure out how to make a true `eui-btn--steel-blue` with the inherent eui properties so I left it the way it is now.

PR #626, opened 2020-08-12
MMT-2335 Rejection Feedback fields in dMMT can be submitted with no feedback

ryanmiller-1:
Updated form and validation to display required icons and error messages only if at least one field has data.
Updated controller to handle empty feedback from form.
Updated mailer to handle empty feedback scenario
Changed logging for string bodies.
Changed URS requests to use post bodies instead of url parameters


app/assets/javascripts/proposals.coffee
```diff
@@ -9,6 +9,17 @@ $(document).ready ->
       # validate or revalidate on close
       $(this).valid()
 
+    # The jquery validation below is not sufficient to drive the error displays
+    # so changing the note should trigger a class change on the reasons select2
+    $('#rejection_note').change ->
+      if $('#rejection_note').val()
```
charleshuang80, line 15: so calling `updateRequiredIcons()` is not sufficient, you need to add or remove the class on the other field for this to work?

app/mailers/proposal_mailer.rb
```diff
@@ -1,4 +1,5 @@
 class ProposalMailer < ApplicationMailer
+  DMMT_USER_GUIDE_URL = "https://wiki.earthdata.nasa.gov/display/ED/Draft+MMT+%28dMMT%29+User%27s+Guide"
   def proposal_submitted_notification(user, short_name, version, id, request_type)
```
charleshuang80, line 3: nitpick - think there should be a space after the constant (don't change unless you have to make other changes)

app/assets/javascripts/proposals.coffee
```diff
@@ -21,11 +32,25 @@ $(document).ready ->
...
+          required: 
+            depends:
+              -> $('#proposal-rejection-reasons').val().length > 0
       messages:
         'rejection[reasons][]':
           required: 'Reason(s) are required'
         'rejection[note]':
           required: 'Note is required'
+          
+    updateRequiredIcons = ->
```
charleshuang80, line 48: No need to do anything now, but just thinking that it seems like we do this sort of thing quite a bit with many of the other forms, and we might want to try and make a generic method for simpler fields like these.

app/views/proposal_mailer/proposal_rejected_notification.html.erb
```diff
@@ -6,10 +6,14 @@
...
   <p>
-    <%= "The collection draft proposal #{@short_name}_#{@version} has been reviewed and rejected for the following reasons: '#{@rejected_reasons}'." %>
+    <%= "The collection draft proposal #{@short_name}_#{@version} has been reviewed and rejected#{@reasons_text}." %>
   </p>
   <p>
-    <%= simple_format("Additional feedback was provided: '#{@rejected_note}'") %>
+    <% if @rejected_note %>
+      <%= simple_format("Additional feedback was provided: '#{@rejected_note}'") %>
+    <% else %>
+      <%= "Please consult the " %><%= link_to("dMMT User's Guide", @dmmt_user_guide_url) %><%= " for information on CMR metadata best practices." %>
```
charleshuang80, line 15: Did you check if this should be added to the email in all cases?

app/assets/javascripts/proposals.coffee
ryanmiller-1: Correct.

PR #628, opened 2020-08-17
MMT-1876: Link SKW, Data Centers, instruments, and platforms in a collection's preview to Earthdata Search collections with the same keywords

ChristianTrummer99:
- added an `edsc_url` method to read the current MMT environment (sit,prod,uat,etc) and return the proper edsc url for passage to the preview gem

### DIFF
--- a/app/helpers/collections_helper.rb
+++ b/app/helpers/collections_helper.rb
@@ -41,4 +41,14 @@ module CollectionsHelper
       content_tag(:span, 'NRT', class: 'eui-badge nrt')
     end
   end
+
+  def edsc_url
+    if Rails.env.production?
+      'https://search.earthdata.nasa.gov/search'
+    elsif Rails.env.sit?
+      'https://search.sit.earthdata.nasa.gov/search'
+    elsif Rails.env.uat?
+      'https://search.uat.earthdata.nasa.gov/search'
+    end
+  end
 end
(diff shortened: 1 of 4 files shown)


app/helpers/collections_helper.rb
```diff
@@ -41,4 +41,19 @@ def nrt_badge(metadata)
       content_tag(:span, 'NRT', class: 'eui-badge nrt')
     end
   end
+
+  def edsc_url
+    config_services = Rails.configuration.services
+    if Rails.env.production?
+      config_services.dig('earthdata', 'ops', 'edsc_root')
+    elsif Rails.env.sit? || Rails.env.development? || Rails.env.test?
```
jdf-eos, line 49: Is it possible to run a local EDSC?

ChristianTrummer99: I dont think so (just trying https://search.localtest.earthdata.nasa.gov/search doesn't work) but @ryanmiller-1 @charleshuang80 may have insight.

ryanmiller-1: EDSC has instructions in their repo to run locally.  I don't think that this work warrants the extra complication to development and testing.

william-valencia: If you edit your local services.yml, you can just point that location to a local EDSC I believe.

PR #629, opened 2020-08-18
MMT-2069 Delete all proposals > 30days old and delete proposals after publishing.

ryanmiller-1:


lib/tasks/proposals.rake
```diff
@@ -2,8 +2,8 @@ namespace :delete_proposals do
   desc 'Delete proposals that are older than specified date old'
   task expired_proposals: :environment do
     if Rails.configuration.proposal_mode
-      count = CollectionDraftProposal.delete_all("proposal_status = 'in_work' AND updated_at < '#{30.days.ago}'")
-      Rails.logger.info("Automated check for proposals that are 'in_work' and 30 days old deleted #{count || 0} proposals.")
+      count = CollectionDraftProposal.where("updated_at < '#{30.days.ago}'").delete_all
```
william-valencia, line 5: Would 30 days ever change? i.e. should we make it configurable?

ryanmiller-1: I doubt it, but I have added it as a configuration, so that we can adjust it if we need to.

PR #630, opened 2020-08-18
MMT-1782: Display tags for collections on search results and collections show page

charleshuang80:
* enable retrieving tag information
* add a tags controller for search
* add display of tag information on collection search results page
* add display of tag information on collection show page
* add tests


app/assets/javascripts/collections.coffee
```diff
@@ -5,3 +5,101 @@ $(document).ready ->
...
+  $('.col-tag-key-link').on 'click', (e) ->
+    # link should only be present if there are tags
+
+    # grab tag keys
+    fieldId = $(this).data('colTagKeys') + '_' # id needs the trailing '_' b/c it is an array field
+    tagKeys = $("##{fieldId}").val().split(' ')
+
+    $.ajax
+      url: "/tags"
+      data: {'tag_key': tagKeys, 'gibberish': 'blarg'}
```
charleshuang80, line 62: clean up params

spec/support/tags_helpers.rb
```diff
@@ -0,0 +1,48 @@
+module Helpers
+  module TagsHelpers
+    def setup_tag_permissions
+      # this method is necessary because the local cmr system token is not
+      # automatically provided the system Tags ACL
+      # this can be removed after CMR-6654 is worked
```
ryanmiller-1, line 6: Update with our ticket number MMT-2360

app/assets/javascripts/tags.coffee
```diff
@@ -0,0 +1,92 @@
+$(document).ready ->
```
ryanmiller-1, line 1: Since the process for getting tags for the show page and search page are different, I think adding a couple high level overview comments could help guide future developers to the right places.

spec/features/collections/collection_with_tags_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+    associate_tag_to_collection_by_short_name(tag_2_key, short_name)
+  end
+
+  context 'when viewing a collection with tags' do
+    before do
+      login
+    end
+
+    context 'when retrieving all tag information succeeds' do
+      before do
```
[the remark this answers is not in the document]
charleshuang80: I'm not sure I follow in this case. We are checking something and then clicking on a link for the modal and then checking the modal content.
We can discuss further if you like, but I think many of our actions (like login) have to be done in a before each block, not a before all - I think it has to do with mocking some requests. If requests don't have to be in a before each, then I think it is fine to do as much as we can in a before all block and then test for the things but I'm not following how we should do that in this file.

spec/features/search/collection_search_result_spec.rb
```diff
@@ -139,6 +152,116 @@
     end
   end
 
+  context 'when searching by short name for a collection which has a tag count' do
+    search_tag_1_key = 'tag.search.example.01'
+    search_tag_1_description = 'This is a search example tag'
+    tag_collection_short_name = 'Collection Tagging search example 01'
+
```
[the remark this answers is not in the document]
charleshuang80: I could but I'm not sure if that is more helpful. In one case the tag count is wrong and so would be different, and in the other 2 we are actually checking for a link so it is more than just the tag count value.

lib/cmr/cmr_client.rb
```diff
@@ -183,6 +183,62 @@ def get_controlled_keywords(type)
...
+    def create_tag(tag_key, token, description = nil)
+      url = 'http://localhost:3003/tags'
+      body = { tag_key: tag_key }
+      body[:description] = description unless description.nil?
+      headers = { 'Content-Type' => 'application/json' }
+      # need to use 'access_token_admin'
+      post(url, body.to_json, headers.merge(token_header(token)))
+    end
+
+    def associate_tag_by_collection_concept_id(tag_key, collection_id, token)
```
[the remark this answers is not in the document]
charleshuang80: I was able to make it work, but it is not being used, I think using the query method is a bit more versatile. I will take it out, but can put it back in if/when we want to use it.

spec/features/collections/collection_with_tags_spec.rb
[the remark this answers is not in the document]
ryanmiller-1: No, I think you're right about login.

app/concerns/cmr_collections_helper.rb
```diff
@@ -19,7 +19,11 @@ def get_revisions(concept_id, revision_id)
       break if latest && latest['meta']['revision-id'] >= revision_id.to_i
       attempts += 1
       sleep 0.05
+      Rails.logger.info("Retrieving a collection record in `get_revisions`. Need to loop, about to try attempt number #{attempts}")
```
ChristianTrummer99, line 22: This is very minor semantics but when I read this I thought it might sound better as "make attempt" rather than "try attempt"

app/helpers/search_helper.rb
```diff
@@ -64,6 +64,22 @@ def display_last_modified(record)
     end
   end
 
+  def display_tag_count(record)
+    tag_keys = record.fetch('added_fields', {}).fetch('tags', {})&.keys
+    tag_count = tag_keys.nil? ? 0 : tag_keys.count
```
ChristianTrummer99, line 69: Since `tag_keys` will be `nil` if `&.keys` is not valid and the optional return parameter `{}` passed to `fetch` will not be used I think it would be cleaner to use `#dig`. This could be used `tag_keys = record.dig('added_fields', 'tags')&.keys` since `#dig` returns `nil` when one of the passed keys is not found

PR #633, opened 2020-08-26
MMT-1876: Revert "MMT-1876: Link SKW, Data Centers, instruments, and platforms …

charleshuang80:
…in a collection's preview to Earthdata Search collections with the same keywords (#628)"

This reverts commit 07eb14a393df89a54a74f10eff066fbb0d514ea5.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -78,7 +78,7 @@ gem 'browser'
 # bundle config local.cmr_metadata_preview /path/to/local/git/repository
 # make sure to delete the local config when done making changes to merge into master
 # bundle config --delete local.cmr_metadata_preview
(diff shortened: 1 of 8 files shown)


ryanmiller-1:
If the original PR broke a test that was fixed in another branch, isn't reverting it going to need to revert that test too?

charleshuang80:
> If the original PR broke a test that was fixed in another branch, isn't reverting it going to need to revert that test too?

I don't think so. I ran that set of tests before pushing up and didn't run into issues. Also, the only thing I did to fix the test was to add a `normalize_ws` option to the assertion.
In Bamboo looks like the test being discussed passed but I had an intermittent failure so running it again.

PR #634, opened 2020-08-27
hotfix/MMT-2313-1: removed xml comments and whitespace from collection before generating loss report

ChristianTrummer99:
- Francell uncovered some errors that occurred when trying to generate a loss report when there is irregular whitespace and comments in the xml collection. 
- updated the compare_xml_collections.rake and loss_report_helper.rb files to avoid these errors
- added tests to make sure these kinds of collections don't cause errors with loss report generation


lib/tasks/compare_xml_collections.rake
```diff
@@ -6,16 +6,18 @@ namespace :collection do
     args.with_defaults(:version => '1.15.3')
     args.with_defaults(:disp => 'show')
 
-    abort 'FORMAT INVALID' unless args.format == 'echo10' || args.format == 'dif10' || args.format == 'iso19115'
+    abort 'FORMAT INVALID' unless args.format == 'echo10' || args.format == 'dif10' || args.format == 'iso19115' || args.format == 'iso:smap'
```
ryanmiller-1, line 9: I think I prefer 'NOT SUPPORTED' here.

app/controllers/collections_controller.rb
```diff
@@ -158,20 +158,36 @@ def ensure_correct_collection_provider
...
+    original_collection_native_xml.xpath('//comment()').remove
 
     return {
-      original_collection_native_xml: original_collection_native_xml.body,
-      translated_collection_native_xml: translated_collection_native_xml.body,
-      original_collection_native_hash: Hash.from_xml(original_collection_native_xml.body),
-      translated_collection_native_hash: Hash.from_xml(translated_collection_native_xml.body),
+      original_collection_native_xml: original_collection_native_xml,
+      translated_collection_native_xml: translated_collection_native_xml,
+      original_collection_native_hash: Hash.from_xml(original_collection_native_xml.to_xml),
```
ryanmiller-1, line 189: If these need to be `to_xml`ed, should they be named something else?

ChristianTrummer99: I would say yes, but in this case the `*_collection_native_xml` variables are [nokogiri XML document objects](https://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri/XML/Document) so the variable name isnt a total misnomer. The method `.to_xml` just converts the nokogiri xml object to an xml string. I'll add `_noko_obj` to the end of the `*_collection_native_xml`

app/controllers/collections_controller.rb
```diff
@@ -158,20 +158,36 @@ def ensure_correct_collection_provider
...
+    return { error: 'Failed to retrieve collection from CMR' } unless cmr_response_original.success?
 
-    content_type = original_collection_native_xml.headers.fetch('content-type').split(';')[0]
+    content_type = cmr_response_original.headers.fetch('content-type').split(';')[0]
     return { error: 'This collection is already in UMM format so there is no loss report' } if content_type.include?('application/vnd.nasa.cmr.umm+json')
 
-    translated_collection_native_xml = cmr_client.translate_collection(JSON.pretty_generate(@collection), "application/#{Rails.configuration.umm_c_version}; charset=utf-8", content_type,  skip_validation=true)
-    return { error: 'Failed to translate collection from UMM back to native format' } unless translated_collection_native_xml.success?
+    cmr_response_translated = cmr_client.translate_collection(JSON.pretty_generate(@collection), "application/#{Rails.configuration.umm_c_version}; charset=utf-8", content_type,  skip_validation=true)
+    return { error: 'Failed to translate collection from UMM back to native format' } unless cmr_response_translated.success?
```
ryanmiller-1, line 168: I think we should log the collections which cause these errors.  It will make it easier to debug in the future.  See the logging for a failed publish attempt (collection draft controller) for guidance.

app/controllers/collections_controller.rb
```diff
@@ -158,20 +158,49 @@ def ensure_correct_collection_provider
...
+      Rails.logger.info("User #{current_user.urs_uid} attempted to translate collection #{@collection['EntryTitle'] || params[:id]} (with provider #{current_user.provider_id}) from #{Rails.configuration.umm_c_version} to #{content_type} without success.")
+      return { error: "Failed to translate collection from UMM JSON back to #{content_type}" }
+    end
+
+
+    # ISO and DIF collections (in XML form) contain namespaces that cause errors when passed to loss_report_helper.rb.
+    # Specifically, when nodes are evaluated individually, (their namespace definitions remaining at the top of the xml)
+    # their prefixes are undefined in the scope of the evaluation and therefore raise errors. Removing the namespaces
+    # eliminates this issue.
+    if content_type.include?('iso') || content_type.include?('dif')
```
charleshuang80, line 187: I think this would be _slightly_ less repetitive and a bit clearer if you changed this to:
```
original_collection_native_xml_noko_obj = Nokogiri::XML(cmr_response_original.body) { |config| config.strict.noblanks }
      translated_collection_native_xml_noko_obj = Nokogiri::XML(cmr_response_translated.body) { |config| config.strict.noblanks }

if content_type.include?('iso') || content_type.include?('dif')
  original_collection_native_xml_noko_obj.remove_namespaces!
  translated_collection_native_xml_noko_obj.remove_namespaces!
end
```

ChristianTrummer99: Yeah, this is cleaner

PR #635, opened 2020-08-31
MMT-2276: View Collection Link in proposal_published_notification needs to point to the cmr collection preview, not MMT 

ChristianTrummer99:
- changed the link in proposal_published_notification.html.erb to reference services.yml
- Specs still need to be written, hence the "draft" status of this PR. However, I suspect there will be change suggestions regarding the URL generation for the link in the erb file. Feel free to make comments before I push the specs

### DIFF
--- a/app/views/proposal_mailer/proposal_published_notification.html.erb
+++ b/app/views/proposal_mailer/proposal_published_notification.html.erb
@@ -10,8 +10,8 @@
   <% unless @request_type == 'delete' %>
     <div class="center">
       <% # not sure if there needs to be some revision parameter in the URL %>
-      <%= link_to 'View Collection', "#{proposal_published_cmr_root}/search/concepts/#{@concept_id}", class: 'btn btn-blue' %>
-      <% puts "#{proposal_published_cmr_root}/search/concepts/#{@concept_id}" %>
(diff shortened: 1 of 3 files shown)


app/views/proposal_mailer/proposal_published_notification.html.erb
```diff
@@ -9,7 +9,8 @@
   </p>
   <% unless @request_type == 'delete' %>
     <div class="center">
-      <%= link_to 'View Collection', collection_url(@concept_id, revision_id: @revision_id), class: 'btn btn-blue' %>
+      <% # not sure if there needs to be some revision parameter in the URL %>
```
ChristianTrummer99, line 12: not sure if there needs to be some revision parameter in the URL, my understanding is that the most recent revision is the one that would be returned

app/helpers/proposals_helper.rb
```diff
@@ -131,4 +131,14 @@ def get_available_actions_text
   def display_submitter_name(submitter_id, user_hash = {})
     user_hash[submitter_id] || 'Pending'
   end
+
+  def proposal_published_notification_view_collections_link
+    if Rails.env.development? || Rails.env.test?
+      Rails.configuration.services.dig('earthdata', 'sit', 'cmr_root')
```
charleshuang80, line 137: I thought we talked about these pointing to the local cmr.

ChristianTrummer99: Okay I think I fixed this. Let me know if it doesnt work

PR #636, opened 2020-09-01
MMT-2363 Allow ENV to control EDL login access in dMMT

ryanmiller-1:


spec/features/users/user_login_spec.rb
```diff
@@ -295,5 +295,24 @@
       end
 
     end
+
+    context 'when both Earthdata Login and Launchpad Login requirements are turned off', js: true do
```
charleshuang80, line 299: shouldn't we have this test for both MMT and dMMT?

charleshuang80: also maybe check for the buttons?

ryanmiller-1: The test already existed for MMT.  Updated both tests to have modern expectations.  They are now using the have_no_link version of what other tests use to confirm the presence of the buttons.

PR #641, opened 2020-09-11
MMT-2325 updating data migration to account for improperly formatted existing variables

ryanmiller-1:
Data migration failed because there were existing drafts which had a string for characteristics and strings do not respond to dig.

### DIFF
--- a/db/migrate/20200903193441_umm_v_1_7_data_migration.rb
+++ b/db/migrate/20200903193441_umm_v_1_7_data_migration.rb
@@ -36,8 +36,11 @@ class UmmV17DataMigration < ActiveRecord::Migration[5.2]
     characteristics = draft['Characteristics']
     return draft unless characteristics.is_a?(Hash) && characteristics['GroupPath']
 
+    # There may be cases where an existing group_path contains the name already
+    # If there is no name or if the name is already within the group_path,
+    # save that as new name.
     group_path = characteristics['GroupPath']
-    if group_path && !draft['Name']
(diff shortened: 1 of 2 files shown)


jdf-eos (approved):

db/migrate/20200903193441_umm_v_1_7_data_migration.rb
```diff
@@ -31,23 +31,34 @@ def down
...
+      return draft
+    end
+
+    draft['Name'] = if draft['Name'].start_with?('/') && group_path.end_with?('/')
+                      "#{group_path.chop}#{draft['Name']}"
+                    elsif draft['Name'].start_with?('/') || group_path.end_with?('/')
+                      "#{group_path}#{draft['Name']}"
                     else
-                      "#{draft.dig('Characteristics', 'GroupPath')}/#{draft['Name']}"
+                      "#{group_path}/#{draft['Name']}"
```
william-valencia, line 50: There might be a chance that the group path already includes the base variable name (Most variables generated with UVG would be like this). If that is the case then the new name might have a double base name. Maybe add a check where 
if group_path.end_with?(draft['Name'])
  "#{group_path}"

charleshuang80 (approved):

db/migrate/20200903193441_umm_v_1_7_data_migration.rb
ryanmiller-1: done

william-valencia (approved):

PR #644, opened 2020-09-16
MMT-2331: notify user of associated variable when deleting collection

charleshuang80:
* add getting variable count on collections show page
* modify modal to display proper messages for granules and/or variables
* ensure workflows for deleting collection with granules or variables all work
* update tests for deleting collection

* add helper methods for modal close links and use on collections show page
* some whitespace changes in application controller to appease rubocop
* add try/catch for echo provider token in development

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -38,7 +38,7 @@ class CollectionsController < ManageCollectionsController
   end
 
   def destroy
-    if @num_granules > 0 && params['confirmation-text'] != DELETE_CONFIRMATION_TEXT
+    if (@num_granules > 0 || @num_variables > 0) && params['confirmation-text'] != DELETE_CONFIRMATION_TEXT
       flash[:error] = 'Collection was not deleted because incorrect confirmation text was provided.'
       render :show and return
     end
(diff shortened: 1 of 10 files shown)


app/controllers/collections_controller.rb
```diff
@@ -144,6 +144,7 @@ def set_collection
       @native_id = meta['native-id']
       @provider_id = meta['provider-id']
       @num_granules = meta['granule-count']
+      @num_variables = meta.fetch('associations', {}).fetch('variables', []).count
```
ryanmiller-1, line 147: I think this can dig.

app/helpers/modals_helper.rb
```diff
@@ -0,0 +1,28 @@
+module ModalsHelper
+
+  def modal_close_x_link(alternate_link: nil, alternate_link_condition: nil)
```
ryanmiller-1, line 3: I like the readability improvement here.

app/controllers/collections_controller.rb
ChristianTrummer99: @ryanmiller-1 `.count` raises an error when it's evaluated on a nil object (ie. `nil.count`) and `.dig` returns a nil object if the keys passed aren't found (or if the desired value happens to be nil), but `.fetch` allows that optional parameter to be passed which helps with the error handling

ChristianTrummer99 (approved):

ryanmiller-1 (approved):

PR #647, opened 2020-09-23
MMT-2352: Updated schema with new TilingID enum

ChristianTrummer99:
- Updated UMM-C to 1.15.4 which includes one addition to the TilingIdentificationEnum 
- added to existing spec to ensure this new value is valid


spec/features/collection_drafts/forms/spatial_information_form_spec.rb
```diff
@@ -103,6 +103,19 @@
               fill_in 'Maximum Value', with: '15.0'
             end
           end
+          click_on 'Add another Tiling Identification System'
+
+          within '.multiple-item-2' do
```
ryanmiller-1, line 108: I am of two minds.  

1. I don't think that we need to test filling in a specific enum when we change the schema. To that end, I think it would be sufficient to check that the list of choices is correct.
2. The entire point of this schema change is to add this enum, so testing that the schema change was successful means publishing this value.

config/application.rb
```diff
@@ -46,8 +46,8 @@ class Application < Rails::Application
     config.services = YAML.load_file(Rails.root.join('config/services.yml'))
 
     # Versions of UMM for the different metadata types MMT is on
-    config.umm_c_version = 'vnd.nasa.cmr.umm+json; version=1.15.3'
-    config.umm_var_version = 'vnd.nasa.cmr.umm+json; version=1.7'
```
ryanmiller-1, line 50: This needs to be 1.7

spec/features/collection_drafts/publish_collection_draft_spec.rb
```diff
@@ -3,6 +3,32 @@
     login
   end
 
+  context 'when publishing a collection draft with newly added forms, values, features when UMM-C is updated' do
+    before do
+      @email_count = ActionMailer::Base.deliveries.count
```
ryanmiller-1, line 8: not using this

spec/features/collection_drafts/publish_collection_draft_spec.rb
```diff
@@ -3,6 +3,32 @@
...
+
+      draft = create(:full_collection_draft, user: User.where(urs_uid: 'testuser').first, draft_short_name: '12345', draft_entry_title: 'Draft Title')
```
ryanmiller-1, line 10: I would generally avoid using specific short_names/titles when publishing unless you need to.  Especially if they do not contain a random component.

spec/features/collection_drafts/forms/spatial_information_form_spec.rb
charleshuang80: I don't know that we need a specific new test based on this change. However, I am a little surprised that we don't have a test for the form that this enum appears on, where we test that all the options appear in the select. If that does not exist we should create such a test.

charleshuang80:
I am thinking/debating that we might need a migration for this schema change. It seems like the `up` method might not have anything, but the `down` method should. It may help us avoid the data migration issue we had with the UMM-Var update.

config/application.rb
ChristianTrummer99: must've happened when I pulled in master

spec/features/collection_drafts/publish_collection_draft_spec.rb
ChristianTrummer99: ok

spec/features/collection_drafts/publish_collection_draft_spec.rb
ChristianTrummer99: ok

db/migrate/20200924150827_umm_c_15_3_to_15_4.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def down
+    collection_drafts = CollectionDraft.where.not(draft: {})
+    proposals = CollectionDraftProposal.where.not(draft: {})
+    templates = CollectionTemplate.where.not(draft: {})
+
+    records = collection_drafts + proposals + templates
+
+    records.each do |record|
+      draft = record.draft
```
ryanmiller-1, line 13: Unused

db/migrate/20200924150827_umm_c_15_3_to_15_4.rb
```diff
@@ -0,0 +1,19 @@
...
+      record['TilingIdentificationSystems'].each do |tiling_id_system|
+        tiling_id_system['TilingIdentificationSystemName'].delete if tiling_id_system['TilingIdentificationSystemName'] = 'Military Grid Reference System'
+      end if record['TilingIdentificationSystems'].present?
+    end
```
ryanmiller-1, line 17: Need to save the record for the changes to stick, this modifies the data and then discards the modifications.  Should write a test to prove the migration works.

spec/features/collection_drafts/forms/spatial_information_form_spec.rb
```diff
@@ -122,6 +122,21 @@
         expect(page).to have_content('Collection Draft Updated Successfully!')
       end
 
+      it 'contains the correct dropdown options' do
+        within '#draft_tiling_identification_systems_0_tiling_identification_system_name' do
+          expect(page).to have_css('option', count: 10)
```
charleshuang80, line 127: There is an assertion for select that has an `options` option, which would be more efficient

charleshuang80: https://www.rubydoc.info/github/jnicklas/capybara/Capybara%2FNode%2FMatchers:has_select%3F

spec/features/collection_drafts/publish_collection_draft_spec.rb
```diff
@@ -3,6 +3,30 @@
     login
   end
 
+  context 'when publishing a collection draft with newly added forms, values, features when UMM-C is updated' do
+    before do
+      draft = create(:full_collection_draft, user: User.where(urs_uid: 'testuser').first)
+      visit collection_draft_path(draft)
+    end
+
+    context 'when new enum value is added in UMM 1.15.4' do
```
charleshuang80, line 12: I still don't think it's necessary to have this specific test here, but if you want to keep it, I think it should be moved. 
I am not sure where at the moment, but I'm thinking possibly a different file for 'publishing the latest version' (but that is redundant) or possibly in the migrations tests folder, again for 'publishing the latest version' (which is still redundant).

spec/features/collection_drafts/forms/spatial_information_form_spec.rb
ChristianTrummer99: Yeah that's much cleaner

spec/features/collection_drafts/publish_collection_draft_spec.rb
ChristianTrummer99: Let me know if you come up with another location, open to any ideas

spec/features/collection_drafts/forms/spatial_information_form_spec.rb
```diff
@@ -122,6 +122,12 @@
         expect(page).to have_content('Collection Draft Updated Successfully!')
       end
 
+      it 'contains the correct dropdown options' do
+        within '#draft_tiling_identification_systems_0_tiling_identification_system_name' do
+          page.has_select?('Tiling Identification System Name', with_options: ['CALIPSO', 'MISR', 'MODIS Tile EASE', 'MODIS Tile SIN', 'WELD Alaska Tile', 'WELD CONUS Tile', 'WRS-1', 'WRS-2', 'Military Grid Reference System'])
```
ryanmiller-1, line 127: This should use `options:` and not `with_options:`
with_options is checking partial set match, meaning adding new options doesn't cause the test to fail.  I think we want this test to be brittle.

spec/migrations/umm_c_15_3_to_15_4_spec.rb
```diff
@@ -0,0 +1,41 @@
+require File.join(Rails.root, 'db', 'migrate', '20200924150827_umm_c_15_3_to_15_4')
+
+describe 'Migration tests for UMM-C v1.15.3 => 1.15.4' do
+  context 'when doing the down migration for MMT' do
+    before do
+      set_as_mmt_proper
```
ryanmiller-1, line 6: This shouldn't be necessary.

ryanmiller-1 (approved):

jdf-eos (approved):

spec/features/collection_drafts/forms/spatial_information_form_spec.rb
```diff
@@ -122,6 +122,12 @@
         expect(page).to have_content('Collection Draft Updated Successfully!')
       end
 
+      it 'contains the correct dropdown options' do
+        within '#draft_tiling_identification_systems_0_tiling_identification_system_name' do
+          page.has_select?('Tiling Identification System Name', options: ['CALIPSO', 'MISR', 'MODIS Tile EASE', 'MODIS Tile SIN', 'WELD Alaska Tile', 'WELD CONUS Tile', 'WRS-1', 'WRS-2', 'Military Grid Reference System'])
```
charleshuang80, line 127: this should follow how all the other tests are written - `expect(page).to have_select(...`

ChristianTrummer99: @charleshuang80 That was my first thought when you suggested using this matcher, but as I tried to adapt it to the `expect(page).to` style, my tests kept failing. I've tried `expect(page).to have_select?('Tiling Identification System Name', options: ...` and `expect(page).to has_select?('Tiling Identification System Name', options: ...` with no success. I think that matcher only works as it is above

charleshuang80 (approved):

william-valencia (approved):

PR #650, opened 2020-09-30
MMT-2383 Upgraded Coffeescript to 2.5.1

william-valencia:
Upgraded Coffeescript to 2.5.1 and Updated the local_cmr.rake file to use harmony flag

### DIFF
--- a/lib/tasks/local_cmr.rake
+++ b/lib/tasks/local_cmr.rake
@@ -191,9 +191,7 @@ namespace :cmr do
 
           puts "\nCompressing..."
           
-          #uncomment the following if you need to compile with JS 6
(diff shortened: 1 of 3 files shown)


ChristianTrummer99 (approved):

charleshuang80 (approved):
if we're not up to date, should we update via npm?
Also, is there anything in the readme that should be updated?

william-valencia:
> if we're not up to date, should we update via npm?
> Also, is there anything in the readme that should be updated?

After the merge, you should update your global coffeescript to be 2.5.1
"coffee -v" should return version 2.5.1
Nothing in MMT readme should be updated, but it will be updated in preview gem.

PR #657, opened 2020-10-14
MMT-2355: Create a variable draft with the associated collection filled out

ChristianTrummer99:
- Added text field on manage variables page for user to supply concept-id of collection they want to associate the variable with
- Added link on collection show page that allows a user to create an associated variable draft
- added functionality to the _not_current_provider_modal and change_current_provider.coffee files so that the not-current-provider-modal is displayed when a user clicks on the 'Create Associated Variable' link 
- added a child `new` method to the variable_drafts_controller

### DIFF
--- a/app/controllers/variable_drafts_controller.rb
+++ b/app/controllers/variable_drafts_controller.rb
@@ -25,6 +25,8 @@ class VariableDraftsController < BaseDraftsController
       )
 
       if current_collection_response.success? && current_collection_response.body['hits'] > 0
+        # this banner needs to be deliberated
+        flash[:success] = I18n.t("controllers.draft.variable_drafts.update_associated_collection.flash.success")
         super
       else
         Rails.logger.info("Error retrieving Collection - Concept ID: #{@associated_collection_id}")
--- a/app/controllers/variable_drafts_controller.rb
+++ b/app/controllers/variable_drafts_controller.rb
@@ -17,17 +17,21 @@ class VariableDraftsController < BaseDraftsController
   end
 
   def new
-    @associated_collection_id = params[:associated_collection_id]
+    unless params[:associated_collection_id].blank?
+      @associated_collection_id = params[:associated_collection_id].strip
 
-    current_collection_response = cmr_client.get_collections_by_post(
-      { concept_id: @associated_collection_id }, token
-    )
+      current_collection_response = cmr_client.get_collections_by_post(
+        { concept_id: @associated_collection_id }, token
+      )
 
-    if current_collection_response.success? && current_collection_response.body['hits'] > 0
-      super
+      if current_collection_response.success? && current_collection_response.body['hits'] > 0
+        super
+      else
+        Rails.logger.info("Error retrieving Collection - Concept ID: #{@associated_collection_id}")
+        redirect_to manage_variables_path, flash: { error: 'Concept ID invalid.'}
+      end
     else
-      Rails.logger.info("Error retrieving Collection")
-      redirect_to manage_variables_path, flash: { error: 'Concept ID invalid.'}
+      super
     end
   end
 
--- a/app/controllers/variable_drafts_controller.rb
+++ b/app/controllers/variable_drafts_controller.rb
@@ -24,13 +24,14 @@ class VariableDraftsController < BaseDraftsController
         { concept_id: @associated_collection_id }, token
       )
 
-      if current_collection_response.success? && current_collection_response.body['hits'] > 0
-        # this banner needs to be deliberated
-        flash[:success] = I18n.t("controllers.draft.variable_drafts.update_associated_collection.flash.success")
+      if current_collection_response.success? && current_collection_response.body['hits'] > 0 && current_provider?(current_collection_response.body.dig('items',0,'meta','provider-id'))
         super
-      else
-        Rails.logger.info("Error retrieving Collection - Concept ID: #{@associated_collection_id}")
-        redirect_to manage_variables_path, flash: { error: 'Concept ID invalid.'}
+      elsif !current_collection_response.success?
+        redirect_to manage_variables_path, flash: { error: 'Unable to search CMR.' }
+      elsif current_collection_response.body['hits'] == 0
+        redirect_to manage_variables_path, flash: { error: 'Collection not found.' }
+      elsif !current_provider?(current_collection_response.body.dig('items',0,'meta','provider-id'))
+        redirect_to manage_variables_path, flash: { error: 'Provider context must be changed.' }
       end
     else
       super
--- a/app/controllers/variable_drafts_controller.rb
+++ b/app/controllers/variable_drafts_controller.rb
@@ -27,7 +27,7 @@ class VariableDraftsController < BaseDraftsController
       if current_collection_response.success? && current_collection_response.body['hits'] > 0 && current_provider?(current_collection_response.body.dig('items',0,'meta','provider-id'))
         super
       elsif !current_collection_response.success?
-        redirect_to manage_variables_path, flash: { error: 'Unable to search CMR.' }
+        redirect_to manage_variables_path, flash: { error: current_collection_response.body['errors'].first }
       elsif current_collection_response.body['hits'] == 0
         redirect_to manage_variables_path, flash: { error: 'Collection not found.' }
       elsif !current_provider?(current_collection_response.body.dig('items',0,'meta','provider-id'))
--- a/app/controllers/variable_drafts_controller.rb
+++ b/app/controllers/variable_drafts_controller.rb
@@ -17,21 +17,19 @@ class VariableDraftsController < BaseDraftsController
   end
 
   def new
-    unless params[:associated_collection_id].blank?
-      @associated_collection_id = params[:associated_collection_id].strip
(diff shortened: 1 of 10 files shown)


app/controllers/variable_drafts_controller.rb
```diff
@@ -16,6 +16,25 @@ def edit
...
+      elsif current_collection_response.body['hits'] == 0
+        redirect_to manage_variables_path, flash: { error: "No matches were found for #{@associated_collection_id}" }
+      elsif !current_provider?(current_collection_response.body.dig('items',0,'meta','provider-id'))
+        redirect_to manage_variables_path, flash: { error: "Variables can only be associated to collections within the same provider. To create a variable for #{@associated_collection_id} you must change your provider context." }
+      end
+    else
+      super
+    end
+  end
+
```
ChristianTrummer99, line 37: @charleshuang80 @ryanmiller-1 I have given this a bit of thought and there are 2 things that I'm unsure about: 

1. Regarding the **first two** `elsif` statements; the first one was originally meant to flash a message when the call to cmr was unsuccessful (network error, invalid permission, whatever). However, since the `current_collection_response.success?` is `false` when a user enters jibberish, for instance, this infringes on the purpose of the next `elsif` statement which is to flash a message when the supplied concept-id is not found in CMR. To rectify this, all I've done is surface the cmr error response. I was hoping to get your take on the overlap between these two conditionals and if it even matters. 
2. I am displeased with the readability of this whole controller method because of the nested conditionals but I can't think of a simpler way to do this. I was hoping you guys might have advice for this sort of thing as well.

spec/features/variable_drafts/create_variable_draft_with_associated_collection_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+      end
+
+      it 'does not navigate to the new_variable_draft_path' do
+        expect(page).to have_no_content('Variable Information')
+        expect(page).to have_no_css('.umm-form')
+        expect(page).to have_no_css('.variable-form')
+      end
+    end
+
+    context 'when no concept-id is supplied' do
```
ryanmiller-1, line 92: I thought that this would have been covered in the manage variable tests, but I didn't see one covering it.  This one does make me wonder if all of these tests should be in the manage variables files.

spec/features/variable_drafts/create_variable_draft_with_associated_collection_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+        within '#variable-draft-collection-association-table tbody' do
+          expect(page).to have_no_content('CIESIN_SEDAC_ANTHROMES_v2_1700')
+          expect(page).to have_content('No Collection Association found. A Collection must be selected in order to publish this Variable Draft. Each Variable can only be associated with a single Collection.')
+          expect(page).to have_css('tr', count: 1)
+        end
+      end
+    end
+  end
+
+  context "when creating a variable draft and associating a collection from the collection's show page" do
```
ryanmiller-1, line 116: Same question as manage variables; should these go with collections?

ryanmiller-1 (changes requested):

app/controllers/variable_drafts_controller.rb
ryanmiller-1: Also, FWIW, I think that the number of times a user enters an invalid concept_id will be very small.

ChristianTrummer99: - I like that, definitely going to implement the guard clause (I'll make sure to check it as well)
- Okay that makes sense, I will leave it
- Yeah I suppose so.

app/views/collections/show.html.erb
```diff
@@ -81,6 +81,13 @@
 
             <%= link_to "Tags (#{@num_tags || 0})", '#tags-modal', class: 'display-modal bar-after' %>
 
+            <% if current_provider?(@provider_id) %>
```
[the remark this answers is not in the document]
ChristianTrummer99: Not sure. I slacked you with clarification

spec/features/variable_drafts/create_variable_draft_with_associated_collection_spec.rb
ChristianTrummer99: The reason I put everything with the variable drafts files was because the spec in the manage_variables folder doesn't test the creation of variable drafts, it just makes sure the open drafts on the manage variables page is being rendered properly. Let me know what you think I should do.

spec/features/variable_drafts/create_variable_draft_with_associated_collection_spec.rb
ChristianTrummer99: Good question. My rationale was that since we are not _affecting_ a collection but rather a variable draft, it should be in the variable drafts folder. But I see where you are coming from, additionally, it is tricky because there isn't necessarily precedent to reference for a situation like this.

app/controllers/variable_drafts_controller.rb
charleshuang80: For 1: I think surfacing a CMR error is sufficient, we would like their error messages to be as user friendly as possible and should avoid having to interpret them for the user. For the case of 0 hits, did you run that language by the PO? I feel like we might want a tiny bit more information in that error message, but if it was approved by PO then that's fine.

For 2: I agree a guard close would help. I would actually add a second guard clause, checking for the current provider. I don't think we even need to make the call to CMR to verify the collection if it's the wrong provider.

spec/features/variable_drafts/create_variable_draft_with_associated_collection_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+      end
+    end
+
+    context 'when the supplied concept-id is not found' do
+      before do
+        fill_in 'associated_collection_id', with: 'C1-INVALID'
+        click_on 'Create New Record'
+      end
+
+      it 'does not navigate to the new_variable_draft_path' do
```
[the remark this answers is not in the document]
charleshuang80: I agree, I think this should be reworded to say where it takes you, checking something to verify that page, and then also checking that the other content is not on the page.

app/controllers/variable_drafts_controller.rb
ChristianTrummer99: 1 - Yeah, PO requested it; originally had something different.

2 - That's a great point, will do!

charleshuang80: so the way that a collection concept id is constructed is `C` `<number>` `-` `<provider_id>`. I think you can split it on the `-` and assume the last part is the provider id. But if you think it is safer or would prefer to use the provider id from the response you don't have to add the guard clause the way I suggested.

ChristianTrummer99: Honestly, after thinking about changing this, I am on the fence. Your suggestion would work fine, I really can't find any flaws - it would remove one of the conditionals and be easy to implement. However, I am on the fence because we don't really do this to get the `provider-id` elsewhere. What do you think I should do?

spec/features/variable_drafts/create_variable_draft_with_associated_collection_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+      before do
+        fill_in 'associated_collection_id', with: invalid_concept_id
+        click_on 'Create New Record'
+      end
+
+      it 'surfaces the CMR error message' do
+        expect(page).to have_content('Concept-id [asdf] is not valid.')
+      end
+
+      it 'remains on the manage variables page' do
```
charleshuang80, line 83: You should be checking for some content (like above) that you are on the right page as well.

ChristianTrummer99: Good catch.

app/controllers/variable_drafts_controller.rb
ChristianTrummer99: @charleshuang80

charleshuang80: Leaving it as is is definitely a bit 'safer'. But I lean towards adding the additional guard clause - the concept id construction doesn't change and I believe we use it for groups (esp verifying system groups) and validating subscriptions.

app/controllers/variable_drafts_controller.rb
```diff
@@ -16,6 +16,23 @@ def edit
...
   end
 
+  def new
+    super and return if params[:associated_collection_id].blank?
+
+    @associated_collection_id = params[:associated_collection_id].strip
+    current_collection_response = cmr_client.get_collections_by_post({ concept_id: @associated_collection_id }, token)
+
+    if current_collection_response.success? && current_collection_response.body['hits'] > 0 && current_provider?(current_collection_response.body.dig('items',0,'meta','provider-id'))
+      super
```
jdf-eos, line 26: super already called.

jdf-eos (changes requested):

app/controllers/variable_drafts_controller.rb
jdf-eos: nvm.     I don't use and as it is largely obfuscatory.

jdf-eos (approved):

charleshuang80 (approved):

PR #661, opened 2020-10-26
MMT-2378: update UMM-S forms to v1.3.4

charleshuang80:
* update form json to add Subset fields
* update existing and add new UMM Form Field classes to accommodate new field types
* add js for form
* update js for validation and required fields
* update test and factory


lib/json_schema_form/umm_form_elements/umm_checkbox_dependent_fields.rb
```diff
@@ -0,0 +1,35 @@
+# UmmCheckboxDependentFields are used for form fields that should be hidden
+# the fields have a value or the user chooses to show them
```
ryanmiller-1, line 2: I feel like this is missing a word.

app/assets/javascripts/validation.coffee
```diff
@@ -358,6 +357,21 @@ $(document).ready ->
...
+
+    if error.dataPath.split('/')[error.dataPath.split('/').length - 1] == 'AllowMultipleValues'
+      # Allow Multiple Values fields, under Subset, are boolean fields represented
+      # by radio buttons
+      # for a required error, because these are radio buttons, we want the error
+      # to be surfaced with the radio parent container, which the path
+      # transformation has taken care of
+      if error.message == 'should be boolean'
+        # html cannot hold boolean values, they are represented as string
+        # TODO: we can check the values to verify they are 'true' and 'false'.
```
ryanmiller-1, line 369: This feels excessive to me because I'm not sure what scenario leads to this value not being true or false.

ryanmiller-1: OTOH, it shouldn't be burdensome to check...

ryanmiller-1 (approved):

spec/features/service_drafts/forms/options_form_spec.rb
```diff
@@ -7,7 +7,42 @@
 
   context 'when submitting the form' do
     before do
-      select 'Spatial', from: 'Subset Types'
+      check 'Spatial'
```
jdf-eos, line 10: Is there a reason why the other types were omitted ( line, circle)?

jdf-eos (approved):

spec/features/service_drafts/forms/options_form_spec.rb
charleshuang80: I did not feel that every type needed to be populated and saved, but rather each unique type (based on the fields they contain). Circle has the same fields as Point, and Line has the same fields as Polygon, so I felt that those types were covered.

PR #662, opened 2020-10-27
MMT-2298: Top level required fields to display required icons on the variable, service, and tool form pages

ChristianTrummer99:
- Added the appropriate classes for top-level fields on collection_drafts and added `"htmlClass": "eui-required-o always-required"` to the umm-metadata-form.json's 
- added functionality to the UmmForm to recognize the parsed json
- wrote tests to check each accordion for the required icon


spec/features/collection_drafts/forms/required_fields_spec.rb
```diff
@@ -129,6 +129,76 @@
         end
       end
     end
+
+    context 'when checking the accordion headers for required icons' do
+      it 'displays required icons on the Collection Information accordion' do
```
ryanmiller-1, line 134: The visits should be outside the it blocks.  I think these tests fit in each forms test file, so, spec/features/collection_drafts/forms/collection_information_form_spec.rb, for instance.

app/views/collection_drafts/forms/_acquisition_information.html.erb
```diff
@@ -1,7 +1,7 @@
 <!-- Platforms [r] -->
 <fieldset class="eui-accordion is-closed" id="platforms">
   <div class="eui-accordion__header">
-    <h3 class="eui-accordion__title eui-required-o">Platforms</h3>
+    <h3 class="eui-accordion__title eui-required-o always-required">Platforms</h3>
```
charleshuang80, line 4: So, I did not check and did not realize that collections did it this way. I may not have insisted on adding `always-required` in that case, but I still think this makes it more clear.

lib/json_schema_form/umm_form.rb
```diff
@@ -329,7 +329,7 @@ def render_markup
 # :nodoc:
 class UmmFormAccordion < UmmForm
   def render_markup
-    content_tag(:fieldset, class: "eui-accordion is-closed #{parsed_json['htmlClass']}", id: accordion_id) do
+    content_tag(:fieldset, class: "eui-accordion is-closed", id: accordion_id) do
```
charleshuang80, line 332: I am thinking we may not want to move the `parsed_json['htmlClass']` from the top level tag for accordions, even though it might not be used at the moment. I am thinking we should use another key for the accordion header classes (something like `accordionHeaderClass` maybe) to be more specific about where they'll be applied.

spec/features/collection_drafts/forms/acquisition_information_form_spec.rb
```diff
@@ -4,6 +4,19 @@
     draft = create(:collection_draft, user: User.where(urs_uid: 'testuser').first)
     visit collection_draft_path(draft)
   end
+  
+  context 'when checking the accordion headers for required icons' do
+    before do
+      within '.metadata' do
```
ryanmiller-1, line 10: Since this is shared across both contexts, this can be moved into the describe's before block.

ChristianTrummer99: Okay will do - there are a lot of files like this, so I can make sure they're all adjusted

lib/json_schema_form/umm_form.rb
ChristianTrummer99: Okay, will do

ryanmiller-1 (approved):

charleshuang80 (approved):

PR #663, opened 2020-10-27
MMT-2416 Adding Collection-Tool Associations

ryanmiller-1:
Added manage associations link to published tool page.
Added correct underlining when looking at tool association pages.
Updated collection associations controller and change provider modal
Added associated routes
Updated cmr_client to share code for service/tool associations
Copied tests from service associations and updated for tools

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -224,14 +224,6 @@ en:
           flash:
             success: 'Collection Association Updated Successfully!'
             error: 'Collection Associaiton was not updated successfully.'
-        destroy:
-          flash:
-            success: 'Collection Association Deleted Successfully!'
-            error: 'Collection Association was not deleted successfully'
-        create:
-          flash:
-            success: 'Collection Association Created Successfully!'
-            error: 'Collection Association was not created successfully'
       services:
         destroy:
           flash:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -227,12 +227,21 @@ en:
         destroy:
           flash:
             success: 'Collection Association Deleted Successfully!'
-            error: 'Collection Association were not deleted successfully'
+            error: 'Collection Association was not deleted successfully'
         create:
           flash:
             success: 'Collection Association Created Successfully!'
-            error: 'Collection Association were not created successfully'
+            error: 'Collection Association was not created successfully'
       services:
+        destroy:
+          flash:
+            success: 'Collection Associations Deleted Successfully!'
+            error: 'Collection Associations were not deleted successfully'
+        create:
+          flash:
+            success: 'Collection Associations Created Successfully!'
+            error: 'Collection Associations were not created successfully'
+      tools:
         destroy:
           flash:
             success: 'Collection Associations Deleted Successfully!'
--- a/lib/cmr/cmr_client.rb
+++ b/lib/cmr/cmr_client.rb
@@ -99,10 +99,18 @@ module Cmr
     end
 
     def add_collection_assocations_to_service(concept_id, collection_ids, token)
+      add_collection_assocations(concept_id, collection_ids, token, 'services')
+    end
+
+    def add_collection_assocations_to_tool(concept_id, collection_ids, token)
+      add_collection_assocations(concept_id, collection_ids, token, 'tools')
+    end
+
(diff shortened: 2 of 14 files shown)


config/locales/en.yml
```diff
@@ -227,11 +227,11 @@ en:
         destroy:
```
charleshuang80, line 227: if these actions are not possible anymore, should we remove the messages here?

charleshuang80 (approved):

lib/cmr/cmr_client.rb
```diff
@@ -99,10 +99,18 @@ def test_query(query, token, headers = {})
...
 
     def add_collection_assocations_to_service(concept_id, collection_ids, token)
+      add_collection_assocations(concept_id, collection_ids, token, 'services')
+    end
+
+    def add_collection_assocations_to_tool(concept_id, collection_ids, token)
+      add_collection_assocations(concept_id, collection_ids, token, 'tools')
+    end
+
+    def add_collection_assocations(concept_id, collection_ids, token, concept_type)
```
ChristianTrummer99, line 109: You might consider adding something like `concept_type = (concept_id[0] == 'T' ? 'tools' : 'services')` instead of using different methods. The way I see it, the `add_collection_associations` method is only used for non-collection metadata so maintenance would be low if we just skimmed the first character of the concept-id. Your methods are a bit more readable though, so either way works

lib/cmr/cmr_client.rb
```diff
@@ -113,10 +121,18 @@ def add_collection_assocations_to_service(concept_id, collection_ids, token)
...
 
     def delete_collection_assocations_to_service(concept_id, collection_ids, token)
+      delete_collection_associations(concept_id, collection_ids, token, 'services')
+    end
+
+    def delete_collection_assocations_to_tool(concept_id, collection_ids, token)
+      delete_collection_associations(concept_id, collection_ids, token, 'tools')
+    end
+
+    def delete_collection_associations(concept_id, collection_ids, token, concept_type)
```
ChristianTrummer99, line 131: same as above

spec/features/services/delete_collection_assocation_spec.rb
```diff
@@ -59,7 +57,7 @@
             click_link 'refresh the page'
           end
 
-          it 'reloades the page and dislay the correct associations' do
+          it 'reloads the page and dislay the correct associations' do
```
ChristianTrummer99, line 60: 'display' is also misspelled

spec/features/tools/create_collection_association_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+      before do
+        within '#collection-search' do
+          select 'Entry Title', from: 'Search Field'
+          find(:css, "input[id$='query_text']").set('MODIS-I*')
+
+          click_button 'Submit'
+        end
+      end
+
+      it 'display the correct search results' do
```
ChristianTrummer99, line 93: being picky: grammar

ChristianTrummer99 (commented):
- Getting an error banner when I click ‘Manage Associated Collections’ on published tool show page
- Getting actual failure when I click ‘Add Associated Collection’ due to "undefined method 'any?' for nil:NilClass"
- I suggest Rails logging for the `get_all_associated_collections` method for app/controllers/collection_associations_controller.rb:149 because I can't tell if this is due to the conditions not being met or the `association_response` failing
- app/controllers/collection_associations_controller.rb:155 is causing issues because `partial_associated_collections` is `nil`
- The above issues could be due to something ill-configured on my machine given the content of the error banner
- spec/features/tools/create_collection_association_spec.rb is failing locally for me
In summary, I think these issues might be something wrong with my machine, but the handling of `partial_associated_collections` being `nil` should probably be friendlier

lib/cmr/cmr_client.rb
ryanmiller-1: Resolving this by changing the call in the controller.

ryanmiller-1:
I believe that all of those problems are resolved with an uberjar update.  An old uberjar won't recognize tool_concept_id as a parameter for collection searching and will not have the end points to make tool associations.  Regardless, I have made failure for that collection search more friendly.

ChristianTrummer99 (approved):

PR #664, opened 2020-10-29
Feature/MMT-2344 Merging UMM-S v1.3.4 Feature Branch into Master

ryanmiller-1:


william-valencia (approved):

app/assets/javascripts/required_fields.coffee
```diff
@@ -165,6 +165,13 @@ $(document).ready ->
         index = id.lastIndexOf(last)
         id = id.slice(0, index)
       label = $("label[for='#{id}']")
+      # if element is a radio button, we want to add the icon to the label
+      # for the group of radio buttons
+      # this was added for UMM-S v1.3.4, but should be evaluated for other types
+      if $(element).is(':radio')
+        $radioParent = $(element).closest('.radio-group-parent')
+        label = $radioParent.siblings('label.radio-group-label')
+
```
ChristianTrummer99, line 174: The required icons for ‘Allow multiple values’ only show up when the user clicks the radio button which means they don’t know the field is required until after they click (which I assume is not the expected behavior). This appears to be caused by the fact that `getLabels` is invoked in the `$optionallyRequiredWithValueFields` loop on 197 in required_fields.coffee

spec/features/service_drafts/form_navigation/options_form_nav_spec.rb
```diff
@@ -144,7 +144,7 @@
 
     it 'displays the correct number of required fields' do
       # 1 for each input and output format fields in each paired reformatting
-      expect(page).to have_selector('label.eui-required-o', count: 4)
+      expect(page).to have_selector('label.eui-required-o', count: 9)
```
ChristianTrummer99, line 147: This is passing because the factory draft is full (which means the required icons will be displayed). Ignore this is my assumption about the required icons behavior is incorrect.

ChristianTrummer99 (approved):
I think the verbiage for the help section of the 'Allow Multiple Values' could be improved:  Instead of “A value of false or the element is not used means the service only accepts a single <spatial subset>”,  Maybe: “Select nothing or False to allow the service to accept only a single <spatial subset>”. More than anything I was confused by “A value of false or the element is not used”. I give Approval here so if my assumption about the required icons is wrong you can move this ticket along.

app/assets/javascripts/required_fields.coffee
ryanmiller-1: This is the expected behavior. The parent object is not required, so those fields are not always required. When the parent field exists, however, it must contain these fields.

ryanmiller-1:
That verbage is derived from the schema.  We do not typically modify it on our own, but we can look at it the next time the service schema is being modified.

jdf-eos (approved):

PR #667, opened 2020-11-04
MMT-2418: DraftMMT User reported issue with metadata creation date

ChristianTrummer99:
- added auto-populate feature when publishing proposals
- tests check to make sure the `CREATE` and `UPDATE` fields are being correctly populated


app/controllers/manage_proposal_controller.rb
```diff
@@ -56,6 +56,8 @@ def show
   def publish_proposal
     proposal = JSON.parse(params['proposal_data'])
 
+    add_dates(draft: proposal['draft'])
```
ryanmiller-1, line 59: An alternate way that reduces duplicate code is something like:
```
proposal_with_dates = CollectionDraftProposal.new(draft: proposal['draft']).add_metadata_dates
proposal = proposal_with_dates.draft
```

spec/features/manage_proposals/approver_workflow_spec.rb
```diff
@@ -8,6 +8,50 @@
   end
 
   context 'when loading the manage proposals page' do
+    context 'when publishing a create metadata proposal' do
```
ryanmiller-1, line 11: Should have a different context text for different examples

spec/features/manage_proposals/approver_workflow_spec.rb
```diff
@@ -8,6 +8,50 @@
...
+        within '.open-draft-proposals tbody tr:nth-child(1)' do
+          click_on 'Publish'
+        end
+        select 'MMT_2', from: 'provider-publish-target'
+        within '#approver-proposal-modal' do
+          click_on 'Publish'
+        end
+      end
+
+      it 'displays expected success messages' do
```
ryanmiller-1, line 39: I think this test is fine as a sanity check, but not strictly necessary since this is being covered below.

spec/features/manage_proposals/approver_workflow_spec.rb
```diff
@@ -8,6 +8,50 @@
...
+        expect(page).to have_content('Collection Metadata Successfully Published!')
+      end
+
+      it 'deletes the proposal in dMMT' do
```
ryanmiller-1, line 43: I think this test can be removed since it is covered below.

spec/features/manage_proposals/approver_workflow_spec.rb
```diff
@@ -8,6 +8,50 @@
...
+        expect(CollectionDraftProposal.where(short_name: short_name).count).to eq(0)
+      end
+
+      it 'contains the correct CREATE and UPDATE dates' do
```
ryanmiller-1, line 47: I think it is worth checking that the proposal did not have these dates before being published. Should trigger an alarm if somebody modifies the factory, for some reason.

app/controllers/manage_proposal_controller.rb
ChristianTrummer99: I didn't do this because the rest of the `manage_proposal_controller` is set up to process the `proposal` variable as a hash instead of a `CollectionDraftProposal` object which causes multiple failures. Good examples of this are the `publish_create_or_update_proposal`, `update_proposal_status_in_dmmt`, and `publish_delete_proposal`. Here are our options: 
1. change the paradigm for how this controller processes proposals - a lot of careful adjustments and potentially some rewriting
2. Use the `CollectionDraftProposal` solely as a means to access the `add_metadata_dates` method and set `proposal['draft'] = proposal_with_dates.draft` essentially just updating this part of the `proposal` hash. 

Thoughts?

ChristianTrummer99: Had to make some slight changes for this to work

ryanmiller-1 (approved):

charleshuang80 (approved):

william-valencia (approved):

PR #670, opened 2020-11-19
MMT-2400 MMT-2401: add GKR Keyword Recommendations to Descriptive Keywords form

charleshuang80:
* add to client to get recommendations from GKR
* add models and associations for recording if recommendations shown for collection drafts, templates, proposals
* display keyword recommendations for science keywords on form
* add js for marking when recommendations viewed
* add feature toggle for gkr functionality
* add tests
* add VCR cassette

### DIFF
--- a/app/concerns/controlled_keywords.rb
+++ b/app/concerns/controlled_keywords.rb
@@ -46,6 +46,17 @@ module ControlledKeywords
     @science_keywords = keywords
   end
 
+  def fetch_keyword_recommendations
+    abstract = get_resource.draft.fetch('Abstract', '')
+    response = cmr_client.fetch_keyword_recommendations(abstract)
+    @keyword_recommendations = if response.success?
+                                 response.body['recommendations'].map { |rec| rec['keyword'] }
+                               else
+                                 []
+                               end
+    # TODO log recommendations
+  end
+
   def set_service_keywords
     keywords = fetch_science_keywords
     if keywords.key? 'category'
--- a/app/concerns/controlled_keywords.rb
+++ b/app/concerns/controlled_keywords.rb
@@ -46,17 +46,6 @@ module ControlledKeywords
     @science_keywords = keywords
   end
 
-  def fetch_keyword_recommendations
-    abstract = get_resource.draft.fetch('Abstract', '')
-    response = cmr_client.fetch_keyword_recommendations(abstract)
-    @keyword_recommendations = if response.success?
-                                 response.body['recommendations'].map { |rec| rec['keyword'] }
-                               else
-                                 []
-                               end
-    # TODO log recommendations
-  end
-
   def set_service_keywords
     keywords = fetch_science_keywords
     if keywords.key? 'category'
--- a/app/controllers/collection_drafts_controller.rb
+++ b/app/controllers/collection_drafts_controller.rb
@@ -71,7 +71,7 @@ class CollectionDraftsController < BaseDraftsController
       Rails.logger.info("Audit Log: Metadata update attempt when #{current_user.urs_uid} successfully modified #{resource_name.titleize} with title: '#{get_resource.entry_title}' and id: #{get_resource.id} for provider #{current_user.provider_id}")
       flash[:success] = I18n.t("controllers.draft.#{plural_resource_name}.update.flash.success")
 
-      if params[:recommended_keywords_viewed] == 'true'
+      if gkr_enabled? && params[:recommended_keywords_viewed] == 'true'
         # create keyword_recommendation
         get_resource.record_recommendation_provided
 
@@ -482,7 +482,7 @@ class CollectionDraftsController < BaseDraftsController
     set_country_codes
     set_language_codes            if @form == 'collection_information' || @form == 'metadata_information'
     set_science_keywords          if @form == 'descriptive_keywords'
-    fetch_keyword_recommendations if @form == 'descriptive_keywords' && get_resource.keyword_recommendation_needed?
+    fetch_keyword_recommendations if @form == 'descriptive_keywords' && gkr_enabled? && get_resource.keyword_recommendation_needed?
     set_platform_types            if @form == 'acquisition_information'
     set_instruments               if @form == 'acquisition_information'
     set_projects                  if @form == 'acquisition_information'
--- a/app/controllers/collection_drafts_controller.rb
+++ b/app/controllers/collection_drafts_controller.rb
@@ -71,6 +71,13 @@ class CollectionDraftsController < BaseDraftsController
       Rails.logger.info("Audit Log: Metadata update attempt when #{current_user.urs_uid} successfully modified #{resource_name.titleize} with title: '#{get_resource.entry_title}' and id: #{get_resource.id} for provider #{current_user.provider_id}")
       flash[:success] = I18n.t("controllers.draft.#{plural_resource_name}.update.flash.success")
 
+      if params[:recommended_keywords_viewed] == 'true'
+        # create keyword_recommendation
+        get_resource.record_recommendation_provided
+
+        # TODO: if/else and logging
+      end
+
       case params[:commit]
       when 'Done'
         redirect_to get_resource
@@ -473,15 +480,16 @@ class CollectionDraftsController < BaseDraftsController
 
     # Set instance variables depending on the form requested
     set_country_codes
-    set_language_codes        if @form == 'collection_information' || @form == 'metadata_information'
-    set_science_keywords      if @form == 'descriptive_keywords'
-    set_platform_types        if @form == 'acquisition_information'
-    set_instruments           if @form == 'acquisition_information'
-    set_projects              if @form == 'acquisition_information'
-    set_temporal_keywords     if @form == 'temporal_information'
-    set_location_keywords     if @form == 'spatial_information'
-    set_data_centers          if @form == 'data_centers' || @form == 'data_contacts'
-    load_data_contacts_schema if @form == 'data_contacts'
+    set_language_codes            if @form == 'collection_information' || @form == 'metadata_information'
+    set_science_keywords          if @form == 'descriptive_keywords'
+    fetch_keyword_recommendations if @form == 'descriptive_keywords' && get_resource.keyword_recommendation_needed?
+    set_platform_types            if @form == 'acquisition_information'
+    set_instruments               if @form == 'acquisition_information'
+    set_projects                  if @form == 'acquisition_information'
+    set_temporal_keywords         if @form == 'temporal_information'
+    set_location_keywords         if @form == 'spatial_information'
+    set_data_centers              if @form == 'data_centers' || @form == 'data_contacts'
+    load_data_contacts_schema     if @form == 'data_contacts'
   end
 
   def collection_validation_setup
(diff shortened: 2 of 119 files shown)


app/controllers/collection_drafts_controller.rb
```diff
@@ -473,15 +480,16 @@ def edit_view_setup
...
-    set_platform_types        if @form == 'acquisition_information'
-    set_instruments           if @form == 'acquisition_information'
-    set_projects              if @form == 'acquisition_information'
-    set_temporal_keywords     if @form == 'temporal_information'
-    set_location_keywords     if @form == 'spatial_information'
-    set_data_centers          if @form == 'data_centers' || @form == 'data_contacts'
-    load_data_contacts_schema if @form == 'data_contacts'
+    set_language_codes            if @form == 'collection_information' || @form == 'metadata_information'
+    set_science_keywords          if @form == 'descriptive_keywords'
+    fetch_keyword_recommendations if @form == 'descriptive_keywords' && get_resource.keyword_recommendation_needed?
```
charleshuang80, line 485: add feature toggle

app/views/collection_drafts/forms/_descriptive_keywords.html.erb
```diff
@@ -34,6 +45,30 @@
                   <%= hidden_field_tag("draft[science_keywords][#{index}][variable_level_3]", science_keyword.fetch('VariableLevel3', '')) %>
                   <%= hidden_field_tag("draft[science_keywords][#{index}][detailed_variable]", science_keyword.fetch('DetailedVariable', '')) %>
                 </li>
+                <% selected_keywords_index += 1 %>
+              <% end %>
+
+              <% unless @keyword_recommendations.blank? %>
+                <% @keyword_recommendations.each_with_index do |recommended_keyword, recommendation_index| %>
+                  <% recommended_keyword_array = recommended_keyword.split('>') %>
```
ryanmiller-1, line 53: I think I would favor trimming the keywords so that we don't need to coordinate the change, then we can write a ticket to remove the trimming after they've changed.

app/views/collection_drafts/forms/_descriptive_keywords.html.erb
```diff
@@ -34,6 +45,30 @@
...
+                  <% recommended_keyword_string = recommended_keyword_array.join(' > ') %>
+                  <% next if science_keywords.any? { |stored_science_keyword| keyword_string(stored_science_keyword) == recommended_keyword_string } %>
```
ryanmiller-1, line 55: Does the logic for matching keywords belong in the controller?

app/concerns/controlled_keywords.rb
```diff
@@ -46,6 +46,17 @@ def set_science_keywords
     @science_keywords = keywords
   end
 
+  def fetch_keyword_recommendations
```
ryanmiller-1, line 49: I favor moving this to its own concern for GKR.  The recommendations are interacting with a controlled keyword field, but this is a different behavior than the rest of these.  It's not setting options.

spec/models/collection_draft_proposal_spec.rb
```diff
@@ -268,37 +268,104 @@
...
+      end
     end
 
-    it '`create_from_collection` should fail to save' do
-      collection = { 'ShortName' => '12345', 'EntryTitle' => 'test title' }
-      user = User.create(urs_uid: 'testuser', provider_id: 'MMT_2')
-      native_id = 'test_id'
-      collection_draft_proposal = CollectionDraftProposal.create_from_collection(collection, user, native_id)
+    context 'when there are proposals' do
+      # TODO: is this really needed?
```
ryanmiller-1, line 351: I think this is unnecessary

app/controllers/collection_drafts_controller.rb
```diff
@@ -517,4 +525,9 @@ def set_associated_services
 
     get_associated_services(collection_response.body['items'])
   end
+
+  ## Feature Toggle for GKR recommendations
+  def gkr_enabled?
+    Rails.configuration.gkr_enabled == true
```
ryanmiller-1, line 531: Comparison unnecessary

spec/models/collection_draft_spec.rb
```diff
@@ -232,4 +232,66 @@
...
+
+    it '"record_recommendation_provided" returns `nil`' do
+      collection_draft = create(:full_collection_draft)
+      expect(collection_draft.record_recommendation_provided).to eq(nil)
+    end
+  end
+
+  context 'when Keyword Recommendations is turned on' do
+    before do
+      allow(Mmt::Application.config).to receive(:gkr_enabled).and_return(true)
```
ryanmiller-1, line 255: If test is true in the config, why do we need these?

spec/models/keyword_recommendation_spec.rb
```diff
@@ -0,0 +1,44 @@
+
+# cannot create recommendation by itself
+# draft can create a recommendation
+# template can create a recommendation
+
+# in proposal mode, proposal does not create a recommendation ?????
+
+describe KeywordRecommendation do
```
ryanmiller-1, line 8: Did you consider a test for cascade delete? I'm leaning towards "it's testing rails", but am open to others' thoughts.

app/concerns/controlled_keywords.rb
```diff
@@ -46,6 +46,17 @@ def set_science_keywords
...
+    abstract = get_resource.draft.fetch('Abstract', '')
+    response = cmr_client.fetch_keyword_recommendations(abstract)
+    @keyword_recommendations = if response.success?
+                                 response.body['recommendations'].map { |rec| rec['keyword'] }
+                               else
+                                 []
```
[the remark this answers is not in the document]
charleshuang80: I think we should be logging the recommendations when we get them, so I would prefer that we also add the logging for  failures at that time

app/models/collection_draft.rb
```diff
@@ -1,4 +1,8 @@
 class CollectionDraft < Draft
+  # TODO: we currently allow one to be created, but may allow more in the future
+  # should we just make this has_one for now?
```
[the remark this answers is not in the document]
charleshuang80: I don't know about other changes to the table for future functionality or logging. But I don't think there are any table changes necessary to changing the association, I think the changes are only necessary in the models. I think we would need to change:
* obviously the association type from `has_many` to `has_one`
* singularize the association name (to `keyword_recommendation`)
* singularize the association methods (same as above), and I believe we would need to expect the object rather than an ActiveRecord collection/array
* of course change all this back if/when we are going to present recommendations multiple times for a given record

If we were certain that we will be presenting recommendations multiple times in the future, I would not even have added the comment/question. But it seems unclear to me if we will or not and I don't know which is the more likely outcome.

app/views/collection_drafts/forms/_descriptive_keywords.html.erb
charleshuang80: what do you mean by trimming, something like `strip`?

app/views/collection_drafts/forms/_descriptive_keywords.html.erb
charleshuang80: I will consider this. Doing it before getting to the view would certainly prevent one of the edge cases we discussed in the morning (if all recommendations match previously selected keywords). I would prefer to do it in a concern but at this point I think the helper methods are scattered in a few different locations.

spec/models/collection_draft_spec.rb
charleshuang80: I don't know that it will, but if the previous context block gets run first (mocking it to be `false`), it could interfere with what is in the actual config, so this is just to make sure we have the setting we want.

spec/models/keyword_recommendation_spec.rb
charleshuang80: I did not consider testing it, but it might be a good idea. I think it is also 'testing rails' a bit, but couldn't hurt. It should probably go in the other side of the associations though.

app/views/collection_drafts/forms/_descriptive_keywords.html.erb
charleshuang80: I think I implemented this

app/views/collection_drafts/forms/_descriptive_keywords.html.erb
charleshuang80: moved this around, see if it is along the lines of what you meant

app/concerns/controlled_keywords.rb
charleshuang80: moved

spec/models/collection_draft_proposal_spec.rb
charleshuang80: removed

spec/models/keyword_recommendation_spec.rb
charleshuang80: added tests

app/controllers/collection_drafts_controller.rb
charleshuang80: removed

ryanmiller-1 (approved):

app/views/collection_drafts/forms/_descriptive_keywords.html.erb
ryanmiller-1: 👍

app/controllers/collection_drafts_controller.rb
```diff
@@ -473,15 +481,29 @@ def edit_view_setup
...
+    set_projects                if @form == 'acquisition_information'
+    set_temporal_keywords       if @form == 'temporal_information'
+    set_location_keywords       if @form == 'spatial_information'
+    set_data_centers            if @form == 'data_centers' || @form == 'data_contacts'
+    load_data_contacts_schema   if @form == 'data_contacts'
+  end
+
+  def reconcile_recommendations(keyword_recommendations)
+    keyword_recommendations.reject do |recommendation|
+      get_resource.draft['ScienceKeywords'].any? do |science_keyword|
```
ChristianTrummer99, line 498: getting `nil` error on `get_resource.draft['ScienceKeywords'].any?`

charleshuang80: fixed

app/models/proposal_keyword_recommendation.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  before_create :proposal_mode_enabled?, :gkr_enabled?
+  before_save :proposal_mode_enabled?, :gkr_enabled?
+  # Need to bypass this before action to delete when publishing while in dev
+  before_destroy :proposal_mode_enabled? unless Rails.env.development? || Rails.env.test?
+
+  private
+
+  def proposal_mode_enabled?
+    throw(:abort) unless Rails.configuration.proposal_mode
```
ChristianTrummer99, line 14: another way to write this could be 
```
def proposal_mode_enabled?
     Rails.configuration.proposal_mode || throw(:abort)
end
```
But I am not entirely sure. If `throw(:abort)` ends the method execution (and `true` is not returned) then my suggestion would work. If the `throw(:abort)...` line is executed and the method still returns the last line = `true` (independent of the result of the first line) then my suggestion would not work. Either way, up to you

app/models/proposal_keyword_recommendation.rb
```diff
@@ -0,0 +1,22 @@
...
+    true
+  end
+
+  def gkr_enabled?
+    throw(:abort) unless Rails.configuration.gkr_enabled
```
ChristianTrummer99, line 19: Same as above.

ChristianTrummer99 (approved):
Left feedback on style things.  I couldn't find any bugs/issues when tinkering with the UI locally - all specs passed locally.

PR #674, opened 2020-12-03
Mmt 2410   As GKR, I want MMT to log users keyword selections (Implementation)

jdf-eos:
Added GKR Logging package.
Updated db tables for keyword_recommendations and proposal_keyword_recommendations to hold GKR request_Id and recommendations.
Add spec tests for GKR Logging.


app/controllers/collection_drafts_controller.rb
```diff
@@ -139,7 +146,7 @@ def publish
       render :show
     end
   end
-
+
```
jdf-eos, line 149: clear spaces

app/controllers/collection_drafts_controller.rb
```diff
@@ -74,9 +74,7 @@ def update
 
       if gkr_enabled? && params[:recommended_keywords_viewed] == 'true'
         # create keyword_recommendation
-        get_resource.record_recommendation_provided
-
-        # TODO: if/else and logging
+        get_resource.record_recommendation_provided(params[:gkr_request_id], params[:recommended_keyword_list])
```
ryanmiller-1, line 77: Do we want to record the recommendation when we error?

charleshuang80, line 77: So the recommendations that are stored are what get logged, and these are the recommendations that are in the hidden input field. The recommendations in the hidden input are after they are 'reconciled', so after removing what was already stored in the draft previously. Is that what we want to log, or do we want to log all of the recommendations that were received?

app/concerns/gkr_keyword_recommendations.rb
```diff
@@ -2,14 +2,40 @@
...
-                                []
-                              end
-    # TODO log recommendations (and failure)
+    
+    if response.error?
+      log_gkr_request_failed(user, provider, abstract, response)
+      return {id: "",recommendations: []}
+    end
+      
+    keyword_recommendations = response.body['recommendations'].map { |rec| rec['keyword'] }
```
charleshuang80, line 15: It looks like the first time we are logging (successful) recommendations is on save, not when we first request and receive the recommendations. Is that intended? I thought we were going to log when we first received the recommendations as well but maybe I am mistaken.

ryanmiller-1: Do we want to log recommendations we don't save?

app/controllers/collection_drafts_controller.rb
ryanmiller-1: I think we want to log the pre-reconciled list.

app/controllers/collection_drafts_controller.rb
```diff
@@ -89,6 +87,11 @@ def update
       when 'Save'
         # tried to use render to avoid another request, but could not get form name in url even with passing in location
         get_resource_form = params['next_section']
+        if get_resource_form == 'descriptive_keywords' &&  get_resource.gkr_logging_active?
```
ryanmiller-1, line 90: This is only capturing the 'save' action.  All of this tree is actually saving the draft.  This should be moved into the GKR block above.

spec/features/collection_drafts/gkr_logging_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  context 'saving GKR recommendations' do
+    it 'logs the Save operation' do
+      allow(Rails.logger).to receive(:info)
+      expect(Rails.logger).to receive(:info).with(start_with('GkrLog: type: SAVE'))
+      within '.nav-top' do
+        click_on 'Save'
+      end
+    end
+    
+    it 'logs the Publish operation' do
```
ryanmiller-1, line 19: This isn't testing publish, it's testing done.

app/concerns/gkr_keyword_recommendations.rb
```diff
@@ -2,14 +2,40 @@
 module GKRKeywordRecommendations
   extend ActiveSupport::Concern
 
-  def fetch_keyword_recommendations
+  def fetch_keyword_recommendations(user,provider)
```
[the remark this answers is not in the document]
jdf-eos: Hm. rubo linter seems to be out to lunch.  No wonder I had all those trailing blanks.

app/controllers/collection_drafts_controller.rb
jdf-eos: I stored the pre-reconciled list in the form (:recommended_keyword_list), so that is what is being logged here.

ryanmiller-1: `hidden_field_tag("recommended_keyword_list", keyword_recommendation_list(@keyword_recommendations))`
is being set: `@keyword_recommendations = reconcile_recommendations(keyword_recommendations)`
so the list is the reconciled list.

app/concerns/gkr_keyword_recommendations.rb
jdf-eos: We fetch the recommendations from gkr when we first build the descriptive_keywords form. If we store them at that point, the user may close the window, walk away for lunch, etc. and will never see the recommendations.   If we store them on save, we know that they have acknowledged that they saw them.

app/concerns/gkr_keyword_recommendations.rb
```diff
@@ -2,14 +2,40 @@
...
+    
     response = cmr_client.fetch_keyword_recommendations(abstract)
-    keyword_recommendations = if response.success?
-                                response.body['recommendations'].map { |rec| rec['keyword'] }
-                              else
-                                []
-                              end
-    # TODO log recommendations (and failure)
+    
+    if response.error?
```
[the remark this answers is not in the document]
jdf-eos: I will make this consistent with other places.  However,  I'll first point out that:

1. MMT code seems to consistently avoid this structure.  Why use it for this case?
2. Early return on error conditions makes the code cleaner and more readable.
3. Obfuscation is the worst form of complexity.

app/controllers/collection_drafts_controller.rb
jdf-eos: This is code from an early commit.

app/concerns/gkr_keyword_recommendations.rb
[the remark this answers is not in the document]
ryanmiller-1: I'm not sure MMT is consistent about how it handles this.  In many places in the controllers we do things like:
```
if <do something that saves something to our db>
  (success things)
else
  (error things)
end
```
but I can find plenty of places where we use guards as John did here. I agree with John's argument that this creates more readable code.  My primary support there would be the fact that this reduces the likelihood of nesting conditions, especially with multiple possible failure conditions (something I ran into when doing some of the dMMT work).

app/controllers/collection_drafts_controller.rb
jdf-eos: Yes, that is incorrect.

app/concerns/gkr_keyword_recommendations.rb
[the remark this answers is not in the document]
charleshuang80: Maybe this is a discussion we can have after the PR, to come to an agreement and have more consistency. I'm ok with leaving it for now, but I have a response to your points
1. I'm not sure if I agree, can you point me towards some examples? (either of you). I did a quick search for `response.error?` and did not see many examples of returning early. I did see some examples of the if/else starting with `if response.error?` though. I _think_ when we are dealing with CMR responses generally we have the if/else structure.
2. I don't disagree here (but may want to think through it a bit more), I was only going for consistency
3. I would like to hear more about why it is obfuscating

Again, I'm ok with leaving it, but would like to come to an agreement sometime in the near future.

app/concerns/gkr_keyword_recommendations.rb
charleshuang80: @ryanmiller-1 - I _thought_ we wanted to log all recommendations provided, and then we can also log what is saved and then published, to be comprehensive in what we record as we aren't sure what we're sending as feedback just yet.
@jdf-eos - that makes sense.

app/concerns/gkr_keyword_recommendations.rb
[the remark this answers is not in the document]
ryanmiller-1: https://github.com/nasa/mmt/commit/630414ab305783b3678852c2f1df4ff126224644 is a PR that shows the change from `if... else...` to guard clauses

app/controllers/collection_drafts_controller.rb
jdf-eos: No, it shouldn't be recorded.  I added a guard in set_keyword_recommendations to ensure that the hidden params would not be generated in the descriptive_keywords form.

app/concerns/gkr_keyword_recommendations.rb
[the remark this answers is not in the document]
jdf-eos: OK, updated the linter and fixed all of these problems. Thanks!

app/controllers/collection_drafts_controller.rb
jdf-eos: Changed the code to use the original keyword_recommendations in the hiden_field_tag.

app/controllers/collection_drafts_controller.rb
jdf-eos: moved the block out of the "when 'Save'" and into the main routine.

charleshuang80: agree

app/concerns/gkr_keyword_recommendations.rb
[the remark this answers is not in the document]
jdf-eos: I'm pressed for time today, so I'm going to leave the code as is.    We can talk some more about obfuscation after @charleshuang80  finishes his travels.

charleshuang80 (approved):

ryanmiller-1 (approved):

PR #678, opened 2020-12-11
MMT-2428: Draft Forms Add Too Many Required Icons

ChristianTrummer99:
- Removed non-required data levels from the `requiredDataLevels` array
- Wrote tests according to error recreation described in the ticket


app/assets/javascripts/required_fields.coffee
```diff
@@ -44,6 +44,17 @@ $(document).ready ->
 
   # Add required icons
   addRequiredIcons = (field) ->
+
```
ChristianTrummer99, line 47: I'd like @charleshuang80 opinion on this. Since I believe you wrote this file, you probably understand the nuances here better than I do. To briefly explain, here is the ticket's instructions to recreate the error:
1. Check both Temporal and Variable Subsets and select a boolean value for both
2. Uncheck both, re-check them, and select a value for just one of them

I figured out the reason this was happening was that the data level of the "Allow Multiple Values" Radio groups behind the Temporal and Variable Subsets checkboxes (and Spatial Subsets, I recreated the same error in the radio groups within Spatial Subsets) was remaining in the `requiredDataLevels` array even after it's value was cleared (by unchecking the parent checkbox field). This error appears to occur only for the `checkbox-dependent-field`s. And I believe the reason is because checkboxes aren't handled in the required_fields.coffee file (which is why I added `input[type="checkbox"]` to one of the below lines. 

That being said, I'm not sold that this is the simplest solution - it feels too specific as I had to call out a specific class in the conditional below. Any thoughts/suggestions?

charleshuang80: So my suspicion was that the fields needed to be removed from `requiredDataLevels`. I think that this accomplishes that, but I think we should not have this code in this function, we may want to either make removing the fields a separate function to be called, or to have a separate listener for checkboxes. I am leaning towards the latter, but would want to make sure having a separate listener would still play well with everything else.
I don't think that calling out specific classes/ids/fields is problematic if they have specific and different behavior, but we would want to document it well, which you have done.

charleshuang80 (approved):

william-valencia (approved):

PR #680, opened 2020-12-24
MMT-2425: As a MMT User, I Want To Select My Related URL > Get Data > Data Format Values From a Pick-List Using Values From KMS

ChristianTrummer99:
- updated schema 
- implemented KMS keywords for Get Data > Format dropdown
- db migration - deletes Get Data form if the Format is not recognized against either the KMS or old schema enum; if Formats are equal ignoring case, the migration adjusts the keyword's case. 
- changed/updated specs to reflect new KMS keywords

### DIFF
--- a/app/helpers/related_urls_helper.rb
+++ b/app/helpers/related_urls_helper.rb
@@ -424,4 +424,10 @@ module RelatedUrlsHelper
     ['Simple Subset Wizard (SSW)', 'SIMPLE SUBSET WIZARD (SSW)'],
     ['Subsetter', 'SUBSETTER']
   ]
+
+  def kms_granule_data_formats
+    response = cmr_client.get_controlled_keywords('granule_data_format').body['short_name']
+    response.map! { |h| h['value'] }
+    response.sort << 'Not provided' unless response.include?('Not provided')
+  end
 end
--- a/app/helpers/related_urls_helper.rb
+++ b/app/helpers/related_urls_helper.rb
@@ -424,10 +424,4 @@ module RelatedUrlsHelper
     ['Simple Subset Wizard (SSW)', 'SIMPLE SUBSET WIZARD (SSW)'],
     ['Subsetter', 'SUBSETTER']
   ]
-
-  def kms_granule_data_formats
-    response = cmr_client.get_controlled_keywords('granule_data_format').body['short_name']
-    response.map! { |h| h['value'] }
-    response.sort << 'Not provided' unless response.include?('Not provided')
-  end
 end
--- a/db/migrate/20201221134953_umm_c_15_4_to_15_5.rb
+++ b/db/migrate/20201221134953_umm_c_15_4_to_15_5.rb
@@ -1,0 +1,41 @@
+class UmmC154To155 < ActiveRecord::Migration[5.2]
+  def schema_enum
+    ["ascii", "binary", "GRIB", "BUFR", "HDF4", "HDF5", "HDF-EOS4", "HDF-EOS5", "jpeg", "png", "tiff", "geotiff", "kml", "Not provided"]
+  end
+
+  def kms_enum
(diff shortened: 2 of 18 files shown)


spec/features/collection_drafts/forms/invalid_picklists_spec.rb
```diff
@@ -377,15 +377,17 @@
     it 'displays a summary error' do
       within '.summary-errors' do
         expect(page).to have_content('URL Content Type value [badcontenttype] does not match a valid selection option')
-        expect(page).to have_content('Format value [badformat] does not match a valid selection option')
+        # MMT-2425: using kms Format values, means there is no validation
+        # expect(page).to have_content('Format value [badformat] does not match a valid selection option')
```
ChristianTrummer99, line 381: I'm not sure if these lines should be altogether removed or commented out and explained (like so)

spec/support/draft_helpers.rb
```diff
@@ -185,7 +185,7 @@ def add_related_urls(type: nil, button_type: nil)
               fill_in 'URL', with: 'https://example.com/'
 
               # Get Data fields
-              select 'ascii', from: 'Format'
```
ChristianTrummer99, line 188: explanation: `'ascii'` is not contained in the 'test' KMS keywords response (which is necessary to select it from the picklist as is done in the spec)

ChristianTrummer99:
It seems I need to update the business rules according to the [ticket Description](https://bugs.earthdata.nasa.gov/browse/#descriptionmodule_heading). I may need help doing this as I've never done it before.

app/helpers/related_urls_helper.rb
```diff
@@ -424,4 +424,10 @@ module RelatedUrlsHelper
     ['Simple Subset Wizard (SSW)', 'SIMPLE SUBSET WIZARD (SSW)'],
     ['Subsetter', 'SUBSETTER']
   ]
+
+  def kms_granule_data_formats
+    response = cmr_client.get_controlled_keywords('granule_data_format').body['short_name']
```
charleshuang80, line 429: take a look at app/concerns/controlled_keywords.rb. we should be checking for/handling a potential error response from the keywords endpoint

db/migrate/20201221134953_umm_c_15_4_to_15_5.rb
```diff
@@ -0,0 +1,41 @@
...
+        format = related_url.dig('GetData', 'Format')
+        next unless format && !enum.include?(format)
+        # delete the entire GetData key so as not to trigger conditionally required fields errors by only deleting the Format
+        if enum.none? { |val| val.casecmp?(format) }
+          related_url.delete('GetData')
+          next
+        end
+        # At this point, the only situation that could occur is a typecase discrepancy
+        enum.each { |val| related_url['GetData']['Format'] = val if val.casecmp?(format) }
+      end if draft['RelatedUrls'].present?
```
charleshuang80, line 29: I think you should wrap the `if` around the `each` block here. I would need to dig a bit to see what rubocop or the style guide recommends, but I think you only really want to put conditionals after blocks if they're on one line.

charleshuang80: But also, this is minor and in a migration that probably won't be touched again so don't change this unless you are making other changes.

spec/features/collection_drafts/forms/invalid_picklists_spec.rb
charleshuang80: You can remove the lines. And also, look at the factory it is using, if this value is being populated there you should just remove it.

spec/migrations/umm_c_15_4_to_15_5_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+      template['RelatedUrls'][2] = base_related_url.merge('GetData' => getdata_hash_binary_format_up)
+      template['RelatedUrls'][3] = base_related_url.merge('GetData' => getdata_hash_geotiff_format_up)
+      template['RelatedUrls'][4] = base_related_url.merge('GetData' => getdata_hash_hdf4_format)
+      template['RelatedUrls'][5] = base_related_url.merge('GetData' => getdata_hash_binary_format_down)
+      template['RelatedUrls'][6] = base_related_url.merge('GetData' => getdata_hash_geotiff_format_down)
+      template['RelatedUrls'][7] = base_related_url.merge('GetData' => getdata_hash_different_format_down)
+      @template.save
+    end
+
+    after do
```
charleshuang80, line 48: I am surprised it is necessary to have this after block to remove the records. I would think rspec would handle that for us.

ChristianTrummer99: I just followed what I saw a few other specs do. Haven't really done any research on it, but I can take a look and maybe make a low-priority ticket to clear out after blocks that do this.

spec/migrations/umm_c_15_4_to_15_5_spec.rb
```diff
@@ -0,0 +1,205 @@
+require File.join(Rails.root, 'db', 'migrate', '20201221134953_umm_c_15_4_to_15_5')
+
+# Testing cases where:
```
charleshuang80, line 3: I actually really like these comments, they helped to quickly/easily grasp was scenarios were being tested.

charleshuang80 (approved):

jdf-eos (approved):

charleshuang80 (changes requested):
UMM-C version needs to be updated

charleshuang80 (approved):

william-valencia (approved):

PR #684, opened 2021-01-13
Mmt 2261 alt

jdf-eos:


william-valencia (approved):

config/initializers/content_security_policy.rb
```diff
@@ -1,33 +1,33 @@
...
+    policy.img_src 'self', '*.earthdata.nasa.gov', '*.sit.earthdata.nasa.gov', '*.uat.earthdata.nasa.gov', 'https://fonts.googleapis.com'
+    policy.manifest_src 'self'
+    policy.media_src 'self'
     policy.object_src :none
-    policy.script_src :self, :https, :http
-    policy.style_src :self, :https, :http
-    policy.worker_src :self, :https
-    
+    policy.script_src 'self'
+    policy.script_src 'self', '*.earthdata.nasa.gov', '*.sit.earthdata.nasa.gov', '*.uat.earthdata.nasa.gov'
```
charleshuang80, line 26: I'm generally ok with everything, just waiting to see if we can/need to be more specific here

jdf-eos: script violations come from:

- https://cdn.earthdata.nasa.gov
- https://cdn.sit.earthdata.nasa.gov
- https://fbm.earthdata.nasa.gov

charleshuang80: the last one took me a few min, but I believe it's the feedback module.

So should we specify those sources, or should we still keep it a little more open? If we make it more closed, I'm wondering how much extra testing we should do and how to know what areas of the application.

jdf-eos: I prefer to specify those sources since we are only logging violations, not enforcing them. I think we should be on the path toward enforcement, and this moves us in that direction.

charleshuang80 (approved):

PR #689, opened 2021-02-02
MM-2365: UMM-V/S/T draft progress circle links do not take users to appropriate field

ChristianTrummer99:
- created `anchor` method to ensure that the correct link is applied to each progress circle by checking specific fields in the schema form
- added tests that checked the links of the progress circles as only the classes were being checked (for incomplete, valid, optional, invalid, required, etc)

### DIFF
--- a/lib/json_schema_form/umm_form.rb
+++ b/lib/json_schema_form/umm_form.rb
@@ -312,8 +312,7 @@ class UmmForm < JsonObj
   def label_id
     value = parsed_json.fetch('key','').split('/').first
     value = parsed_json.fetch('key','').split('/')[1] if value.blank?
-    # the conditionals in this method should only be used to ensure that id's are not given to labels that aren't top level fields
-    p field_title_differs_from_accordion_title?, 'afewasdf'
+    
     if value == top_key && !parsed_json['noLabel'] && field_title_differs_from_accordion_title?
       top_key.underscore.dasherize + '-label'
     else
--- a/lib/json_schema_form/umm_form.rb
+++ b/lib/json_schema_form/umm_form.rb
@@ -308,11 +308,10 @@ class UmmForm < JsonObj
     accordion_titles.none? { |accordion_title| accordion_title == title }
   end
 
-
   def label_id
     value = parsed_json.fetch('key','').split('/').first
     value = parsed_json.fetch('key','').split('/')[1] if value.blank?
-    
+
     if value == top_key && !parsed_json['noLabel'] && field_title_differs_from_accordion_title?
       top_key.underscore.dasherize + '-label'
     else
--- a/lib/json_schema_form/umm_form.rb
+++ b/lib/json_schema_form/umm_form.rb
@@ -316,7 +316,7 @@ class UmmForm < JsonObj
     # value == top_key: checks if the label is a top-level field (we don't want to put label id's on fields that aren't
       # linked via progress circle)
     # !parsed_json['noLabel']: there are no progress circles that link to fields with a truthy 'noLabel' key.
-    # field_title_differs_from_accordion_title?: makes sure the fields that share the same title as their accordion don't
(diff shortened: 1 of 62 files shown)


lib/json_schema_form/umm_form.rb
```diff
@@ -298,6 +298,36 @@ def accordion_id
       top_key.underscore.dasherize
     end
   end
+
```
ChristianTrummer99, line 301: Style feedback for the below changes is appreciated. As I write this I think it'd be helpful to add comments to the explain each argument in the conditionals. Let me know if you think this would help.

lib/json_schema_form/umm_form.rb
```diff
@@ -298,6 +298,36 @@ def accordion_id
...
+  def field_title_differs_from_accordion_title?
```
charleshuang80, line 302: minor quibble and semantic, but I think using `is` or `not` would be more accurate that `differs_from` in the method name.

lib/json_schema_form/umm_form.rb
```diff
@@ -298,6 +298,36 @@ def accordion_id
...
+    value = parsed_json.fetch('key','').split('/')[1] if value.blank?
+
+    if value == top_key && !parsed_json['noLabel'] && field_title_differs_from_accordion_title?
+      top_key.underscore.dasherize + '-label'
+    else
+      nil
+    end
+  end
+
+  def anchor
```
charleshuang80, line 322: It would definitely be helpful to have an explanation of what is going on here

lib/json_schema_form/umm_form.rb
```diff
@@ -298,6 +298,36 @@ def accordion_id
...
+  def field_title_differs_from_accordion_title?
+    accordion_titles = Array.new
+    json_form.parsed_json['forms'].each do |form|
+      form['items'].each { |accordion| accordion_titles << accordion['title'] }
+    end
+
+    accordion_titles.none? { |accordion_title| accordion_title == title }
+  end
+
+  def label_id
```
charleshuang80, line 311: would probably be helpful to have some explanation for this method

spec/features/service_drafts/preview/empty/service_information_preview_spec.rb
```diff
@@ -23,12 +23,13 @@
   end
 
   it 'displays the correct progress indicators for required fields' do
+    anchors = ['name-label','long-name-label','type-label','version-label','description-label']
```
charleshuang80, line 26: In general, I don't have issues with running the test through a list, though it can create some issues. I think that you should put this into a `let` block (https://rspec.rubystyle.guide/#let-blocks), and you can probably use `%w[]` since the anchors don't have spaces (I'm surprised rubocop isn't yelling at you about it). You should make these changes to all the places this is being done.

I thought about declaring this in the helper files so you don't need to repeat, but haven't found anything to support doing it that way.

charleshuang80 (approved):

lib/json_schema_form/umm_form.rb
```diff
@@ -298,6 +298,51 @@ def accordion_id
...
+    value = parsed_json.fetch('key','').split('/').first
+    value = parsed_json.fetch('key','').split('/')[1] if value.blank?
+
+    # note: field != form != accordion in the context of the below comment:
+    # value == top_key: checks if the label is a top-level field (we don't want to put label id's on fields that aren't
+      # linked via progress circle)
+    # !parsed_json['noLabel']: there are no progress circles that link to fields with a truthy 'noLabel' key.
+    # field_title_differs_from_accordion_title?: makes sure the fields that share the same title as their accordion don't
```
william-valencia, line 319: Minor comment error. Still mentions old method. Should me field_title_not_accordion_title.

william-valencia (approved):

lib/json_schema_form/umm_form.rb
ChristianTrummer99: good catch.

PR #690, opened 2021-02-09
MMT-2524 Use Constraint progress circle not working correctly

jdf-eos:


app/controllers/collection_drafts_controller.rb
```diff
@@ -192,12 +192,20 @@ def validate_metadata
...
-    errors = Array.wrap(JSON::Validator.fully_validate(@json_schema, get_resource.draft))
+    errors = Array.wrap(JSON::Validator.fully_validate(@json_schema, fix_metadata_fields(get_resource.draft)))
     errors = validate_paired_fields(errors, get_resource.draft)
     errors_before_generate = validate_picklists(errors, get_resource.draft)
     errors_before_generate.map { |error| generate_show_errors(error) }.reject(&:nil?).flatten
   end
 
+  def fix_metadata_fields(meta)
+    if meta.dig('UseConstraints','LicenseURL')
+      meta['UseConstraints']['LicenseUrl']  = meta['UseConstraints']['LicenseURL']
```
ChristianTrummer99, line 203: nitpick - extra space after `['LicenseUrl']`

ChristianTrummer99 (approved):

app/controllers/collection_drafts_controller.rb
```diff
@@ -192,12 +192,20 @@ def validate_metadata
...
     JSON::Validator.register_format_validator('date-time', date_time_format_proc)
 
-    errors = Array.wrap(JSON::Validator.fully_validate(@json_schema, get_resource.draft))
+    errors = Array.wrap(JSON::Validator.fully_validate(@json_schema, fix_metadata_fields(get_resource.draft)))
     errors = validate_paired_fields(errors, get_resource.draft)
     errors_before_generate = validate_picklists(errors, get_resource.draft)
     errors_before_generate.map { |error| generate_show_errors(error) }.reject(&:nil?).flatten
   end
 
+  def fix_metadata_fields(meta)
```
charleshuang80, line 201: I think this fix should be renamed to be more explicit about what it is fixing. Also, I think this should be moved to the model, that's where other fixes like this are.

william-valencia (approved):

app/models/collection_draft.rb
```diff
@@ -94,8 +94,11 @@ def update_draft(params, editing_user_id)
 
       # Convert {'0' => {'id' => '123'}} to [{'id' => '123'}]
       params = convert_to_arrays(params.clone)
-      # Convert parameter keys to CamelCase for UMM
-      json_params = params.to_camel_keys
+
+      # Convert parameter keys to CamelCase for UMM;
+      # amend_license_url_key changes the 'LicenseURL' key to 'LicenseUrl'; the .to_camel_keys method has been patched to return 'LicenseURL'
+      # when 'license_url' is passed which is only valid for Service and Tool metadata; see awrence_patch.rb for details
+      json_params = amend_license_url_key(params.to_camel_keys)
```
charleshuang80, line 101: I think we should leave this line as it originally was, and call the new method in the next line (and move the comments with it) so it doesn't change the original operation but also is clear that it is another step. I know in ruby we often want to combine things into one line, but this is a fairly complicated method that I think it is beneficial to make each operation more discrete.

app/models/collection_draft.rb
```diff
@@ -94,8 +94,14 @@ def update_draft(params, editing_user_id)
 
       # Convert {'0' => {'id' => '123'}} to [{'id' => '123'}]
       params = convert_to_arrays(params.clone)
-      # Convert parameter keys to CamelCase for UMM
+
       json_params = params.to_camel_keys
+
+      # Convert parameter keys to CamelCase for UMM;
```
charleshuang80, line 100: I think you accidentally moved this comment also.

ChristianTrummer99: yes. good catch, thanks.

charleshuang80 (approved):

PR #693, opened 2021-03-05
MMT-2463 update ruby version

william-valencia:
Using an older version of mini_racer that doesn't require an updated GCC

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -2,6 +2,3 @@
 --format documentation
 --color
 --require 'rails_helper'
-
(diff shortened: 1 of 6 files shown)


ChristianTrummer99 (approved):

william-valencia (commented):
Changes for skipping test look good to me. Can't approve since I'm the author but since Charles made the change I will just put my approval in writing.

charleshuang80 (approved):

PR #694, opened 2021-03-08
Mmt 2532

jdf-eos:


app/assets/javascripts/draft_form_selectors.coffee
```diff
@@ -322,6 +322,34 @@ $(document).ready ->
...
+      when 'LicenseText'
+        $parent.find('.license-text-fields').show()
+        $parent.find('.license-text-fields').find('label').addClass('eui-required-o')
+        $parent.find('.license-url-fields').find('input, select').val ''
+      when 'LicenseUrl'
+        $parent.find('.license-url-fields').show()
+        $parent.find('.license-text-fields').find('textarea').val ''
+
+  # Handle License Text required icon
+  $('#draft_use_constraints_license_text').change ->
```
charleshuang80, line 347: I was asking about his because we have a separate file that handles the required icons, and we _should_ try and have it all work through that functionality. I think this can be removed, with a suggestion to the license text field below

app/views/collection_drafts/forms/fields/_use_constraints_fields.html.erb
```diff
@@ -1,46 +1,63 @@
...
-    validate: true,
-    required: true
-   ) %>
-</div>
+    <!-- License Text -->
+    <div class="license-text-fields not-available row sub-fields" style="<%= 'display: none' unless license_text_checked%>">
+      <%= mmt_text_area(
+        name: 'license_text',
+        title: 'License Text',
+        prefix: name_prefix,
```
charleshuang80, line 56: if you change this to `"#{name_prefix}|license_text|_"`, it should alter the data-level enough so it acts similar to the License Url grouping for the one field. if there is data in the field the required icon will appear, if data is removed, the icon goes away.

charleshuang80: After further testing, this change does work for the required icons issue. However, it also creates a problem with the name of the field (and therefore saving and displaying the value). I have a potential solution, but I think it would be best to discuss the best course of action here on a screen share so let me know when you have some time for that.

app/views/collection_drafts/forms/fields/_use_constraints_fields.html.erb
```diff
@@ -1,46 +1,63 @@
-<div class="row sub-fields">
+<div class="license-group">
+  <% license_url_checked =  !use_constraints['LicenseUrl'].nil? %>
+  <% license_text_checked =  !use_constraints['LicenseText'].nil? %>
+  
   <!-- Description -->
   <%= mmt_label(
     name: 'description',
-    title: 'Description',
+    title: ' ',
```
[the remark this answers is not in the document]
jdf-eos: Good catch!   Removed this.

app/assets/javascripts/draft_form_selectors.coffee
jdf-eos: Removed this javascript.   The problem was that the required icon handling required a blur event.  Added the blur triggers to the radio buttons.

app/views/collection_drafts/forms/fields/_use_constraints_fields.html.erb
jdf-eos: I cleaned up some of the html and made this change.

william-valencia (approved):

charleshuang80 (approved):

ChristianTrummer99 (approved):

PR #695, opened 2021-03-09
MMT-2482 Removing email sent to CMR for subscriptions and changing Su…

william-valencia:
…bsribers to Subscriber


app/controllers/subscriptions_controller.rb
```diff
@@ -54,7 +55,6 @@ def edit
   def create
     authorize :subscription
     @subscription = subscription_params
-    @subscription['EmailAddress'] = get_subscriber_email(@subscription['SubscriberId'])
```
charleshuang80, line 57: were you going to remove this from `update` also?

william-valencia: Yeah that's part of line 84. But running some tests so I commented it out. But it will be back.

william-valencia: its back.

charleshuang80 (approved):

ChristianTrummer99 (approved):

jdf-eos (approved):

PR #696, opened 2021-03-10
MMT-2456: Update to UMM-C v 1.16 (Data Access Policy)

ChristianTrummer99:
- Updated Schema
- Created new forms
- Wrote specs to ensure:
        1. drafts can be saved successfully
        2. drafts can be published successfully
        3. progress circle is present and in proper state
        4. metadata preview appears properly
 - added new information to factory data
 - adjusted help modal javascript
 - adjusted javascript that sets `id` and `name` fields for `multiple` elements
 - added a css `-width` class
- added to awrence patch file
- added to awrence patch file


william-valencia (approved):

lib/assets/schemas/collections/umm-cmn-json-schema.json
```diff
@@ -343,7 +343,7 @@
         "properties": {
             "Format": {
                 "description": "The format of the data.",
-                "$ref": "#/definitions/GetDataTypeFormatEnum"
+                "$ref": "#/definitions/GetServiceTypeFormatEnum"
```
charleshuang80, line 346: I'm wondering why this is showing up now, did we miss it before?

spec/migrations/umm_c_1_15_5_to_1_16_spec.rb
```diff
@@ -0,0 +1,33 @@
+require File.join(Rails.root, 'db', 'migrate', '20210312165348_umm_c_1_15_5_to_1_16')
```
charleshuang80, line 1: I think it's good that you did this, to stay in line with the process. I'm wondering, though, for such a simple migration in the future, if it is worth writing the migration tests for.

charleshuang80 (approved):

lib/assets/schemas/collections/umm-cmn-json-schema.json
ChristianTrummer99: I guess so

spec/migrations/umm_c_1_15_5_to_1_16_spec.rb
ChristianTrummer99: Yeah, I'm not really sure. Part of the reason the tests are helpful is just to see if the migration actually works (while I'm writing the migration) because it's just more convenient than actually executing the migration.

PR #702, opened 2021-03-25
MMT-2539 Update to UMM-C forms to v1.16.1 (Multiple DOIs)

william-valencia:
Adding Associated DOIs


ChristianTrummer99 (approved):

lib/assets/schemas/collections/umm-c-merged.json
```diff
@@ -123,7 +131,7 @@
       "minItems": 0
     },
     "ISOTopicCategories": {
-      "description": "Identifies the topic categories from the EN ISO 19115-1:2014 Geographic Information – Metadata – Part 1: Fundamentals (http://www.isotc211.org/) Topic Category Code List that pertain to this collection, based on the Science Keywords associated with the collection. An ISO Topic Category is a high-level thematic classification to assist in the grouping of and search for available collections.  The controlled vocabulary for ISO topic categories is maintained in the Keyword Management System (KMS).",
```
charleshuang80, line 126: nitpick: I don't see this change in the umm-c or the umm-c-cmn file. Could it be a mistake?

charleshuang80 (approved):
Looks good. I think that extra space is minor yet annoying, but if we have time, we should probably remove it to be in line with the actual schema

lib/assets/schemas/collections/umm-c-merged.json
william-valencia: The extra space was in our previous merged file but it is not in the umm-c file. I copied and pasted the entire umm-c file so this one with only one space is correct.

https://git.earthdata.nasa.gov/projects/EMFD/repos/unified-metadata-model/browse/collection/v1.16.1/umm-c-json-schema.json?until=ef61770ea7b5fbc279e897354eda2b55e698aa9d&untilPath=collection%2Fv1.16.1%2Fumm-c-json-schema.json

william-valencia:
> Looks good. I think that extra space is minor yet annoying, but if we have time, we should probably remove it to be in line with the actual schema

The extra space was removed as part of my changes. That is how it is in the umm-c schema.

PR #705, opened 2021-03-30
MMT-2539 Fixing the Doi to DOI

william-valencia:


app/views/collection_drafts/forms/_collection_information.html.erb
```diff
@@ -117,7 +117,7 @@
           values: draft.draft['AssociatedDOIs'] || [{}],
           name_prefix: 'draft_|associated_dois|_',
           multiple: true,
-          options: {} } %>
+          options: { associated_doi_title: 'Associated DOIs' } } %>
```
ChristianTrummer99, line 120: I want to get @charleshuang80 opinion on this, should we be making these changes in the awrence_patch.rb? or like this^ ? It's tricky bc there's precedent for both

jdf-eos (approved):

app/views/collection_drafts/forms/_collection_information.html.erb
william-valencia: I think the awrence_patch is for camelization. I don't think that applies in this case. but we can wait for @charleshuang80 to chime in.

charleshuang80: I am not totally sure myself. @william-valencia, have you tried making the fix in awrence_patch and seeing if it fixes this?
My suspicion is that if this was the only place that was doing it wrong, then the other locations where it might be an issue were already being caught by what is in the awrence_patch (but that's why a test would help). And if that is the case then this would still be necessary. And as Christian said, there is precedent for needing this change.

charleshuang80 (approved):
I am approving<s>, pending a test just to verify what was said in the comment thread<s>.

PR #707, opened 2021-03-31
MMT-2544: merge feature branch for UMM-C 1.16 and 1.16.1 changes into master

jdf-eos:


app/helpers/preview_circles_helper.rb
```diff
@@ -31,6 +31,10 @@ module PreviewCirclesHelper
         required: false,
```
ChristianTrummer99, line 31: this should be true right?

william-valencia (approved):

ChristianTrummer99 (approved):

app/helpers/preview_circles_helper.rb
ChristianTrummer99: I see that this fix is included in MMT-2551, so ignore

charleshuang80 (approved):

PR #727, opened 2021-05-04
MMT-2488:  SortSite Level A: MMT form has fields without LABEL elements or TITLE attributes

ChristianTrummer99:
- added `title` attribute to each field that was lacking a `label` tag or `title` attribute
- adjusted spec

### DIFF
--- a/app/views/manage_collections/show.html.erb
+++ b/app/views/manage_collections/show.html.erb
@@ -22,7 +22,8 @@
                   options_for_select(@templates),
                   prompt: 'Select Template Name',
                   class: 'template-select',
-                  style: 'display:none;'
+                  style: 'display:none;',
+                  title: 'Template Name'
                 ) %>
                 </p>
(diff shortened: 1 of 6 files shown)


william-valencia (approved):

app/views/manage_collections/show.html.erb
```diff
@@ -22,7 +22,8 @@
                   options_for_select(@templates),
                   prompt: 'Select Template Name',
                   class: 'template-select',
-                  style: 'display:none;'
+                  style: 'display:none;',
```
cgokey, line 25: why add the display:none just curious?

ChristianTrummer99: I actually only added the comma, so it shows up as a change

cgokey: Ah ok, I missed that, thanks.

cgokey (approved):

PR #734, opened 2021-05-18
MMT-2534: Gracefully handle error response from Legacy Services endpoints (potentially because of timeout) 

ChristianTrummer99:
- adjusted controller
- added handling for HTML error body
- added logging and flash messages for all echo soap calls that lacked them (there were some echo soap calls to which I only added logging for specific reasons)


jdf-eos (approved):

jdf-eos (approved):

jdf-eos:
Its a little ugly that the flash error message is duplicated in so many places.   But, I don't feel strongly about it so I approved the PR.

cgokey (approved):

charleshuang80:
> Its a little ugly that the flash error message is duplicated in so many places. But, I don't feel strongly about it so I approved the PR.

ChristianTrummer99:
> > Its a little ugly that the flash error message is duplicated in so many places. But, I don't feel strongly about it so I approved the PR.
> 
> I agree, it might be nice to try and refactor the flash message to be a bit cleaner
> 
> Sorry, I did not mean to close the PR, I was trying to respond to John's comment, not sure what happened.

I agree as well. The only issue we run into trying to refactor the flash message is that it becomes difficult to provide detail to the user about which request caused the timeout. For instance, on the Create Service Option Assignments page if the request for Service Options times out (and the Service Options select2 field is empty) there would be a flash message saying that we're unable to retrieve Service Options at this time along with the `request.uuid`. Do you think I should try to make this more programmatic and assign the general part of the message as a variable/method?

ie. something like this *in the controller
```
flash[:error] = "#{request.uuid} - #{generic_message('service options')}" ```

PR #741, opened 2021-05-31
GKR-30  Handle error codes (500-505) returned by NGAP LB when the GKR…

jdf-eos:
… is down.

### DIFF
--- a/app/concerns/gkr_keyword_recommendations.rb
+++ b/app/concerns/gkr_keyword_recommendations.rb
@@ -7,12 +7,22 @@ module GKRKeywordRecommendations
     response = cmr_client.fetch_keyword_recommendations(abstract)
 
     if response.error?
+      if response.status >= 500 && response.status <= 505    # NGAP LB can't connect to GKR.
+        log_gkr_comm_error(user, provider, abstract, response.status, 'Communication failure with GKR')
+        return { 'error': 'CommError' }
+      end
+
       log_gkr_request_failed(user, provider, abstract, response)
-      { id: '', recommendations: [] }
+      return { id: '', recommendations: [] }
     end
 
     keyword_recommendations = response.body['recommendations'].map { |rec| rec['keyword'] }
     { id: response.body['uuid'], recommendations: keyword_recommendations }
+  end
+
+  def log_gkr_comm_error(user, provider, abstract, status, reason)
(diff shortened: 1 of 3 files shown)


app/concerns/gkr_keyword_recommendations.rb
```diff
@@ -7,14 +7,24 @@ def fetch_keyword_recommendations(user, provider)
...
       log_gkr_request_failed(user, provider, abstract, response)
-      { id: '', recommendations: [] }
+      return { id: '', recommendations: [] }
     end
 
     keyword_recommendations = response.body['recommendations'].map { |rec| rec['keyword'] }
     { id: response.body['uuid'], recommendations: keyword_recommendations }
   end
 
+  def log_gkr_comm_error(user, provider, abstract, status, reason)
```
cgokey, line 23: Can we add a request id # to this?  So we can match up these log statements to the user's error displayed

cgokey: See ProviderOrdersController.rb, it looks like they just use the request object to grab the uuid,  i.e.
    flash[:alert] = "The order request #{request.uuid} timed out showing the order. Please try again or contact #{view_context.mail_to('support@earthdata.nasa.gov', 'Earthdata Support')}."

jdf-eos: Currently, the GKR API creates a request id as it returns the keywords.   In this case,   we haven't yet communicated with the API, so we don't know what the request id will be.   In the successful cases, the log messages will log the request id, and it will be stored in the db.

All of this violates our security control which says the client will create the request id and send it through a standard header.  I believe we have a separate ticket to fix this, but it will require changes by the impact team as well as  mmt.

william-valencia (approved):

app/concerns/gkr_keyword_recommendations.rb
```diff
@@ -7,14 +7,24 @@ def fetch_keyword_recommendations(user, provider)
...
+    Rails.logger.info("GkrLog: type: FAILED - date: #{Time.new} - env: #{Rails.env} - user_id: #{user} - provider: #{provider}"\
```
cgokey, line 24: Can we include there request id that we are displaying to the user also in the log statements?   This way we can match them up?

jdf-eos: Ouch.  I should have caught that.  Will do.

cgokey (approved):

PR #744, opened 2021-06-04
MMT-2555: update to UMM-C v 1.16.3 - add new enums for Related URL Type, Related URL Subtype, and URL Mime Type

charleshuang80:
* add new enums
* update helpers so enums show up
* update js for correct Related URL behavior with new Type enum
* write migration

### DIFF
--- a/app/assets/javascripts/related_urls.coffee
+++ b/app/assets/javascripts/related_urls.coffee
@@ -57,6 +57,9 @@ $(document).ready ->
 
     if typeValue?.length > 0
       switch typeValue
+        when 'GET CAPABILITIES'
+          $parent.find('.get-data-fields').show()
+          $parent.find('.get-service-fields').find('input, select').val ''
         when 'GET DATA'
           $parent.find('.get-data-fields').show()
           $parent.find('.get-service-fields').find('input, select').val ''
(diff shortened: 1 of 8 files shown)


charleshuang80:
I looked into and considered if we need to add a migration test or make any updates to the existing collection drafts forms tests. I decided against adding or updating tests, but if anyone feels strongly in the other direction, please let me know. Here's why I didn't think we needed the migration test or to update the other tests:
For the migration test, we only have a down migration. I tested it locally to make sure it worked, I encourage others to do the same. Also, the previous migration with only a down migration did not have a migration test written. Therefore, with it working and the likelyhood of the down migration being needed (I don't think we've had to roll one back to my knowledge) I felt we did not need a migration test.
For the other forms tests, we do not have full coverage of all possibilities for Related URLs or where all Online Resources are tested. For Related URLs, we don't have a test for the Use Service API option, so I thought we did not need one for Get Capabilities. And for the Mime Type, I did not see the mime type being used in tests for use constraints, publication reference, or collection citations, so did not feel the need to check for the new enum.

htranho (approved):

jdf-eos (approved):

william-valencia (approved):

PR #745, opened 2021-06-08
Mmt 2587 operations user should be able to view all groups in MMT permissions section

jdf-eos:


ChristianTrummer99 (approved):

cgokey (approved):

william-valencia (approved):

app/controllers/manage_cmr/groups_controller.rb
```diff
@@ -21,18 +21,23 @@ def index
...
     # as a user cannot have CMR in their available_providers we have to determine when to add it to the filter drop down
     if @filters['provider'].present?
       @query['provider'] = @filters['provider']
       # if provider is CMR, it should be an option in the select
-      @groups_provider_ids << 'CMR' if @filters['provider'].include?('CMR')
     elsif @filters['provider_segment'] == 'current'
       @query['provider'] = [current_user.provider_id]
     elsif @filters['provider_segment'] == 'available'
-      @query['provider'] = @groups_provider_ids
+      @query['provider'] = current_user.available_providers.dup
```
charleshuang80, line 40: The way I had thought of doing this if the user had system group permissions was to just remove the provider parameter from the query. That should return all groups the user['s token] has permission to see. Have you tried that, or think it would work?

spec/features/manage_cmr/groups/groups_list_spec.rb
```diff
@@ -62,7 +62,6 @@
               within '.groups-table' do
                 expect(page).to have_no_content('Administrators CMR Administrators CMR 2')
                 expect(page).to have_no_content('Administrators_2 The group of users that manages the CMR. CMR')
-                expect(page).to have_no_content('SEDAC Admin Group Test group for provider SEDAC')
```
charleshuang80, line 65: Is this test failing? The way the ACLs for the group are set up, I think that only the admin user can see this group, which is why it's grouped with the system groups for the test

app/controllers/manage_cmr/groups_controller.rb
jdf-eos: Talked with Rosy about this and we both believe this is a better way.

spec/features/manage_cmr/groups/groups_list_spec.rb
jdf-eos: Reverted.

spec/features/manage_cmr/groups/groups_list_spec.rb
```diff
@@ -237,7 +236,7 @@
```
charleshuang80, line 236: Looking at the comment on line 191, it seems like maybe we should remove some of the providers in the login line to ensure that the modification for admin users that don't have all the providers still show groups from all the providers. What do you think?

jdf-eos: Yes, I removed 3 of the providers and updated the context description.

charleshuang80 (approved):

PR #747, opened 2021-06-14
MMT-2603: Collections->Data Identification->Publication References: Form validation error even though 'DOI Not Available' is selected and populated

ChristianTrummer99:
- Changed form
- Wrote migration + migration tests

### DIFF
--- a/spec/migrations/fix_publication_doi_fields_spec.rb
+++ b/spec/migrations/fix_publication_doi_fields_spec.rb
@@ -36,7 +36,31 @@ describe 'Migration tests for Publication DOIs fix' do
 
     it 'removes the DOI field from the second Publication Reference but not the first' do
       expect(CollectionDraft.find(@draft.id).draft.dig('PublicationReferences',0,'DOI')).to eq(available_doi)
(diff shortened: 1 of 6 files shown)


cgokey (approved):

spec/migrations/fix_publication_doi_fields_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      FixPublicationDoiFields.new.change
+    end
+
+    after do
+      CollectionDraft.delete([@draft.id])
+      CollectionTemplate.delete([@template.id])
+    end
+
+    it 'removes the DOI field from the second Publication Reference but not the first' do
+      expect(CollectionDraft.find(@draft.id).draft.dig('PublicationReferences',0,'DOI')).to eq(available_doi)
```
william-valencia, line 38: Shouldn't it be removed from both the template and the draft? Why is it only removed from the template and not the draft? shouldnt 1 on both be nil?
Should we be checking that 0 on both is as it was before and 1 on both is now nil?

william-valencia (approved):

PR #752, opened 2021-06-23
MMT-2543: For Collections->Descriptive Keywords->Additional Attributes, Value is not being Validated against Data Type

ChristianTrummer99:
- added javascript validation for Additional Attribute "Value" and "Data Type" fields
- added Rails record validation for Additional Attribute "Value" and "Data Type" fields so the progress circles work properly
- tests


ChristianTrummer99:
I have questions for reviewers:
- Firstly, are these changes worth it? It seems like a minor issue and there are quite a large # of changes needed to support this validation 
- I don't like the idea of hardcoding this kind of validation in with regex - would much rather do an ajax to cmr, but cmr will only validate this on ingestion, and idk if that's a worthwhile route to take
- How do you feel about the placement of the method `validate_value_data_type` in the `validate_paired_fields` method (in collection_drafts_controller.rb)? Would the fields I'm validating be considered "paired fields"?
- Is the placement of the regex constants appropriate? (in manage_metadata_controller.rb) 
- My regex closely approximates the CMR ingestion criteria. Regarding the `DATETIME_REGEX`, there already exists a `DATE_TIME_REGEX` constant that appears to be used for other Datetime fields. However, it is much more strict than CMR is for the specific field in question: Additional Attributes > DataType > DATETIME. Should we be using this anyway?

app/controllers/manage_metadata_controller.rb
```diff
@@ -238,6 +238,15 @@ def get_page(fields)
...
 
+  # The following are used to validate AdditionalAttributes/#{index}/DataType, as
+  # these errors are not captured in the schema, they are business logic being
+  # enforced in the CMR, and so they are being validated in collection_drafts_controller.rb
+  FLOAT_REGEX = /^[+-]?\d+(\.\d+)?([eE][-+]?\d+)?$/
```
htranho, line 244: Are there any tests for bad float and int?

ChristianTrummer99: Yes, they're in there 👍

htranho (approved):

charleshuang80:
My thoughts on your questions:
- If PO/SM have prioritized this work and they think it was worth doing, I would go with that. Or if you think it's too much code to do this, I would double check with them if they still want these changes. If it were acceptable to just have CMR kick back an ingest attempt with the errors I would think the ticket would not have been approved or prioritized.
- I think it is ok to use a regex, as I think it might be hard to just check these fields with CMR (though I haven't researched it). Have you checked with the CMR folks to see if they are using a specific regex that you can copy so it's the same check?
- I think the placement of the new method is fine. I do think the name `validate_value_data_type` is not as clear as it could be. I am ambivalent on whether these could be considered paired fields - I think they could be but that method is doing a specific check (so maybe that method name could be updated to be more explicit?)
- since the regex is not needed by the other metadata types, I would push it down as far as is possible, I _think_ that would be in the collection drafts controller. Also, I don't like the idea of having the regex in 2 separate places - if any changes need to be made it runs the risk of only being changed in one place. I would try passing it to the javascript from the controller so the regex can be in one place.
- without more information I don't know the difference between the regexes. I would maybe check with the PO to see which regex to use.

cgokey (approved):

app/views/collection_drafts/forms/fields/_additional_attribute_fields.html.erb
```diff
@@ -22,6 +22,12 @@
     required: true
    ) %>
 
+  <!-- Passing regex for validation of Value and Data Type fields -->
+  <%= hidden_field_tag 'float-regex-for-validation', CollectionDraftsController::FLOAT_REGEX %>
+  <%= hidden_field_tag 'integer-regex-for-validation', CollectionDraftsController::INT_REGEX %>
+  <%= hidden_field_tag 'date-regex-for-validation', CollectionDraftsController::DATE_REGEX %>
+  <%= hidden_field_tag 'time-regex-for-validation', CollectionDraftsController::TIME_REGEX %>
+  <%= hidden_field_tag 'datetime-regex-for-validation', CollectionDraftsController::DATETIME_REGEX %>
```
ChristianTrummer99, line 30: taken from this [article](http://railscasts.com/episodes/324-passing-data-to-javascript)

app/controllers/collection_drafts_controller.rb
```diff
@@ -282,12 +282,68 @@ def validate_parameter_ranges(errors, metadata)
...
 
   def validate_paired_fields(errors, metadata)
+    errors = validate_additional_attribute_value_field(errors, metadata)
     errors = validate_parameter_ranges(errors, metadata)
     errors = validate_project_paired_dates(errors, metadata)
     errors = validate_temporal_paired_dates(errors, metadata)
     validate_tiling_identification_systems_paired_fields(errors, metadata)
   end
 
+  # The following regex are used to validate AdditionalAttributes/#{index}/DataType, as
```
ChristianTrummer99, line 292: Thomas helped me dig up the validation that CMR uses and I created regex for it. CMR uses Java parsing functions that accepted the formats outlined below. (It would've been nice if there was regex I could copy but the regex I created works)

charleshuang80: I think it is worthwhile/helpful to a future dev to add a small note about CMR using parsing functions and the regex is an approximation of their checks

app/views/collection_drafts/forms/_descriptive_keywords.html.erb
```diff
@@ -145,6 +145,14 @@
   </div>
 </fieldset>
 
+<!-- Passing regex for validation of Value and Data Type fields -->
+<%= hidden_field_tag 'float-regex-for-validation', CollectionDraftsController::FLOAT_REGEX %>
```
charleshuang80, line 149: Not necessarily asking for a change, but just asking: Did you try a script tag? That is the way we pass most data to javascript (see keywords and _I think_ others)

app/assets/javascripts/validation.coffee
```diff
@@ -779,6 +780,54 @@ $(document).ready ->
...
+  # These errors are not captured in the schema, they are business logic being
+  # enforced in the CMR. Since we cannot publish records with these conditions,
+  # we should provide that feedback to the user. Specifically, in the
+  # Additional Attribute the Value supplied by the user needs to conform to the
+  # Data Type the user selects. The checks in the errorPresent block are meant
+  # to APPROXIMATE those of CMR ingestion with the intention that there are
+  # likely some input values that the below logic will permit, and CMR will not,
+  # but never vice versa.
+  validateAdditionalAttributeValueField = (errors) ->
+    # pulling the regex's from the hidden field tags used to share them between
```
charleshuang80, line 792: I don't think this is a big deal, so not worth a lot of time. But I'm wondering if it's worth trying a guard clause to exit out of this method if we're not on the right form and avoid setting up the regex vars for no reason. Just a thought - probably it doesn't save a lot of time

ChristianTrummer99: Ohh good point. I think it's worth doing

app/controllers/collection_drafts_controller.rb
ChristianTrummer99: Will do.

app/views/collection_drafts/forms/_descriptive_keywords.html.erb
ChristianTrummer99: I used this [article](http://railscasts.com/episodes/324-passing-data-to-javascript) and considered the script tag option, but thought it may be more complex as I'm not sure if I should define the regex's as variables, returned from a function, or global variable. Nonetheless, if script tags are the norm for us, I think its worth the change to be consistent. Looking for examples in the repo now

ChristianTrummer99: Yeah I definitely think script tags are the way to go. We use them in a ton of places.

app/views/collection_drafts/forms/_descriptive_keywords.html.erb
```diff
@@ -186,3 +186,13 @@
 <script type="text/javascript">
   setupScienceKeywords(<%= raw @science_keywords.to_json %>);
 </script>
+
+<!-- Passing regex for validation of Value and Data Type fields -->
+<script>
+  var floatRegexString = "<%= Regexp.quote  CollectionDraftsController::FLOAT_REGEX %>";
```
charleshuang80, line 192: nitpick: 1) you probably don't need another script tag, I think you can just stick it in the existing script tag above; 2) if a standalone script tag I would think you need the type attr as the above tag has

app/assets/javascripts/validation.coffee
```diff
@@ -779,6 +780,52 @@ $(document).ready ->
...
+
+    if $('#draft_additional_attributes').length > 0
+      floatRegex = new RegExp floatRegexString
+      intRegex = new RegExp intRegexString
+      boolRegex = new RegExp boolRegexString
+      dateRegex = new RegExp dateRegexString
+      timeRegex = new RegExp timeRegexString
+      dateTimeRegex = new RegExp dateTimeRegexString
+
+    $('#draft_additional_attributes').children('.multiple-item').each (index, element) ->
```
charleshuang80, line 801: another nitpick: I think this probably works fine, if it were me, I'd indent this to be nested under the check you just added.

charleshuang80 (approved):

charleshuang80 (approved):

PR #762, opened 2021-07-21
MMT-2644: Updated drafts controller show to authenticate only by EDL token if client is requesting to download_json

cgokey:


spec/controllers/collection_draft_proposals_controller_spec.rb
```diff
@@ -0,0 +1,11 @@
+describe Proposal::CollectionDraftProposalsController do
+  it 'shows draft collection proposal record for proposal mode' do
+    set_as_proposal_mode_mmt
+    @collection_draft_proposal = create(:full_collection_draft_proposal, user: get_user)
+    allow_any_instance_of(Cmr::UrsClient).to receive(:validate_dmmt_token).and_return(Cmr::Response.new(Faraday::Response.new(status: 200, body: '{"uid":"testuser"}', response_headers: {})))
+    get :show, params: { id: @collection_draft_proposal.id }, as: :json
+    parsed_body = JSON.parse(response.body)
+    draft_json = @collection_draft_proposal.draft
```
ChristianTrummer99, line 8: nitpicks: I'd like to see lines 3-8 in a `before` block as well as a `context` block surrounding the `before` and `it` blocks.

spec/controllers/collection_drafts_controller_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+  before do
+  end
+
+  it 'shows draft collection record for mmt proper' do
+    set_as_mmt_proper
+    @collection_draft = create(:full_collection_draft, user: get_user)
+    allow_any_instance_of(Cmr::UrsClient).to receive(:validate_mmt_token).and_return(Cmr::Response.new(Faraday::Response.new(status: 200, body: '{"uid":"testuser"}', response_headers: {})))
+    get :show, params: { id: @collection_draft.id }, as: :json
+    parsed_body = JSON.parse(response.body)
+    draft_json = @collection_draft.draft
```
ChristianTrummer99, line 11: nitpick: same here, I think it would be nice if 6-11 were in the `before` block and there was a `context` block around the `before` and `it`

app/controllers/collection_drafts_controller.rb
```diff
@@ -28,6 +28,41 @@ def show
...
+
+    if Rails.configuration.proposal_mode
+      token_response = cmr_client.validate_dmmt_token(token)
+    else
+      token_response = cmr_client.validate_mmt_token(token)
+    end
+    if token_response.success?
+      # Verify the user owns the draft
+      json = token_response.body
+      json = JSON.parse json if json.class == String # for some reason the mock isn't return hash but json.
```
cgokey, line 48: I should say isn't returning hash but a json string

spec/controllers/collection_draft_proposals_controller_spec.rb
cgokey: Good point.

spec/controllers/collection_drafts_controller_spec.rb
cgokey: Will change here too.

ChristianTrummer99 (approved):

htranho (approved):

app/controllers/application_controller.rb
```diff
@@ -6,13 +6,13 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   protect_from_forgery with: :exception
 
-  before_action :ensure_user_is_logged_in
+  before_action :ensure_user_is_logged_in, except: [:download_json]
```
charleshuang80, line 9: Other metadata types have a `download_json` method as well, so this would skip in unintended places. Also, I think in general it is better to push down the skipping or excepting of before actions as far down as possible in the controller inheritance chain. So I think you probably should remove the `except`s here and skip the before action in the various collection controllers

PR #764, opened 2021-07-27
Mmt 2644 - Updated drafts controller to download_json of a draft or proposal using authenticated EDL token

cgokey:


app/controllers/collection_drafts_controller.rb
```diff
@@ -28,6 +30,50 @@ def show
...
+
+    authorized = false
+
+    if token_response.success?
+      token_info = token_response.body
+      token_info = JSON.parse token_info if token_info.class == String # for some reason the mock isn't return hash but json string.
+      token_user = User.find_by(urs_uid: token_info['uid']) # the user assoc with the token
+      draft_user = User.find_by(id: get_resource.user_id) # the user assoc with the draft collection record
+
+      authorized = false
```
cgokey, line 56: this should be removed, it is defined above already

ChristianTrummer99 (approved):

app/controllers/collection_drafts_controller.rb
```diff
@@ -4,11 +4,13 @@ class CollectionDraftsController < BaseDraftsController
   include ControlledKeywords
   include CMRCollectionsHelper
   include GKRKeywordRecommendations
-  before_action :set_resource, only: [:show, :edit, :update, :destroy, :publish]
+  before_action :set_resource, only: [:download_json, :show, :edit, :update, :destroy, :publish]
   before_action :load_umm_c_schema, only: [:new, :edit, :show, :publish]
   before_action :collection_validation_setup, only: [:show, :publish]
   before_action :verify_valid_metadata, only: [:publish]
   before_action :set_associated_concepts, only: [:show]
+  before_action :ensure_user_is_logged_in, except: [:download_json]
```
charleshuang80, line 12: this should be inherited from ApplicationController, so all that is necessary is to `skip_before_action` for the method you're specifying

cgokey: Ah makes sense, will change.

william-valencia (approved):

charleshuang80 (approved):

PR #766, opened 2021-08-10
hotfix/MMT-2655-1: fix for safari not showing icons and hover-over incorrectly rendering markup

ChristianTrummer99:
The issue was - despite FF and Chrome experiencing no issue displaying the s3 icons in the Chooser widget (using the `::before` pseudo element) - safari was not rendering the `::before` element as this is invalid html: `option` tags are not supposed to have any children and pseudo elements are considered children when nested within `option` tags. The solution was to simply use inline emojis. This seems to be much easier (and is, of course, valid). 

Additionally, to solve the issue of the hover-over incorrectly rendering the markup, a selector change was implemented.


spec/features/chooser/chooser_disable_selected_options.rb
```diff
@@ -32,15 +32,26 @@
```
william-valencia, line 32: Just noticed this file doesn't end with spec.rb,  I think we have a convention wherein all spec tests end with spec.rb

htranho (approved):

cgokey (approved):

PR #771, opened 2021-08-31
MMT-2660: allow Progressive Update - updating a collection with warnings or existing errors

charleshuang80:
* add ability to ingest a collection draft that is an update (revision) of an existing Collection with existing errors or warnings
* add calls to CMR validation endpoint
* add controller method to handle cmr validation call
* update view and js for ajax call to validation endpoint and handling of resulting messages
* add test data and calls to ingest collections with errors for testing purposes
* add tests for updating a collection with existing errors
* add tests for new validation behavior


lib/cmr/cmr_client.rb
```diff
@@ -273,6 +273,28 @@ def ingest_collection(metadata, provider_id, native_id, token, content_type = ni
...
+    # Publish a collection for progressive update feature - CMR allowing an update
+    # to an existing collection that may include some existing errors if it does
+    # not have new errors
+    # this is only for testing as it requires the system token and a special header
+    def ingest_progressive_update_collection(metadata, provider_id, native_id)
+      # if native_id is not url friendly or encoded, it will throw an error so we check and prevent that
+      url = if Rails.env.development? || Rails.env.test?
+              "http://localhost:3002/providers/#{provider_id}/collections/#{encode_if_needed(native_id)}"
+            else
+              "/ingest/providers/#{provider_id}/collections/#{encode_if_needed(native_id)}"
```
charleshuang80, line 285: take out non dev and non test call

spec/support/ingest_helpers.rb
```diff
@@ -73,6 +73,41 @@ def publish_collection_draft(revision_count: 1, include_new_draft: false, provid
...
+        }
+
+        draft = build(:progressive_update_collection_first, draft_attributes)
+        ingest_response = cmr_client.ingest_progressive_update_collection(draft.draft.to_json, draft.provider_id, draft.native_id)
+
+        # Synchronous way of waiting for CMR to complete the ingest work
+        wait_for_cmr
+
+        raise Array.wrap(ingest_response.body['errors']).join(' /// ') unless ingest_response.success?
+        puts ">>> ingested! ingest_response:\n#{ingest_response.body}"
```
charleshuang80, line 96: remove output

app/assets/javascripts/validation.coffee
```diff
@@ -885,6 +885,22 @@ $(document).ready ->
 
   # Validate the whole page on page load
   if $('.metadata-form, .umm-form').length > 0
+    # if url has a hash, user clicked on a specific field or progress circle
+    # we should visit the field (or nested fields) to show the user any issue
+    # with the field since they have chosen to visit the field directly
+    if (window.location.hash)
```
charleshuang80, line 891: make this only for collection drafts

app/views/collection_drafts/show.html.erb
```diff
@@ -22,15 +22,33 @@
     <section class="action wide-content">
       <div class="wide-content-inside">
         <% if resource_name == 'collection_draft' %>
-          <% if get_resource.new_record? || (@errors && @errors.size > 0) %>
+          <% if get_resource.new_record? || (!@is_revision && @errors && @errors.size > 0) %>
```
cgokey, line 25: Consider rewriting to:
if (get_resource.new_record? || (@errors && @errors.size > 0)
  if (@is_revision)
       do this
  else 
       do that
end

app/assets/javascripts/validation.coffee
```diff
@@ -900,6 +902,25 @@ $(document).ready ->
...
       validateFromFormChange()
 
+  # For collection drafts only (b/c implemented as part of the progressive update feature MMT-2660),
+  # Validate a specific field or fieldset on load, if the user has visited it directly
+  if $('.metadata-form').length > 0
+    # if url has a hash, user clicked on a specific field or progress circle
+    # we should visit the field (or nested fields) to show the user any issue
+    # with the field since they have chosen to visit the field directly,
+    # presumably the user wants to fill in or correct any issue with that field
+    if (window.location.hash)
```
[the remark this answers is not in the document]
charleshuang80: good point. I could explore that option, but in this case I think we are ok with the behavior coming from progress circles as well as any future links, so how it is will work for that. But definitely something to keep in mind if we want to differentiate the behavior in the future.

app/views/collection_drafts/show.html.erb
charleshuang80: yeah that seems cleaner, made the change

cgokey (approved):

ChristianTrummer99 (approved):

PR #776, opened 2021-09-22
MMT-2696:   Used sanitize function to replace html_escape to remove html tags in URS name

jdf-eos:


htranho (approved):

spec/features/manage_cmr/groups/filtering_groups_spec.rb
```diff
@@ -157,7 +157,7 @@
           end
 
           it 'displays the script tag in the user name' do
```
william-valencia, line 159: Looks like we need to change the description on this test since it specifies displaying the script tag in the user name.

ChristianTrummer99 (approved):

william-valencia (approved):

PR #796, opened 2021-11-19
Mmt 2757 Recommended keywords should not be added to collection by default

htranho:


cgokey (approved):

william-valencia:
Looks good but I think Joe wanted to add some text about the suggestions only being showed once.

william-valencia (approved):
Approved, but please check Joe for extra wording on keyword suggestions only being shown once.

PR #798, opened 2021-12-02
MMT-2710: update UMM-C to v1.16.6

charleshuang80:
* copy and merge UMM-C v1.16.6
* add FreeAndOpenData field to UseConstraints fields
* modify UseConstraints type radio button selection and behavior
* modify clear radio button link data attrs and update function
* update validation for UseConstraints fields with oneOf and new field behavior
* update factory
* update tests
* add migration

### DIFF
--- a/app/views/collection_drafts/forms/fields/_geometry_fields.html.erb
+++ b/app/views/collection_drafts/forms/fields/_geometry_fields.html.erb
@@ -32,8 +32,8 @@
       prefix: name_prefix,
       required: true
     ) %>
-    <a href="javascript:void(0)" class="clear-radio-button" data-content="coordinate-system-type">Clear</a>
-    <div class="coordinate-system-type-group">
+    <a href="javascript:void(0)" class="clear-radio-button" data-clear-fields="coordinate-system-type-clear-link-clear-fields" data-hide-fields="coordinate-system-type-clear-link-hide-fields" data-radio-fields="coordinate-system-type-clear-link-radio-fields">Clear</a>
+    <div class="coordinate-system-type-group coordinate-system-type-clear-link-radio-fields">
       <p class="radio-group">
         <%= radio_button_tag(name_to_param("#{name_prefix}|coordinate_system|"), 'CARTESIAN', coordinate_system == 'CARTESIAN', class: 'coordinate-system-select validate', data: { level: remove_pipes(name_prefix) }) %>
         <%= mmt_label(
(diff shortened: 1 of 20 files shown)


app/views/collection_drafts/forms/fields/_geometry_fields.html.erb
```diff
@@ -32,8 +32,8 @@
       prefix: name_prefix,
       required: true
     ) %>
-    <a href="javascript:void(0)" class="clear-radio-button" data-content="coordinate-system-type">Clear</a>
-    <div class="coordinate-system-type-group">
+    <a href="javascript:void(0)" class="clear-radio-button" data-clear-fields="coordinate-system-type-clear-link-clear-fields" data-hide-fields="coordinate-system-type-clear-link-hide-fields" data-radio-fields="coordinate-system-type-clear-link-radio-fields">Clear</a>
```
cgokey, line 35: should this say: resolution-and-coordinate-system-type to be consistent?

app/views/collection_drafts/forms/fields/_geometry_fields.html.erb
```diff
@@ -32,8 +32,8 @@
...
+    <div class="coordinate-system-type-group coordinate-system-type-clear-link-radio-fields">
```
cgokey, line 36: likewise here

spec/factories/collection_factory_data.rb
```diff
@@ -594,6 +594,7 @@ def collection_one
     'Quality'            => 'Metadata quality summary',
     'UseConstraints' => {
       'Description' => 'These are some use constraints',
+      'FreeAndOpenData' => true,
```
cgokey, line 597: Is this correct?   Shouldn't this have a description and/or type?

charleshuang80: The type is not an actual field, the radio buttons are just so the form knows what fields to show or hide.
The description is just above.

app/views/collection_drafts/forms/fields/_geometry_fields.html.erb
charleshuang80: So this is the reason I changed the other radio buttons to `resolution-and-coordinate-system-type`. This is actually correct and for a different field. So the other was kind of a dupe and made for confusion.

cgokey (approved):

htranho (approved):

PR #805, opened 2022-02-16
Mmt 2814 - Fixed build failures from latest changes to CMR.

cgokey:
Adds test user to local URS/EDL before creating subscription as CMR wiill check to ensure user exists in EDL. Added include_granule_counts to CMR collection query, as deleting a collection requires a check to ensure there are no granules and this info was missing.

### DIFF
--- a/spec/features/manage_cmr/subscriptions/index_subscriptions_spec.rb
+++ b/spec/features/manage_cmr/subscriptions/index_subscriptions_spec.rb
@@ -130,10 +130,6 @@ describe 'Viewing a list of subscriptions', reset_provider: true do
 
       clear_cache
 
-      # CMR does a check to ensure the user 'fakeid' exists in EDL.
-      # So add this user to local cmr.
-      cmr_client.add_users_to_local_cmr(['fakeid'], nil)
-
       _ingest_response2, @search_response3, @subscription3 = publish_new_subscription(provider: 'MMT_1', email_address: 'fake@fake.fake', query: 'polygon=10,10,30,10,30,20,10,20,10,10&equator_crossing_longitude=0,10', subscriber_id: 'fakeid', collection_concept_id: @c_ingest_response2['concept-id'])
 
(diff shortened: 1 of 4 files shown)


spec/features/manage_cmr/subscriptions/index_subscriptions_spec.rb
```diff
@@ -130,6 +130,10 @@
 
       clear_cache
 
+      # CMR does a check to ensure the user 'fakeid' exists in EDL.
+      # So add this user to local cmr.
+      cmr_client.add_users_to_local_cmr(['fakeid'], nil)
```
charleshuang80, line 135: I think it makes more sense to do this in the cmr_client method that creates or updates a subscription. See https://github.com/nasa/mmt/blob/76f38989f3790692e85009f6711274d1a4711e7f/lib/cmr/cmr_client.rb#L578, for adding members to groups. Based on the error and the similarity of this issue to groups, I think that a developer working on subscriptions in development will also likely run into this issue and may not understand what to do. The comments in the code linked to might also be improved to that effect.

spec/features/manage_cmr/subscriptions/show_subscription_spec.rb
```diff
@@ -21,6 +21,10 @@
 
   context 'when visiting the show page' do
     before :all do
+      # CMR does a check to ensure the user 'rarxd5taqea' exists in EDL.
+      # So add this user to local cmr.
+      cmr_client.add_users_to_local_cmr(['rarxd5taqea'], nil)
```
charleshuang80, line 26: Same comment as above.

spec/features/manage_cmr/subscriptions/index_subscriptions_spec.rb
cgokey: That seems cleaner, will change, thanks.

william-valencia (approved):

charleshuang80 (approved):

PR #829, opened 2022-03-21
MMT-2782: Updating GKR with keywords that were not originally recommended.

william-valencia:


app/controllers/collection_drafts_controller.rb
```diff
@@ -90,7 +90,8 @@ def update
         gkr_request_uuid = params[:gkr_request_id];
         accepted_keyword_uuids = accepted_recommended_keywords.map { |keyword| params[:keyword_uuid][keyword]};
         rejected_keyword_uuids = params[:keyword_uuid].values - accepted_keyword_uuids;
-        send_feedback(current_user.urs_uid, request.uuid, current_user.provider_id, gkr_request_uuid, accepted_keyword_uuids, rejected_keyword_uuids)
+        new_keywords = get_resource.draft.fetch('ScienceKeywords', []).map { |keyword| construct_keyword_string(keyword, "")} - accepted_recommended_keywords
```
cgokey, line 93: Will this still work if we change MMT to allow produce the GKR screen multiple times?  i.e., right now I can see it working because GKR is only presented once.

cgokey (approved):

PR #845, opened 2022-04-20
MMT-2915 Use env parameters for access app js and css urls

htranho:


rosycordova (approved):

.env
```diff
@@ -0,0 +1,2 @@
+metadata_preview_js_url="https://access.sit.earthdata.nasa.gov/plugin/metadata-preview.0.0.38.js"
```
cgokey, line 1: I'm wondering if we should check this in?   I guess it will be overwritten by bamboo.   If you keep, wonder if we should put a production version in there?

tle1989 (approved):

.env
htranho: Correct, this should not be checked in. Will remove it.

PR #893, opened 2022-10-04
MMT-3047: As a user, I would like to upload an existing .json file stored on my local machine to MMT

dmistry1:


william-valencia (approved):

app/controllers/collection_drafts_controller.rb
```diff
@@ -200,10 +200,15 @@ def upload_json
...
+                  rescue
+                    nil
+                  end
 
     # if the user has uploaded a .json file -> save the file
-    if uploaded_file
-      json_params = JSON.parse(uploaded_file.read)
+    if uploaded_file && json_params
+
       get_resource.draft = json_params
```
cgokey, line 212: I think you can probably put the begin/rescue around the get_resource.draft = JSON.parse(uploaded_file.read)
and the in the rescue flash a message, than having to use a temporary variable

cgokey: and the rescue would:
       flash[:error] = 'There was an error in the uploaded .json file. Please check the file and try again.'
       redirect_to manage_collections_path

tle1989 (approved):

cgokey (approved):

PR #953, opened 2023-02-10
MMT-3130: Enable metadata curators to add EULA identifiers to collections, so consumers can be notified if a collection is bound by an EULA

dmistry1:


lib/assets/schemas/collections/umm-c-json-schema.json
```diff
@@ -300,6 +300,12 @@
       "description": "This sub-element if true, describes to end users and machines that this collection's data is free of charge and open for any use the user sees fit.",
       "type": "boolean"
     },
+    "EULAIdentifierType": {
```
tle1989, line 303: EULAIdentifiersType or EULAIdentifierType?

tle1989 (approved):

lib/assets/schemas/collections/umm-c-json-schema.json
dmistry1: It is EULAIdentifierType and not with an 's'. I got this definition from Unified Metadata Model repo (https://git.earthdata.nasa.gov/projects/EMFD/repos/unified-metadata-model/browse/collection/v1.17.2/umm-c-json-schema.json)

cgokey (approved):

PR #955, opened 2023-02-15
Mmt-3116: Update MMT to use the new "Order Option" API

htranho:


app/controllers/manage_cmr/order_options_controller.rb
```diff
@@ -4,31 +4,111 @@ class OrderOptionsController < ManageCmrController
...
+    metadata_specification['Name'] = Rails.configuration.order_option_label
+    metadata_specification['Version'] = Rails.configuration.order_option_version
+    metadata_specification['URL'] = Rails.configuration.order_option_url
+    order_option['MetadataSpecification'] = metadata_specification
+
+    response = cmr_client.create_update_order_option(order_option: order_option, provider_id: current_user.provider_id, native_id: native_id, token: token)
+
+    order_option_response_concept_id = response.body.fetch('concept-id', '')
+    if response.success?
+      # Wait for CMR before retrieving the new record
```
htranho, line 90: comment: wait for indexing

htranho: try lower number

htranho: loop to try fetch

app/controllers/manage_cmr/order_options_controller.rb
```diff
@@ -54,30 +131,95 @@ def create
     end
   end
 
-  def show
+  def deprecate_order_option
```
jceaser, line 134: I think a comment may be needed something along the lines of "deprecate AKA remove order option" or something to that affect. To the uninitiated it looks like you renamed a function you want to delete latter. Everyone involved probably understands you, but I'm sure someone in the future will be confused.

jceaser (approved):

app/controllers/manage_cmr/order_options_controller.rb
```diff
@@ -54,30 +131,95 @@ def create
...
+      flash[:success] = 'Order Option was successfully deprecated.'
+    else
+      Rails.logger.error("Deprecate Order Option Error: #{soap_xml_response.error_message}")
+      flash[:error] = soap_xml_response.error_message
+    end
+    redirect_to order_options_path
+  end
+
+  def delete_order_option
+    response = cmr_client.remove_order_option(native_id: params[:id], provider_id: current_user.provider_id, token: token)
```
eudoroolivares2016, line 180: I am slightly concerned about us deleting the `order-options`. When we migrated the order-options from legacy services we associated them to collections and services by adding them as part of the data payload, if we delete the order-option from CMR it will still remain on the payload associated between the service and collection. Querying for an order-option with the concept-id in that payload will of course return no results.

eudoroolivares2016: I don't think that is necessarily part of this ticket's scope though

app/controllers/manage_cmr/order_options_controller.rb
htranho: No, this function is to do deprecating an order option (not it self is deprecated)

PR #962, opened 2023-02-27
Mmt-3181: MMT uses concept_id instead of id to fetch particular order option

htranho:


app/controllers/manage_cmr/order_options_controller.rb
```diff
@@ -242,10 +241,10 @@ def legacy_edit_order_option
   end
 
   def update_order_option
-    order_option_id = params[:id]
+    order_option_concept_id = params[:id]
```
eudoroolivares2016, line 244: Dosen't the params here need to be` :concept-id` ?

htranho: this parameter is part of Rails path, defined as 'id' by the framework. The value is what you put in.

spec/vcr/order_options/order_options_update_spec_vcr.yml
```diff
@@ -2,7 +2,7 @@
 http_interactions:
 - request:
     method: get
-    uri: https://cmr.sit.earthdata.nasa.gov/search/order-options.umm_json?id=848DA05B-51A2-1F3D-6783-6C27E5EA74B4&page_size=2000&provider_id=MMT_2
+    uri: https://cmr.sit.earthdata.nasa.gov/search/order-options.umm_json?concept_id=848DA05B-51A2-1F3D-6783-6C27E5EA74B4&page_size=2000&provider_id=MMT_2
```
eudoroolivares2016, line 5:
```suggestion
    uri: https://cmr.sit.earthdata.nasa.gov/search/order-options.umm_json?concept_id=OO1200459850-CMR_ONLY&page_size=2000&provider_id=MMT_2
```
I don't have any issue with it but, maybe we should change the mock to look more like the real thing.

htranho: agree, fixed

eudoroolivares2016 (approved):

PR #989, opened 2023-04-20
Mmt 3155 - Updates to broken tests.

cgokey:


spec/features/collection_drafts/publish_collection_draft_spec.rb
```diff
@@ -218,6 +218,7 @@
       draft = create(:full_collection_draft, user: User.where(urs_uid: 'testuser').first, native_id: 'not & url, encoded / native id', draft_short_name: 'test short name')
       visit collection_draft_path(draft)
       click_on 'Publish'
+      sleep(1)
```
eudoroolivares2016, line 221: Why does this one need a sleep but, the `when publishing a collection draft record` suggest add comment?

app/controllers/manage_cmr/permissions_controller.rb
```diff
@@ -242,7 +246,7 @@ def get_groups
     filters = {
       'provider'  => current_user.provider_id,
       :page_num  => 1,
-      :page_size => 50
+      :page_size => 1000000
```
eudoroolivares2016, line 249: Do we need to consider needing to paginate here? I'm not sure if we can really retrieve that many records using CMR but, there may also be sufficiently less groups than the MAX page size

eudoroolivares2016 (approved):

htranho (approved):

app/controllers/manage_cmr/permissions_controller.rb
```diff
@@ -197,18 +197,22 @@ def set_collection_permission
...
-      if group_permission['group_id'] =~ /(-CMR)$/ && !policy(:system_group).read?
-        # If this user does not have access to view this group mark it for hiding
-        group_permission['is_hidden'] = true
+      group_response = cmr_client.get_edl_group(group_permission['group_id'])
 
-        next
+      if group_response.success?
+        provider_id = group_response.body["tag"]
+        # If this user does not have access to view this group mark it for hiding
+        if provider_id == "CMR" && !policy(:system_group).read?
```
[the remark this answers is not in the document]
cgokey: Yes, the provider_id is considered the tag and this piece of functionality wants to hide the group.   The old way used to look at the concept id, and if it had CMR in it, it would determine that way if it should be hidden.   EDL has that information in the tags, so that is why it is looking in there.

app/controllers/manage_cmr/permissions_controller.rb
cgokey: This page size is a big misleading, as it is mimicking some backgrounds capability functionality that used to page.   The paging is actually now only used to process a chunk of results returned in order to fetch the # of users.   This is just telling to to process the entire results.

app/controllers/manage_cmr/provider_identity_permissions_controller.rb
```diff
@@ -27,7 +27,7 @@ def index
     groups_response = cmr_client.get_edl_groups(filters)
 
     group_list = if groups_response.success?
-                   group_list = groups_response.body.fetch('items', [])
+                   groups_response.body.fetch('items', [])
```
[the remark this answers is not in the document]
cgokey: No that is just rails syntax for what is returned in the body of the if statement.   Funky syntax but the results is assigned to group_list

tle1989 (approved):

dmistry1 (approved):

spec/features/collection_drafts/publish_collection_draft_spec.rb
```diff
@@ -218,13 +218,14 @@
...
       click_on 'Publish'
+      sleep 1
     end
 
     it 'displays a confirmation message' do
       expect(page).to have_content('Collection Draft Published Successfully!')
     end
 
-    it 'displays the published record page' do
+    it 'displays the published record page', skip:true do
```
eudoroolivares2016, line 228: Is this skip failure due to some recent CMR update? i.e. started appearing today?

cgokey: Yes we are seeing some unexpected behavior from the local CMR uber jar when publishing. It says, "This collection is not available.   It is either being published right now, does not exist, or you have insufficient permissions to view the collection".   The strange thing is that this happens only when the CMR uber jar is first started.   If we run the test a subsequent time, it passes and works.   Has something changes where it might take longer than usual for CMR to make the record viewable?

cgokey: On another note, we verified the functionality works in SIT, we are seeing any issues, just with the uber jar.

spec/features/collection_drafts/publish_collection_draft_spec.rb
cgokey: Next round of changes I'm going to remove the sleep, as it didn't make any difference.

PR #999, opened 2023-04-27
Mmt 3190: Update drafts API to prevent clients from CRUDing other provider's draft records

tle1989:


spec/controllers/api/drafts_controller_spec.rb
```diff
@@ -79,6 +79,21 @@
     assert_equal(parsed_body['error'], 'unauthorized')
   end
 
+  it 'can not update if the user does not belong to the provider list.' do
+    # The draft is created by a MMT_2 user
+    # The user updating the document does not have MMT_2 in their provider list, only 'LARC'
+    allow_any_instance_of(Cmr::UrsClient).to receive(:validate_mmt_token).and_return(Faraday::Response.new(status: 200, body: '{"uid":"testuser2"}', response_headers: { 'Content-Type': 'application/json; charset=utf-8' }))
```
cgokey, line 85: I would just make a note in the comment that testuser2's available provider list is just LARC.

spec/controllers/api/drafts_controller_spec.rb
```diff
@@ -79,6 +79,21 @@
     assert_equal(parsed_body['error'], 'unauthorized')
   end
 
+  it 'can not update if the user does not belong to the provider list.' do
```
cgokey, line 82: I see a test for preventing users from seeing other providers drafts for update, delete, and query.   But I think we need tests or create and retrieve as well to verify these work as expected.

tle1989: Fixed

spec/controllers/api/drafts_controller_spec.rb
tle1989: fixed

spec/controllers/api/drafts_controller_spec.rb
```diff
@@ -79,6 +79,60 @@
...
+      parsed_body = JSON.parse(parsed_body)
+    end
+    response_draft = parsed_body['draft']
+    draft_json = @tool_draft.draft
+    assert_equal(response_draft['a new draft'], draft_json['a new draft'])
+  end
+
+  it 'retrieve a draft with error if a user does not belong to a provider list.' do
+    # The draft is created by testuser and Provider is MMT_2
+    # testeruser2 requesting the document does not have MMT_2 in their provider list, only 'LARC'
```
cgokey, line 129: typo, should be testuser2

spec/controllers/api/drafts_controller_spec.rb
```diff
@@ -79,6 +79,60 @@
...
+    parsed_body = JSON.parse(response.body)
+    unless parsed_body.is_a?(Hash)
+      parsed_body = JSON.parse(parsed_body)
+    end
+    response_draft = parsed_body['draft']
+    draft_json = @tool_draft.draft
+    assert_equal(response_draft['a new draft'], draft_json['a new draft'])
+  end
+
+  it 'retrieve a draft with error if a user does not belong to a provider list.' do
```
cgokey, line 127: The test 'it responds unauthorized if the user does not belong to the provider list.' (line 52-59) can be removed as this is the same test, but I like the comments better in this one.

spec/controllers/api/drafts_controller_spec.rb
```diff
@@ -79,6 +79,60 @@
...
+      "LongName": "a long name",
+      "Version": "10.0"
+    }}.to_json
+    put :update, body: jsonContent, params: { id: @tool_draft.id, draft_type: "ToolDraft" }
+    assert_equal(response.status, 401)
+    parsed_body = JSON.parse(response.body)
+    assert_equal(parsed_body['error'], 'unauthorized')
+  end
+
+  it 'create a draft record.' do
```
cgokey, line 97: So you have a test to prevent updating a draft record with a illegal provider, but not create.   You'll need to do something like this test but with a illegal provider instead.

htranho (approved):

PR #1018, opened 2023-06-06
Mmt-3300: Set timeout waiting for CMR startup

htranho:


lib/test_cmr/load_data.rb
```diff
@@ -57,14 +58,19 @@ def load_data
...
+          begin
+            cmr_up = yield
+          rescue Faraday::ConnectionFailed
+            puts 'CMR is still starting, please wait...'
+            sleep 5
+          end
         end
       end
+    rescue Timeout::Error => e
+      puts('Got timeout waiting for CMR, existing task!')
```
cgokey, line 72: should this probably be Logger.error instead of puts

cgokey: Or maybe both.

htranho: Rake tasks don't have logger setup. Existing task use 'puts', which come into Bamboo log file.

tle1989 (approved):

PR #1025, opened 2023-07-10
MMT-3326: As a user, I should be able to see the Variable preview in MMT

dmistry1:


cgokey (changes requested):
I think only change is what we talked about this morning, was to remove the 30px (?) padding that was making it off center.

cgokey (approved):

PR #1047, opened 2023-08-07
MMT-3352: For the Manage Variables->Instance Information form, error messages for some of the required fields do not contain spaces when the fields are not populated

dmistry1:


app/javascript/components/editor/components/widgets/CustomRadioWidget.tsx
```diff
@@ -9,10 +9,15 @@ interface CustomRadioWidgetProps extends WidgetProps {
 class CustomRadioWidget extends React.Component<CustomRadioWidgetProps, never> {
   render() {
     const {
-      required, id, value, onChange, label = '', options = {}
```
cgokey, line 12: we should remove `options` from the Props as well, like you did the other classes.

app/javascript/components/editor/components/widgets/CustomSelectWidget.tsx
```diff
@@ -103,21 +103,26 @@ class CustomSelectWidget extends React.Component<CustomSelectWidgetProps, Custom
     const selectOptions: SelectOptions[] = []
     const { setFocus, loading } = this.state
     const {
-      required, label, schema, options = { enumOptions: null }, registry, isLoading = loading, disabled
```
cgokey, line 106: Here as well remove options title from the props as well.

app/javascript/components/editor/components/widgets/CustomRadioWidget.tsx
dmistry1: This was removed and replaced with `uiSchema`

mandyparson (approved):

PR #1065, opened 2023-08-23
MMT-3315: As a user I want to see variable instance/zarr store information in the variable tab

dmistry1:


tle1989 (approved):

htranho (approved):

cgokey (approved):

mandyparson (commented):
Looks good just need to change the header to match the others. Should be h1-title with h1-box.

PR #1095, opened 2023-12-22
MMT-3080: Update MMT Status page to only check status of MMT database

mandyparson:
Update the MMT status page to only check the status of MMT database.
https://mmt.earthdata.nasa.gov/status

{"database": true}

### DIFF
--- a/app/middleware/middleware_healthcheck.rb
+++ b/app/middleware/middleware_healthcheck.rb
@@ -18,22 +18,16 @@ class MiddlewareHealthcheck
         # checks the database health when /status?checkDatabase is passed
         if check_database
           response_output = process_database_check
-        else
-          response_output = response_output
         end
 
         # checks the health of launchpad when /status?checkLaunchpad is passed
         if check_launchpad
           response_output = process_launchpad_check
-        else 
-          response_output = response_output
-        end 
+        end
 
         # checks health of both database and launchpad when /status is passed
         if !check_database && !check_launchpad
           response_output = process_status_check
-        else
-          response_output = response_output
         end
 
         response[2] = response_output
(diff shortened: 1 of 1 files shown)


htranho (approved):

app/middleware/middleware_healthcheck.rb
```diff
@@ -5,36 +5,79 @@ def initialize(app)
...
-          Rails.logger.error "Database error: #{e}"
-          db_healthy = false
+        queryParams = env['QUERY_STRING']
+        check_database = (queryParams == 'checkDatabase')
+        check_launchpad = (queryParams == 'checkLaunchpad')
+        responseOutput = ""
+
+        # checks the database health when /status?checkDatabase is passed
+        if check_database
+          begin
```
cgokey, line 18: I think maybe create a couple functions would be helpful to remove redundant code.   maybe have a function called checkDatabase() and checkLaunchpad() and have it return true/false if healthy, than you can use those functions in the different use cases you have.   I think will make the main body the function a bit smaller.

mandyparson: Alright this should be good. Glad you said something because I realized lines 41-45 had been modified as well so I put them back to the way they were.

app/middleware/middleware_healthcheck.rb
```diff
@@ -5,33 +5,38 @@ def initialize(app)
...
+        check_launchpad = (query_params == 'checkLaunchpad')
+        db_healthy = false
+        launchpad_healthy = false
+        response_output = ""
+
+        # checks the database health when /status?checkDatabase is passed
+        if check_database
+          response_output = process_database_check
+        else
+          response_output = response_output
```
cgokey, line 22: you can remove the else in line 22, 29, and 36

cgokey: Other than that, looks good.

cgokey (approved):

tle1989 (approved):

cgokey (approved):

PR #1109, opened 2024-01-25
MMT-3568: Setup a lambda for launchpad login/callbacks

cgokey:
PR focuses on the lambdas only.    The workflow is as follows:

1. Web Client requests to login via https://api.mmt.sit.earthdatacloud.nasa.gov/saml-login (lambda) and includes a `?target` to the route path which they should be returned to.
2. `saml-login` lambda generates a `authorizationUrl` using the `node-saml` package.   The lambda will then redirect to this  launchpad authorization url and will include: the `relayState` (target to return to), callback url (endpoint is another lambda), along with a number of key/values needed by launchpad for authentication (found in `static.config.js`) 
4. `saml-acs` is a lambda that responds to the launchpad callback.  The launchpad token is found in the header.  It needs to parse the body of the response to pull out details such as email, auid, etc.   This is an encoded response, so it uses the `Saml2js` to parse out these details.   It then takes all this information and create a JWT token and will pass (via redirect) this token to the MMT React Application as a `jwt` query parameters in the URL.
5. The MMT React Application will then parse `jwt` query parameter, pull out the launchpad token and use this for CMR graphql queries.  (This implementation will follow in a separate ticket).

Couple things to note:
1. `static.config.js` includes a cert key.   The `node-saml` package requires a value here.   Although I looked through the rails mmt code and it nowhere uses this cert for logging in.   It only uses the cert for validating a launchpad token.   So right now I'm just using `fake_cert` as a value and not seeing any issues testing locally.
2. We will need to put in a request to launchpad to allow requests to come from `api.mmt.[sit|uat].earthdatacloud.nasa.gov` and include respective callback urls for any of the above to work. I tested all this using a local version using `mmt.localtest.earthdata.nasa.gov` to verify the code works and properly authenticates. (these urls are currently registered with launchpad).


package-lock.json
```diff
@@ -19,9 +20,13 @@
         "classnames": "^2.3.2",
         "commafy": "^0.0.6",
         "compact-object-deep": "^1.0.0",
+        "cookie": "^0.6.0",
         "esbuild": "^0.19.5",
+        "fs": "^0.0.1-security",
```
dmistry1, line 25: I don't see this being used anywhere, Is this like a dependent library?

package-lock.json
```diff
@@ -19,9 +20,13 @@
         "classnames": "^2.3.2",
         "commafy": "^0.0.6",
         "compact-object-deep": "^1.0.0",
+        "cookie": "^0.6.0",
```
dmistry1, line 23: This can be removed since it is only using tiny-cookie

cgokey: It is actually the opposite. this is being used in the lambda to pull out the launchpad token.
The tiny cookie isn't being used in this code, but I have a follow on PR that will use the tiny cookie in the browser.

dmistry1 (approved):

htranho (approved):

trevorlang (approved):

PR #1114, opened 2024-01-31
Mmt 3492: Creating a resuable table Component

mandyparson:


static/src/js/components/DraftList/DraftList.jsx
```diff
@@ -55,11 +55,28 @@ const DraftList = ({ draftType }) => {
...
-      conceptId, revisionDate, previewMetadata: {
-        name, longName, shortName, title
-      }
-    } = item
+    let conceptId
+    let revisionDate
+    let title
+    let subTitle
+
+    if (item.ummMetadata) {
```
dmistry1, line 63: I don't think this check is needed. I feel like we can use just the ummMetadata for all cases.

trevorlang (approved):
This looks good, thought I agree with Deep that we could probably simplify things by pulling everything out of the umm metadata, if possible.

static/src/js/components/DraftList/DraftList.jsx
mandyparson: I simplified the logic. Do you think it is better to keep it as is, or change all the graphql queries for get[ConceptType]Drafts?

dmistry1 (approved):

cgokey (approved):

PR #1115, opened 2024-01-31
MMT-3553: As a user, I want to clone a publish record

dmistry1:


static/src/js/components/PublishPreview/PublishPreview.jsx
```diff
@@ -112,6 +113,15 @@ const PublishPreview = () => {
     ingestMutation(derivedConceptType, ummMetadata, nativeId, providerId)
   }
 
+  // Calls ingestDraft mutation with a new nativeId
+  const handleClone = () => {
+    const cloneNativeId = `MMT_${crypto.randomUUID()}`
+    // Removes the value from the metadata that has to be unique
+    removeMetadataKeys(ummMetadata, ['Name', 'LongName', 'ShortName'])
```
cgokey, line 120: Will this work for collection?   Does EntryTitle need to be added to this list?

static/src/js/components/PublishPreview/PublishPreview.jsx
```diff
@@ -187,6 +197,23 @@ const PublishPreview = () => {
             {' '}
             Record
           </Button>
+          {/* Clone Publish record link */}
```
trevorlang, line 200: We can probably remove these comments unless you think theyre really necessary

static/src/js/utils/removeMetadataKeys.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Given a ummMetadata object, removes all the keys that are given in the
+ * keys array.
+ * @param {Object} metadata An object containing the keys and value
+ * @param {Array} keys An array that has all the keys that needs to be removed
+ */
+const removeMetadataKeys = (metadata, keys) => {
+  keys.forEach((key) => {
```
trevorlang, line 8: Rather than modify the existing object, it would probably be safer to return a new object without the keys. Things can get funky when modifying objects that are passed around by reference. This is why it triggers the no-param-reassign warning.

static/src/js/components/PublishPreview/PublishPreview.jsx
dmistry1: Yes this will work for Collections as well. The helper function takes in a list of keys and checks if the key exists. If it does it will remove it. So in this array I am just defining all the keys.

cgokey: I don't see EntryTitle as one of the keys?   Shouldn't this key be included?

dmistry1: In current MMT, ShortName is the only field that is being removed when cloning a draft. I am guessing one ShortName has to be unique for Collections 
image

trevorlang (approved):

PR #1132, opened 2024-02-22
MMT-3588: As a user, I want my launchpad token to stay active and not expire if I'm using MMT

cgokey:
**Supports functionality to refresh a user's launchpad token.**   

During the initial login, the redirect from saml-acs lambda will set a cookie with the user's initial token given to it from launchpad after login.   The `AppContextProvider` will read the cookie and call `setUser` with the token provided.  The `token` object now has 2 values: `tokenValue` and `tokenExp` where `tokenValue` is the launchpad token.   The `tokenExp` is expiration set for 15 minutes (per launchpad folks).   This token expires in 15 minutes, so the `AppContextProvider` creates a timer the fires once per second to check the token to see if the token is about the expire (it will refresh the token 1 minute before expiration).   If it determines the token is about to expired, it will call `saml-refresh-token` which will retrieve a new token and reset the expiration for another 15 minutes.   It will then call `setUser` and set the user object with the new token.   The rest of the code no longer relies on any cookies, only what is in the `user` object.    In order to `App` to have access to the `user` object to retrieve the token, I had to add a HOC called `withProviders` to allow `App` to have access to the `useAppContext` for accessing `user` info.   I'm still messing around with an alternative implementation creating a ApolloProvider but had some issues using it, so I'd like to revisit this next sprint.

**Supports functionality to retrieve a users profile from EDL to display user's name.**

The code will first retrieve a EDL client token, which is necessary in order to do anything with the EDL api.  The client token is used for group managements, as well as an endpoint where you can give it a launchpad token and it will come back with the users profile.   This profile includes the user's name, which we can populate in the header.   In order to get a EDL client token, it needs a EDL_PASSWORD, so the code uses dotenv package.   We need to have bamboo write a dotenv file populating this variable.   Note if the EDL_PASSWORD is not set, it will just return with the auid of the user.


static/src/js/components/MetadataForm/__tests__/MetadataForm.test.js
```diff
@@ -208,7 +230,7 @@ const setup = ({
 describe('MetadataForm', () => {
   describe('when the concept type is a Tool draft', () => {
     describe('when on the related-urls form', () => {
-      test('renders a Form component', async () => {
+      test.only('renders a Form component', async () => {
```
tle1989, line 233: remove only

static/src/js/App.jsx
```diff
@@ -60,34 +60,26 @@ const App = () => {
...
 
   const httpLink = createHttpLink({
     uri: graphQlHost
   })
 
-  // TODO remove eslint-disable after this method expands in MMT-3407
-  // eslint-disable-next-line arrow-body-style
-  const authLink = setContext((_, { headers }) => {
-    // TODO MMT-3407 - get a real token from launchpad to send
+  console.log('using token ', tokenValue)
```
cgokey, line 71: I'd like to keep this log statement and the other log statement showing the refresh token for a short time just to verify things are working as expected in SIT.

serverless/src/samlRefreshToken/handler.js
```diff
@@ -0,0 +1,79 @@
...
+
+  const options = getSamlConfig()
+  const { launchpadRoot, cookieName } = options
+  const path = '/icam/api/pub/sm/v1/keepalive'
+  const fetchUrl = `${launchpadRoot}${path}`
+
+  const response = await fetch(fetchUrl, {
+    method: 'POST',
+    headers: {
+      Origin: 'https://mmt.sit.earthdata.nasa.gov',
```
cgokey, line 47: Pull from config

dmistry1 (approved):

tle1989 (approved):

PR #1140, opened 2024-02-29
Mmt 3588: Refactor how saml-callback sends token

cgokey:
These code changes address a couple issues.

1) The auth code was a bit more complex than it had to be by base64 encoding a cooke with multiple values (token, exp, name, auid, email, etc.) in it.   Not all this info was needed and I believe caused problems with header lengths being too long.
2) The new code addresses the first bullet by only passing the launchpad token in the cookie to the auth callback container, just as a single string, not encoded, so should be smaller.
3) A new lambda was written (edl-profile) that will retrieve this other information (name, email, etc.)
4) The auth context, will now sets the loginInfo cookie with the launchpad token it got from the callback, the name it got from the edl-profile lambda, and it will calculate an expiration.
5) The rest of the code, namely AppContextProvider remains the same, still looking for loginInfo cookie.
6) I've also updated the checkAndRefresh functionality to update the loginInfo cookie with the new token, so if the user refreshes their browser it will have the latest token.   This was overlooked in the original implementation as it only updated the User object, so if they refreshed their browser, it would use the original token retrieved during the initial login.


tle1989 (approved):

static/src/js/components/AuthCallbackContainer/AuthCallbackContainer.jsx
```diff
@@ -9,15 +11,40 @@ import { useSearchParams } from 'react-router-dom'
 
 export const AuthCallbackContainer = () => {
   const [searchParams] = useSearchParams()
+  const { apiHost } = getApplicationConfig()
+
   const path = searchParams.get('target')
+  const auid = searchParams.get('auid')
+
+  const [cookie, setCookie] = useCookies(['launchpadToken', 'loginInfo'])
```
cgokey, line 19: remove loginInfo

serverless/src/edlProfile/handler.js
```diff
@@ -0,0 +1,42 @@
...
+  const getUserName = (profile) => {
+    const firstName = profile.first_name
+    const lastName = profile.last_name
+    const name = firstName == null ? profile.uid : `${firstName} ${lastName}`
+
+    return name
+  }
+
+  let profile
+  if (process.env.EDL_PASSWORD === '') {
```
cgokey, line 21: remove this code and test

static/src/js/components/AuthRequiredContainer/AuthRequiredContainer.jsx
```diff
@@ -22,7 +22,7 @@ export const AuthRequiredContainer = ({
       const nextPath = location.pathname + location.search
       window.location.href = `${apiHost}/saml-login?target=${encodeURIComponent(nextPath)}`
     }
-  }, [])
+  }, [token])
```
cgokey, line 25: Do we need this?

static/src/js/utils/checkAndRefreshToken.js
```diff
@@ -4,12 +4,12 @@ import refreshToken from './refreshToken'
...
  * it will invoke lambda to retrieve a new token and call the provided
- * setUser function to update the token.
+ * setCookie function to update the token.
  * @param {*} token the token currently being  used
  * @param {*} user the user object
- * @param {*} setUser the setUser function to be used to update the token
+ * @param {*} setCookie the setCookie function to be used to update the token
  */
-const checkAndRefreshToken = async (token, user, setUser) => {
+const checkAndRefreshToken = async (token, prevLoginInfo, setCookie) => {
```
cgokey, line 12: callback(newToken)

serverless/src/edlProfile/__tests__/handler.test.js
```diff
@@ -0,0 +1,65 @@
...
+jest.mock('../../../../static/src/js/utils/fetchEdlProfile')
+
+beforeEach(() => {
+  jest.clearAllMocks()
+})
+
+describe('edlProfile when edl password is provided', () => {
+  test('when first, last name is given returns a 200 status code with the full name', async () => {
+    fetchEdlProfile.mockImplementation(() => ({
+      email: 'christopher.d.gokey@nasa.gov',
```
jceaser, line 14: is this something you mean to do?

static/src/js/utils/checkAndRefreshToken.js
```diff
@@ -7,14 +7,14 @@ import refreshToken from './refreshToken'
...
+ * @param {*} callback the callback handler called when it gets a new refresh token.
  */
-const checkAndRefreshToken = async (user, setUser) => {
+const checkAndRefreshToken = async (user, callback) => {
   const { token } = user
   const { tokenValue, tokenExp } = token || {}
 
   if (tokenValue && tokenExp) {
-    // Console.log('time remaining ', new Date(tokenExp), new Date(), ((tokenExp.valueOf() - new Date().valueOf()) / 1000 / 60))
+    // Console.log('time remaining ', new Date(tokenExp), new Date(), ((tokenExp - new Date().valueOf()) / 1000 / 60))
```
jceaser, line 17: is this even needed?

serverless/src/edlProfile/__tests__/handler.test.js
trevorlang: Would be great if we could mock this without the personal info :)

serverless/src/edlProfile/handler.js
```diff
@@ -0,0 +1,34 @@
+import fetchEdlProfile from '../../../static/src/js/utils/fetchEdlProfile'
+
+/**
+ * Handles refreshing user's token
+ * @param {Object} event Details about the HTTP request that it received
+ */
+
+const edlProfile = async (event) => {
```
trevorlang, line 8: This feels like we might want a try catch to handle errors, maybe?

static/src/js/components/AuthCallbackContainer/AuthCallbackContainer.jsx
```diff
@@ -9,17 +11,22 @@ import { useSearchParams } from 'react-router-dom'
...
+
   const path = searchParams.get('target')
+  const auid = searchParams.get('auid')
 
   const navigate = useNavigate()
 
   useEffect(() => {
-    if (path) {
-      navigate(path)
+    updateLoginInfo(auid)
```
trevorlang, line 22: Not a huge fan of the fact we need to pass the auid through to this function. I think the only way around it would be to move the router provider outside of the auth provider, which I am not a huge fan of either...

static/src/js/providers/GraphQLProvider/GraphQLProvider.jsx
```diff
@@ -35,6 +35,10 @@ const GraphQLProvider = ({ children }) => {
   const { token } = user
   const { tokenValue } = token || {}
 
+  if (process.env.NODE_ENV === 'production') {
+    console.log('using ', tokenValue)
```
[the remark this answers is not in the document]
cgokey: I'm just doing it temporarily for 1 run going into SIT, just to make sure the token is being updates properly and there are no issues.   Once that happens I'll remove.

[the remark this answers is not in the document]
trevorlang: Isnt this something that could be confirmed using the network tab somehow? If not, I'm fine leaving it. Seems like something we could verify another way.

static/src/js/providers/AuthContextProvider/AuthContextProvider.jsx
```diff
@@ -54,9 +53,21 @@ const AuthContextProvider = ({ children }) => {
     }
   }, [loginInfo])
 
+  const handleRefreshToken = (refreshToken) => {
+    if (process.env.NODE_ENV === 'production') {
+      const { tokenValue } = refreshToken
+      console.log('refreshing token ', tokenValue)
```
trevorlang: Same comment about possibly testing this in the network tab, instead of having code that needs to be removed later

trevorlang (approved):
Approving with some minor feedback

static/src/js/providers/GraphQLProvider/GraphQLProvider.jsx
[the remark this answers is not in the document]
cgokey: Good point, I'll just remove it.

static/src/js/components/AuthCallbackContainer/AuthCallbackContainer.jsx
cgokey: Yeah agree, I even tried adding ;auid=${auid} to the cookie, so we wouldn't have to pass it via the header but there is something going on in AWS from sending multiple cookies to the browser.

PR #1144, opened 2024-03-05
MMT-3559: As a user, I can see how many tags are associated to published collection

dmistry1:
Added support for Tag, if a collection record has a tag it will show the tag information in the collection search tag and for a individual collection page.

image
image
image

### DIFF
--- a/static/src/js/components/PublishPreview/PublishPreview.jsx
+++ b/static/src/js/components/PublishPreview/PublishPreview.jsx
@@ -62,7 +62,7 @@ const PublishPreview = () => {
   }
 
   const toggleTagModal = (nextState) => {
-    setShowDeleteModal(nextState)
+    setShowTagModal(nextState)
   }
 
   const { addNotification } = useNotificationsContext()
@@ -86,7 +86,6 @@ const PublishPreview = () => {
       }
     },
     onCompleted: (getData) => {
-      console.log('🚀 ~ PublishPreview ~ getData:', getData)
       const fetchedPreviewMetadata = getData[toLowerKebabCase(derivedConceptType)]
       const {
         revisionId: fetchedRevisionId,
@@ -205,7 +204,7 @@ const PublishPreview = () => {
         })
 
         // Hide the modal
-        setShowDeleteModal(false)
+        toggleShowDeleteModal(false)
 
         // Navigate to the manage page
         navigate(`/manage/${pluralize(derivedConceptType).toLowerCase()}`)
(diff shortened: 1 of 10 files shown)


static/src/js/pages/SearchPage/SearchPage.jsx
```diff
@@ -406,13 +407,15 @@ const SearchPage = ({ limit }) => {
                 <ListGroup>
                   <For each={Object.keys(activeTagModalCollection.tags)}>
                     {
-                      (tagKey) => (
+                      (tagKey, index) => (
```
dmistry1, line 410: pull out description up here

static/src/js/components/PublishPreview/PublishPreview.jsx
```diff
@@ -211,11 +228,21 @@ const PublishPreview = () => {
         errorLogger(deleteError, 'PublishPreview: deleteMutation')
 
         // Hide the modal
-        setShowDeleteModal(false)
+        toggleShowDeleteModal(false)
       }
     })
   }
 
+  const getTagCount = () => {
```
dmistry1, line 236: use useCallBack here

static/src/js/components/PublishPreview/PublishPreview.jsx
```diff
@@ -339,6 +383,47 @@ const PublishPreview = () => {
...
+            message={
+              previewMetadata.tagDefinitions
+                ? (
+                  <>
+                    <h3 className="fw-bolder h5">{}</h3>
+                    <ListGroup>
+                      <For each={previewMetadata.tagDefinitions.items}>
+                        {
+                          (tagItems) => (
+                            <ListGroupItem key={tagItems.tagKey}>
```
mandyparson, line 406: Double check this. I'm getting duplicate keys from this one and it looks like it's using a token?
Screenshot 2024-03-06 at 1 48 02 PM

mandyparson: Interesting too since I only have one tag...

cgokey (approved):

static/src/js/components/PublishPreview/PublishPreview.jsx
dmistry1: I looked into it and found this is coming from edsc-metadata-preview. I checked in MMT and we are not using the token as a key

mandyparson (approved):

PR #1175, opened 2024-04-01
MMT-3676: Updates tests to use Vitest

macrouch:
Removes Jest and replaces it with Vitest. All the syntax is the same as Jest, i.e. `jest.mock` just becomes `vi.mock`


.eslintignore
```diff
@@ -1,2 +1,2 @@
-jest.setup.js
+vi.setup.js
```
macrouch, line 1: this was a global replace that didn't need to happen

cgokey (approved):

tle1989 (approved):

PR #1177, opened 2024-04-02
MMT-3602: Search text under "Create New Record" on the MMT homepage does not change regardless of the record type

dmistry1:
Fix the text under Create New Record button to reflect that concept type
image


static/src/js/pages/ManagePage/ManagePage.jsx
```diff
@@ -64,7 +64,11 @@ const ManagePage = () => {
                         {/* TODO need to add the tooltip highlight */}
                         <Button className="p-0 align-baseline" variant="link">search</Button>
                         {' '}
-                        in the top right corner to find published tools to clone or edit.
+                        in the top right corner to find published
```
tle1989, line 67: Add test

tle1989 (approved):

mandyparson (approved):

PR #1179, opened 2024-04-04
MMT-3655: When I get logged out, I occasionally see Error retrieving providers. It shouldn’t be fetching providers on the home page.

tle1989:


abbottry:
All commits _and_ Pull Requests should follow the same format `MMT-3655: Title`

tle1989:
this ticket will be worked differently, so I would just close the PR and let them fix it.   I think there will be a little bit of overhaul of the login functinality

PR #1182, opened 2024-04-05
MMT-3545: As a user, I can mange templates

dmistry1:
# Overview
Adding support for Collection Template on Mange Collections Page. Now there is a`Manage Templates` link and when clicking on it will navigate to Template list page. 

## Screenshots
image
image


static/src/js/App.jsx
```diff
@@ -25,6 +25,7 @@ import { GET_ACLS } from './operations/queries/getAcls'
 import useAppContext from './hooks/useAppContext'
 import withProviders from './providers/withProviders/withProviders'
 import getPermittedUser from './utils/getPermittedUser'
+import TemplatesPage from './pages/TemplatesPage/TemplatesPage'
```
abbottry, line 28: Going forward let's do a better job about grouping our imports.

static/src/js/App.jsx
```diff
@@ -151,6 +152,16 @@ export const App = () => {
                 )
               }
             />
+            <Route
+              path="templates/:templateType/*"
```
abbottry, line 156: What is the purpose of the `*` here?

static/src/js/components/TemplateList/TemplateList.jsx
```diff
@@ -0,0 +1,145 @@
...
+
+      setErrors(error)
+      setTemplateList(response)
+      setLoading(false)
+    }
+
+    fetchTemplates()
+  }, [])
+
+  const buildEllipsisLinkCell = useCallback((cellData, rowData) => {
```
abbottry, line 39: What is the purpose of the `useCallback` with an empty dependency array? My understanding is that this is the equivalent of `useMemo` because without a dependency array the same 'version' of the method is used every single time.

static/src/js/components/TemplateList/TemplateList.jsx
```diff
@@ -0,0 +1,145 @@
...
+  const buildActionsCell = useCallback((cellData, rowData) => {
+    const { id } = rowData
+
+    return (
+      <Row>
+        <Col className="col-auto">
+          <Link to={`/templates/${pluralize(templateType).toLowerCase()}/${id}/collection-information`}>Edit</Link>
+        </Col>
+        <Col className="col-auto">
+          {/* TODO: MMT-3548: As a user, I can delete a collection template */}
```
abbottry, line 58: I would exclude non-functioning UI elements in the future, just add it with the ticket mentioned.

static/src/js/components/TemplateList/TemplateList.jsx
```diff
@@ -0,0 +1,145 @@
...
+    const { id } = rowData
+
+    return (
+      <EllipsisLink to={`/templates/${pluralize(templateType).toLowerCase()}/${id}`}>
+        {cellData}
+      </EllipsisLink>
+    )
+  }, [])
+
+  const buildActionsCell = useCallback((cellData, rowData) => {
```
abbottry, line 49: Same comment regarding the empty dependency array. Just be sure this is doing what youre expecting, we don't want methods being memoized in ways we don't expect.

static/src/js/pages/TemplatesPage/TemplatesPage.jsx
```diff
@@ -0,0 +1,50 @@
...
+import MetadataForm from '../../components/MetadataForm/MetadataForm'
+import TemplateList from '../../components/TemplateList/TemplateList'
+import DraftPreview from '../../components/DraftPreview/DraftPreview'
+
+const TemplatesPage = () => {
+  const { templateType } = useParams()
+
+  const currentTemplateType = urlValueTypeToConceptTypeMap[templateType]
+
+  if (templateType && !currentTemplateType) {
```
abbottry, line 18: If `templateType` isn't provided, wouldn't react-router render a completely different component rather than throw a 404?

static/src/js/utils/getTemplates.js
```diff
@@ -0,0 +1,24 @@
+import { getApplicationConfig } from '../../../../sharedUtils/getConfig'
+
+const getTemplates = async (providerId, token) => {
```
abbottry, line 3: Add a doc block for the method.

abbottry (changes requested):

static/src/js/components/TemplateList/TemplateList.jsx
dmistry1: From what I understand, when I provide an empty dependency it means that the callback function should never change as long as the component is mounted. Which is what I want, I want the function to remain constant through out the component lifecycle.

static/src/js/utils/getTemplates.js
dmistry1: Added js doc

abbottry (approved):

trevorlang (approved):

PR #1183, opened 2024-04-05
Mmt 3495: Creating revisions table

mandyparson:
You will need to create a published record in Collections and Variables || Tools || Services with revisions to test this ticket fully. 

Once you have published records with revisions, search for your record and pull up it's published preview

There will be a link for 'Revisions' with a badge of how many revisions you have in the navigation bar. Click on it. 
Screenshot 2024-04-05 at 11 08 09 AM

You should see a loading screen
Screenshot 2024-04-05 at 11 12 22 AM

Once loaded, you'll see a table with current mmt.sit parity. Everything should work as expected except for the 'Actions' column which will be taken care of in another ticket. 
Screenshot 2024-04-05 at 11 12 40 AM


abbottry (changes requested):

static/src/js/components/RevisionList/RevisionList.jsx
```diff
@@ -0,0 +1,251 @@
+import React, {
```
tle1989, line 1: There is a difference in feature compared to SIT
Screenshot 2024-04-08 at 8 55 28 AM

static/src/js/components/RevisionList/RevisionList.jsx
```diff
@@ -0,0 +1,251 @@
...
+    const isPublishedVersion = (cellData === versions)
+
+    return (
+      <EllipsisLink to={`/${type}/${conceptId}/${cellData}`} inline>
+        {[cellData, ' - ', (isPublishedVersion ? 'Published' : 'Revision')].join('')}
+      </EllipsisLink>
+    )
+  }, [])
+
+  // To be done after GQL-32
```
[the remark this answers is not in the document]
mandyparson: It has? It's referring to the function below. It is currently not functional and awaiting GQL-32 to be completed

static/src/js/hooks/__tests__/useRevisionQuery.test.jsx
```diff
@@ -0,0 +1,262 @@
...
+import PropTypes from 'prop-types'
+import { render, screen } from '@testing-library/react'
+import { MockedProvider } from '@apollo/client/testing'
+
+import { GET_COLLECTION_REVISIONS } from '../../operations/queries/getCollectionRevisions'
+import { GET_VARIABLES } from '../../operations/queries/getVariables'
+import useRevisionsQuery from '../useRevisionsQuery'
+import AppContext from '../../context/AppContext'
+
+const TestComponent = ({ type }) => {
```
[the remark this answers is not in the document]
abbottry: Ultimately though, this should be using code that we're serving to users, not a test component, I think.

static/src/js/components/RevisionList/RevisionList.jsx
abbottry: What is the difference that you're referring to? There are many differences but many are known and expected.

static/src/js/components/RevisionList/RevisionList.jsx
```diff
@@ -0,0 +1,251 @@
...
+                              )
+                            }
+                            {
+                              (count || (!loading && !count)) && (
+                                <span className="text-secondary fw-bolder">{paginationMessage}</span>
+                              )
+                            }
+                          </Col>
+                          {
+                            totalPages > 1 && (
```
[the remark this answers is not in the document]
mandyparson: While the pagination component could include this logic, I'm not sure if it makes sense to keep it out of the parent component. If we took out the logic and just did {pagination}, then in the event of totalPages !> 1 it would render an empty <Col> component, correct?

static/src/js/components/RevisionList/RevisionList.jsx
```diff
@@ -0,0 +1,251 @@
...
+      className: 'col-auto',
+      dataAccessorFn: (cellData) => cellData.split('T')[0]
+    },
+    {
+      dataKey: 'userId',
+      title: 'Action by',
+      className: 'col-auto'
+    },
+    {
+      // Change dataKey to 'revisionId' then fix unique key issue (see line 116)
```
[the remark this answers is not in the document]
mandyparson: I'm not sure if I'll be working on this file when we come back to it in sprint two so it's a note for the next dev. Line 98 will need to take in 'revisionId' instead of 'actions' and this causes a unique key I'd need to dig into to figure out, personally. The keys are being set on line 116.

static/src/js/components/RevisionList/RevisionList.jsx
```diff
@@ -0,0 +1,251 @@
...
+
+    return (
+      <EllipsisLink to={`/${type}/${conceptId}/${cellData}`} inline>
+        {[cellData, ' - ', (isPublishedVersion ? 'Published' : 'Revision')].join('')}
+      </EllipsisLink>
+    )
+  }, [])
+
+  // To be done after GQL-32
+  const buildActionCell = useCallback((cellData) => {
```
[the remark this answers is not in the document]
mandyparson: @trevorlang I had copied this structure from SearchPage.jsx. Do you have any insight?

static/src/js/App.jsx
```diff
@@ -170,6 +171,8 @@ export const App = () => {
             <Route path="*" element={<Navigate to="/404" replace />} />
             <Route path="/:type/:conceptId/" element={<PublishPreview />} />
             <Route path="/:type/:conceptId/:revisionId" element={<PublishPreview />} />
+            <Route path="/:type/:conceptId/revisions/:revisionId" element={<PublishPreview isRevision />} />
+            <Route path="/:type/:conceptId/revisions" element={<RevisionList />} />
```
dmistry1, line 175: Wrap this about AuthRequiredContainer
```
<AuthRequiredContainer>
    <RevisionList />
 </AuthRequiredContainer>
```

static/src/js/components/PublishPreview/PublishPreview.jsx
```diff
@@ -393,6 +398,37 @@ const PublishPreview = () => {
...
+          {
+            isRevision
+              ? (
+                <Button
+                  className="btn btn-link"
+                  type="button"
+                  variant="link"
+                  onClick={handleRevisions}
+                >
+                  Back To Revisions
```
dmistry1, line 411: I do not see this Back to Revisions button in current MMT. I see that it is just routing back to revision table view. Was there a discussion about this button?

static/src/js/components/RevisionList/RevisionList.jsx
tle1989: SIT MMT has  breadcrumbs, Revision Date's format is different

static/src/js/components/PublishPreview/PublishPreview.jsx
mandyparson: There wasn't but I believe changing it to this solves two problems at once. Could I get this approved through Trevor?

dmistry1 (approved):

static/src/js/utils/getRevisionCount.js
```diff
@@ -0,0 +1,20 @@
+import { capitalize } from 'lodash-es'
+import useRevisionsQuery from '../hooks/useRevisionsQuery'
+
+/**
+ * Given a formData with a an array, getRevisionCount will return the length
+ * of of the items
+ * @param {String} conceptId The conceptId we use to query revision Count
+ */
+const getRevisionCount = (conceptId, type) => {
```
abbottry, line 9: This should not be a utility function, just use the hook we created in the one place this utility is used.

abbottry (changes requested):

static/src/js/components/PublishPreview/PublishPreview.jsx
```diff
@@ -471,6 +503,37 @@ const PublishPreview = () => {
...
+          <Row>
+            <Col>
+              <Alert className="fst-italic fs-6" variant="warning">
+                <i className="eui-icon eui-fa-info-circle" />
+                {' '}
+                You are viewing an older revision of this
+                {' '}
+                {`${derivedConceptType}.`}
+                {' '}
+                <span
```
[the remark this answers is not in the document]
mandyparson: A button will require a bit of css finagling as it needs to blend in with the text of the banner. Should I create a whole new PublishPreview.scss for this or do inline styling?

static/src/js/hooks/useRevisionsQuery.js
```diff
@@ -0,0 +1,133 @@
...
+}) => {
+  const [revisions, setRevisions] = useState({})
+  const [error, setError] = useState()
+  const [loading, setLoading] = useState(false)
+
+  // Sets a 'published' flag for records that are published
+  const formatRevisionResults = (results) => {
+    const { items, count, __typename } = results
+
+    const flaggedItems = items.map((item) => {
```
trevorlang, line 24: I think this might break if the sortKey was ever changed. I don't think we can guarantee that the last item in the list is the most recent revision in that case. It might need to be derived by looking at the published date.

One way to do this would be to first run a `reduce` on the items where you can compare the dates and find the most recent one, like in this [example](https://stackoverflow.com/a/57731226). Then use that value to designate that item in the list as published.

trevorlang (changes requested):

abbottry:
@mandyparson Is there a route that we have defined for the /collections/:id/:revisionId that needs to be removed?

abbottry (approved):
Approving, just want to note that if there is a route for collections/:id/:revisionId we should remove it as well.

tle1989 (approved):

PR #1185, opened 2024-04-08
MMT-3632: MMT Vulnerabilities reported March 2024 (B)

tle1989:
To Test local rails mmt: yarn audit
Result: 0 vulnerabilities found

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,8 @@
     "word-wrap": "^1.2.4",
     "@adobe/css-tools": "^4.3.2",
     "postcss": "^8.4.32",
-    "traverse": "^7.23.6"
+    "traverse": "^7.23.6",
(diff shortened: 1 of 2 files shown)


abbottry (changes requested):
Why are we adding express to our package?

tle1989:
> Why are we adding express to our package?

Because we have Vulnerabilities checked on March 27, 2024
2 Moderate
https://github.com/nasa/mmt/security/dependabot/69
Screenshot 2024-04-09 at 10 50 11 AM

tle1989:
Don't have a fix yet, express is a dependency for web pack-dev-server.  When webpack-dev-server updates their dependency they will publish a new version, then we can update webpack-dev-server in our package.json file.

PR #1193, opened 2024-04-18
MMT-3432: As a user, I can view Order Options

dmistry1:
# Overview 

Adding support for viewing Order Options. Now a user can view a list of order options they have for their current provider and by clicking on a single order option, they can view that order option. 

# Screenshots
image
image


static/src/js/components/OrderOptionList/OrderOptionList.jsx
```diff
@@ -0,0 +1,220 @@
...
+import { useLazyQuery } from '@apollo/client'
+import {
+  Col,
+  Placeholder,
+  Row
+} from 'react-bootstrap'
+import { useSearchParams } from 'react-router-dom'
+import moment from 'moment'
+
+import ControlledPaginatedContent from '../ControlledPaginatedContent/ControlledPaginatedContent'
```
abbottry, line 15: 👌

static/src/js/components/OrderOptionList/OrderOptionList.jsx
```diff
@@ -0,0 +1,220 @@
...
+
+  const limit = 20
+  const offset = (activePage - 1) * limit
+
+  const setPage = (nextPage) => {
+    setSearchParams((currentParams) => {
+      currentParams.set('page', nextPage)
+
+      return {
+        ...Object.fromEntries(currentParams)
```
abbottry, line 47: This doesn't need to be an object, you can just return `Object.fromEntries(currentParams)`

static/src/js/components/TemplateList/TemplateList.jsx
```diff
@@ -124,7 +124,7 @@ const TemplateList = () => {
       dataKey: 'lastModified',
       title: 'Last Modified',
       className: 'col-auto',
-      dataAccessorFn: () => moment().format('MM-DD-YYYY')
+      dataAccessorFn: (cellData) => moment(cellData).format('LLLL')
```
abbottry, line 127: Can we define this format as a constant please, that way we can reference it from all of our tables and if we decide to change it, we can change it in one spot.

abbottry (changes requested):
Just a few minor comments I'd like to see addressed.

macrouch (approved):

PR #1204, opened 2024-05-13
MMT-3538: Adds search form to Groups list

macrouch:
# Overview

### What is the feature?

Adds a search form to the Groups list page

Searching for Groups

# Testing

Visit the groups page, search for groups by name/provider/members

# Checklist

### DIFF
--- a/static/src/js/components/GroupList/GroupList.jsx
+++ b/static/src/js/components/GroupList/GroupList.jsx
@@ -116,7 +116,7 @@ const GroupList = () => {
     )
   }, [])
 
-  const buildEllipsisTextCell = useCallback((cellData) => cellData && (
+  const buildEllipsisTextCell = useCallback((cellData) => (
     <EllipsisText>
       {cellData}
     </EllipsisText>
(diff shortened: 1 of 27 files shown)


static/src/js/components/GroupList/GroupList.jsx
```diff
@@ -114,6 +116,12 @@ const GroupList = () => {
     )
   }, [])
 
+  const buildEllipsisTextCell = useCallback((cellData) => cellData && (
+    <EllipsisText>
+      {cellData}
```
trevorlang, line 121: Thought about this a little more. I wonder if we should build this into EllipsisText component so we don't have to handle it at this level in any component that uses it.

static/src/js/components/Header/Header.jsx
```diff
@@ -157,6 +158,7 @@ const Header = () => {
                 >
                   User Guide
                 </Button>
+
```
trevorlang, line 161: Not a huge fan of these empty spaces between the components. I think we are a little inconsistent with it but my vote would be no extra spaces 😅 I think thats what we do in most places.

static/src/js/pages/GroupListPage/GroupListPage.jsx
```diff
@@ -62,6 +56,8 @@ const GroupListPage = () => (
     pageType="secondary"
     header={<GroupListPageHeader />}
   >
+    <GroupSearchForm />
+
```
trevorlang, line 60: See extra space rant above ⬆️ 🙃

trevorlang (approved):
I think it's good to go but have some minor feedback you can take or leave.

abbottry (approved):

PR #1258, opened 2024-07-17
MMT-3784: Added support to disable publish button if any errors are found.

cgokey:
# Overview

### What is the feature?

As an MMT operator, don't allow me to attempt to publish a draft until the metadata will pass ingest without errors

### What is the Solution?

Disables the "Publish" button on the preview page when the metadata record has errors.
Disables the "Save and Publish" button in the navigation menu when the metadata record has errors.

This also has another use case when the form shows no errors, yet since the user hasn't saved the record yet, the the validator is still giving back errors because the metadata specification hasn't been populated yet (this happens on the first save).   So if this is the case, I'm providing a tooltip over the disabled button stating the user must save the record in order to publish.

Navigation menu
Preview Page

# Testing

Try creating a metadata record that has errors.   You should not be able to publish from the preview page or from the form navigation menus (Save and Publish option) if the metadata record has any errors.

# Checklist

### DIFF
--- a/static/src/js/components/FormNavigation/FormNavigation.jsx
+++ b/static/src/js/components/FormNavigation/FormNavigation.jsx
@@ -142,6 +142,7 @@ const FormNavigation = ({
             {
               !isTemplate && (
                 <Dropdown.Item
+                  disabled={errors.length > 0}
                   onClick={() => onSaveClick(saveTypes.saveAndPublish)}
                 >
                   {saveTypesToHumanizedStringMap[saveTypes.saveAndPublish]}
--- a/static/src/js/components/FormNavigation/FormNavigation.jsx
+++ b/static/src/js/components/FormNavigation/FormNavigation.jsx
@@ -75,6 +75,18 @@ const FormNavigation = ({
     onSave(type)
   }
 
+  const publishButtonTooltip = () => {
+    if (errors.length === 0) {
+      return null
+    }
+
+    if (conceptId === 'new') {
+      return 'Publishing disabled until you save the record.'
+    }
+
+    return 'Publishing disabled due to errors in metadata record.'
+  }
+
   return (
     <>
       <div className="mb-4">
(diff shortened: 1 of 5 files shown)


static/src/js/components/FormNavigation/__tests__/FormNavigation.test.jsx
```diff
@@ -163,6 +163,57 @@ describe('FormNavigation', () => {
   })
 
   describe('when clicking the Save & Publish dropdown item', () => {
+    let schema
+
+    beforeEach(async () => {
```
cgokey, line 168: remove async

static/src/js/pages/DraftPage/DraftPage.jsx
```diff
@@ -184,6 +192,7 @@ const DraftPageHeader = () => {
         primaryActions={
           [
             {
+              disabled: validationErrors.length > 0,
```
dmistry1, line 195: We should also add a Tooltip describing why this button is disabled

static/src/js/pages/DraftPage/DraftPage.jsx
```diff
@@ -32,6 +33,7 @@ import ErrorBoundary from '@/js/components/ErrorBoundary/ErrorBoundary'
 import MetadataPreviewPlaceholder from '@/js/components/MetadataPreviewPlaceholder/MetadataPreviewPlaceholder'
 import Page from '@/js/components/Page/Page'
 import PageHeader from '@/js/components/PageHeader/PageHeader'
+import getUmmSchema from '@/js/utils/getUmmSchema'
```
dmistry1, line 36: Move this import to the utils imports, line 16-18

static/src/js/components/FormNavigation/FormNavigation.jsx
```diff
@@ -142,6 +142,7 @@ const FormNavigation = ({
             {
               !isTemplate && (
                 <Dropdown.Item
+                  disabled={errors.length > 0}
```
dmistry1, line 145: Here we should also add a tooltip describing why the button is disabled

mandyparson: Agreed. Because I'm encountering this issue and I'm not sure why: 
Screenshot 2024-07-17 at 3 13 34 PM

mandyparson: ^^This holds true both before and after I click 'save'

static/src/js/components/FormNavigation/__tests__/FormNavigation.test.jsx
```diff
@@ -10,6 +10,7 @@ import {
   Route,
   Routes
 } from 'react-router'
+import { beforeEach } from 'vitest'
```
macrouch, line 13: This import shouldn't be needed. The `globals: true` config in `vite.config.js` (in the test object) means we don't have to import from `vitest` in every test

static/src/js/components/FormNavigation/FormNavigation.jsx
cgokey: Yeah there is a reason for that, it is due to the schemas not being in sync.   That will be fixed with deep's ticket.  If you try creating a umm-var instead of a collection where the schemas are the same the button enables properly.

dmistry1 (approved):

htranho (approved):

PR #1261, opened 2024-07-23
MMT-3794: Fix number field issue for Firefox and Safari

htranho:
# Overview

### What is the feature?

Fixed issue: Firefox and Safari do not prevent input of non numeric characters in number input fields.

### What is the Solution?

Filter out non numeric characters in onChange method

Number input fields, for example Size in RelatedURL

# Testing

### Reproduction steps
Using Firefox (Safari)
- Go to https://mmt.sit.earthdatacloud.nasa.gov/
- Edit a Draft
- In RelatedURL, choose 'DistributionURL' from URL Content Type and choose 'GET DATA' from Type
- Section 'Get Data' will show up with number input field Size
- Any non numeric character can be entered in Size, for example '30MB'. Since that is invalid number, field Size and value don't show up in the json display.
- In the fixed version, only '30' can be entered in the Size field and so the entry is valid and will show up in the json display.

Number fields in this fixed version will accept number, '+', '-', 'E', 'e' and '.'

### Attachments

N/A

# Checklist

### DIFF
--- a/static/src/js/components/CustomTextWidget/CustomTextWidget.jsx
+++ b/static/src/js/components/CustomTextWidget/CustomTextWidget.jsx
@@ -83,7 +83,7 @@ const CustomTextWidget = ({
   const handleBlur = () => {
     setFocusField(null)
 
-    if (onBlur) onBlur(id)
+    onBlur(id)
   }
 
   return (
--- a/static/src/js/components/CustomTextWidget/CustomTextWidget.jsx
+++ b/static/src/js/components/CustomTextWidget/CustomTextWidget.jsx
@@ -70,8 +70,13 @@ const CustomTextWidget = ({
   // Handle the value changing in the field
   const handleChange = (event) => {
     const { value: newValue } = event.target
+    let valueToSet = newValue
+    // If type is number, remove everything except digits, '.', 'e', 'E', '+' and '-'
+    if (type === 'number') {
+      valueToSet = newValue.replace(/[^\d.eE+-]/g, '')
+    }
 
-    onChange(newValue)
+    onChange(valueToSet)
   }
 
   // Handle the field losing focus
@@ -103,7 +108,7 @@ const CustomTextWidget = ({
         placeholder={placeholder}
         ref={focusRef}
         tabIndex={0}
-        type={type && type === 'number' ? 'number' : 'text'}
+        type="text"
         value={value}
       />
     </CustomWidgetWrapper>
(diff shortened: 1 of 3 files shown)


static/src/js/components/CustomTextWidget/CustomTextWidget.jsx
```diff
@@ -70,8 +70,13 @@ const CustomTextWidget = ({
...
+    // If type is number, remove everything except digits, '.', 'e', 'E', '+' and '-'
+    if (type === 'number') {
+      valueToSet = newValue.replace(/[^\d.eE+-]/g, '')
+    }
 
-    onChange(newValue)
+    onChange(valueToSet)
   }
 
   // Handle the field losing focus
```
mandyparson, line 82: I know you inherited this, but as we revisit these test files it would probably be a good idea to zip up any remaining tests we still need. For example, line 86 still needs to be fully tested.

static/src/js/components/PermissionForm/__tests__/PermissionForm.test.jsx
```diff
@@ -278,8 +278,8 @@ describe('PermissionForm', () => {
         await user.type(nameField, 'Test Name')
```
mandyparson, line 278: Likewise here: lines 69-82 and 342

static/src/js/components/CustomTextWidget/CustomTextWidget.jsx
```diff
@@ -103,7 +108,7 @@ const CustomTextWidget = ({
         placeholder={placeholder}
         ref={focusRef}
         tabIndex={0}
-        type={type && type === 'number' ? 'number' : 'text'}
+        type="text"
```
mandyparson, line 111: So I'm guessing there were no use cases that you found in which we used CustomTextWidget as a number?

cgokey (approved):

static/src/js/components/CustomTextWidget/CustomTextWidget.jsx
htranho: Fixed the method handleBlur itself, now line 86 is tested.

static/src/js/components/PermissionForm/__tests__/PermissionForm.test.jsx
htranho: I did not touch the component PermissionForm, just fix the text field component it uses. Not enough knowledge of PermissionForm to add tests there.

static/src/js/components/CustomTextWidget/CustomTextWidget.jsx
htranho: CustomTextWidget is still being used for number field, but not as html element input type="number" because Firefox and Safari have issues with it. See method handleChange:
if (type === 'number') {
      valueToSet = newValue.replace(/[^\d.eE+-]/g, '')
    }

mandyparson (approved):

PR #1267, opened 2024-07-26
MMT-3823: As a user, I want to filter my search results after choosing a provider without loosing my original search results

mandyparson:
# Overview

### What is the feature?

In any search list element (Collection, Variable, Service, or Tool), there is an issue where a user cannot sort their results while a provider has been selected. This ticket fixes that so that after a user selects a provider, they can then use the descending and ascending buttons without losing that provider stipulation. 

### What is the Solution?

Capture the url params before they are deleted and send provider down to function. 

SearchList and tests. Fixed a few other minor issues along the way. 

# Testing

### Reproduction steps

- **Environment for testing: local or sit, just need published records for all concepts
- **Collection to test with: N/A

1. Under Collections select 'All Collections'
2. select a Provider
3. click the up and down arrows for each column header and notice that it only sorts those with the selected provider

### Attachments

Screenshots wouldn't be very helpful here. 

# Checklist

### DIFF
--- a/static/src/js/pages/SearchList/SearchList.jsx
+++ b/static/src/js/pages/SearchList/SearchList.jsx
@@ -19,6 +19,8 @@ import {
 
 import { useSuspenseQuery } from '@apollo/client'
 
+import handleSort from '@/js/utils/handleSort'
+
 import { DATE_FORMAT } from '../../constants/dateFormat'
 import conceptTypeQueries from '../../constants/conceptTypeQueries'
 import conceptTypes from '../../constants/conceptTypes'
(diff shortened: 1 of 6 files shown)


static/src/js/pages/SearchList/SearchList.jsx
```diff
@@ -174,86 +182,86 @@ const SearchList = ({ limit }) => {
...
+        title: 'Name'
       },
       {
-        dataKey: 'longName',
-        title: 'Long Name',
         className: 'col-auto',
         dataAccessorFn: buildEllipsisTextCell,
-        sortFn: (_, order) => sortFn('longName', order)
+        dataKey: 'longName',
+        sortFn,
```
mandyparson, line 248: I did a bunch of sorting but didn't want to let this get lost. Can anyone confirm if this line is needed or not? The test file wanted me to test this line but I couldn't find a use case for it and deleting it did not result in any changes.

static/src/js/utils/handleSort.js
```diff
@@ -0,0 +1,28 @@
+const handleSort = (provider, setSearchParams, key, order) => {
```
dmistry1, line 1: Add jsdoc

dmistry1 (approved):

static/src/js/pages/SearchList/SearchList.jsx
```diff
@@ -1,36 +1,40 @@
...
+import pluralize from 'pluralize'
+import PropTypes from 'prop-types'
 import React, {
-  useEffect,
   useCallback,
+  useEffect,
   useState
 } from 'react'
-import PropTypes from 'prop-types'
+import Row from 'react-bootstrap/Row'
```
cgokey, line 13: move this import to where the other react-bootstrap are

cgokey (approved):

PR #1270, opened 2024-08-01
MMT-3839: Add support for Science Keyword Picker to add a higher level keyword even if child keywords exist

dmistry1:
# Overview

### What is the feature?

Currently in production, we have always been able to add a higher level keyword even if child keywords exist. For example: EARTH SCIENCE SERVICES > DATA MANAGEMENT/DATA HANDLING > DATA CUSTOMIZATION can be curated as-is in production. In UAT, I am unable to add this keyword. It looks like I need to add one of its child keywords, INTERPOLATION, REPROJECTION, RESAMPLING. Currently our tools catalog uses CUSTOMIZATION as a filter option.

### What is the Solution?

Removed the disabled button and let the user add a keywords at any given level

# Checklist

### DIFF
--- a/static/src/js/components/KeywordPicker/KeywordPicker.jsx
+++ b/static/src/js/components/KeywordPicker/KeywordPicker.jsx
@@ -44,8 +44,8 @@ const KeywordPicker = ({
   const [fullPath, setFullPath] = useState([])
   const [loading, setLoading] = useState(true)
   const [finalSelectedValue, setFinalSelectedValue] = useState(null)
-  const [finalSelectedList, setFinalSelectedList] = useState([])
-  const [disableButton, setDisableButton] = useState(true)
+  const [currentSelected, setCurrentSelected] = useState([])
+
   const [marginTop, setMarginTop] = useState(0)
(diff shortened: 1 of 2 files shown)


mandyparson:
The issue here is that the functionality gets a little weird after clicking 'add keyword'. See the screen shot. We lose the upper level keywords by accident if a user tries to continue to add keywords. Is there any way to reset the widget after a user clicks 'add keyword'?
Screenshot 2024-08-01 at 11 46 41 AM

static/src/js/components/KeywordPicker/KeywordPicker.jsx
```diff
@@ -441,8 +441,8 @@ const KeywordPicker = ({
 
             <Button
               className="mt-2"
-              disabled={disableButton}
               onClick={handleSubmit}
+              aria-label="add"
```
cgokey, line 445: alphabetical order?

cgokey (approved):

mandyparson (approved):

cgokey (approved):

PR #1273, opened 2024-08-07
MMT-3845: Error message when paginating through Order Options

mandyparson:
# Overview

### What is the feature?

When users were paginating through order options, it would throw an error on page 2. Turns out that page 2 had an order option on it from a provider called ETIM22 which is not available to the public. It was decided to create the functionality to ONLY see order options from providers that a user had access to (as seen under their profile). 

### What is the Solution?

I utilized useAvailableProviders() to grab the available providers and passed that array of providerIds into the existing GET_ORDER_OPTIONS query. 

OrderOptionList.jsx

# Testing

### Reproduction steps

- **Environment for testing: SIT
- **Collection to test with: N/A

1. Set up your local environment for sit testing. If you are getting an internal error 500 let me know as I have a doc on how to fix this now. 
2. Go to your user profile and select 'providers' to take note of your personal available providers
3. Go to Order Options >> All order options
4. Notice that there is now a banner at the top letting you know which providers you are viewing order options for. You can swap out line 42 with const providerId = ['MMT_2'] or null or whatever you'd like to test out any other functionality. 
5. Regardless of what you choose, you should now be able to paginate freely through the options without running into an error. 

### Attachments

# Checklist

### DIFF
--- a/static/src/js/components/OrderOptionList/OrderOptionList.jsx
+++ b/static/src/js/components/OrderOptionList/OrderOptionList.jsx
@@ -1,10 +1,16 @@
-import React, { useCallback, useState } from 'react'
+import {
+  FaEdit,
+  FaExclamationCircle,
+  FaQuestionCircle,
+  FaTrash
+} from 'react-icons/fa'
+import { Alert } from 'react-bootstrap'
 import { useMutation, useSuspenseQuery } from '@apollo/client'
-import Col from 'react-bootstrap/Col'
-import Row from 'react-bootstrap/Row'
 import { useSearchParams } from 'react-router-dom'
+import Col from 'react-bootstrap/Col'
 import moment from 'moment'
-import { FaEdit, FaTrash } from 'react-icons/fa'
+import React, { useCallback, useState } from 'react'
+import Row from 'react-bootstrap/Row'
 
 import Button from '@/js/components/Button/Button'
 import ControlledPaginatedContent from '@/js/components/ControlledPaginatedContent/ControlledPaginatedContent'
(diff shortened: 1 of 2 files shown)


static/src/js/components/OrderOptionList/OrderOptionList.jsx
```diff
@@ -208,6 +220,26 @@ const OrderOptionList = () => {
 
               return (
                 <>
+                  {
+                    (!providerIds || providerIds.length === 0)
+                      ? (
+                        <Alert variant="info" className="mb-4">
```
dmistry1, line 226: I like this implementation of showing which providers we are showing, but this list page is different than all of the other list page. Don't really have a strong opinion, curious to know what everyone else thinks.

static/src/js/components/OrderOptionList/__tests__/OrderOptionList.test.jsx
```diff
@@ -99,6 +106,7 @@ describe('OrderOptionList', () => {
     test('render a table with 2 order options', async () => {
       setup({})
 
+      expect(await screen.findByText(/You are viewing order options for the following providers:/)).toBeInTheDocument()
```
dmistry1, line 109: Nitpicky here, I would also check to to see if the correct provider is showing. 

`expect(await screen.findByText(/You are viewing order options for the following providers: MMT_2/)).toBeInTheDocument()`

dmistry1 (approved):

cgokey (approved):

static/src/js/components/OrderOptionList/OrderOptionList.jsx
mandyparson: Right and that's because this list page is different from all the others. Other list pages don't require a user to have providers. Anyone can see anyone else's records. This particular list is being rendered specifically from the providers a user has available and nothing more.

PR #1274, opened 2024-08-07
MMT-3848: Order options fails in Forms XML for character limit although the character limit wasn't exceeded

cgokey:
# Overview

### What is the feature?

Order options fails in Forms XML for character limit although the character limit wasn't exceeded
Also noticed that on form load, existing errors would not appear.

### What is the Solution?

Updated order options schema and removed the char limits.   Also made updates to the form logic so that only "new" records have special logic where it hides the error until the field is visited.

Order Options and Groups Form.

# Testing

Easiest way to test this is to use CMR graphql to ingest a order option where the form xml > 1024 characters.
It should now load into the form, not show an error, and allow you to publish a order option with a length more than 1024 characters.

To test the other bug fix, upload an order option where the name is > 85 characters directly via CMR graphql.  When the form comes up, you should see an error.   Same thing can be done with group page, upload a new group where the name is > 85 characters, the form should show an error when loading into MMT form.

### Attachments

# Checklist

### DIFF
--- a/static/src/js/components/OrderOptionForm/__tests__/OrderOptionForm.test.jsx
+++ b/static/src/js/components/OrderOptionForm/__tests__/OrderOptionForm.test.jsx
@@ -301,6 +301,7 @@ describe('OrderOptionForm', () => {
           await screen.findByRole('textbox', { name: 'Name' })
         })
 
+        // Sort key's length is > 5 characters
         expect(await screen.findByText('must NOT have more than 5 characters')).toBeInTheDocument()
       })
     })
--- a/static/src/js/components/OrderOptionForm/__tests__/OrderOptionForm.test.jsx
+++ b/static/src/js/components/OrderOptionForm/__tests__/OrderOptionForm.test.jsx
@@ -2,6 +2,7 @@ import React, { Suspense } from 'react'
 import {
   render,
   screen,
+  waitFor,
   within
 } from '@testing-library/react'
 import { MockedProvider } from '@apollo/client/testing'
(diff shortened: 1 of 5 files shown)


dmistry1 (approved):

static/src/js/components/OrderOptionForm/__tests__/OrderOptionForm.test.jsx
```diff
@@ -255,6 +256,55 @@ describe('OrderOptionForm', () => {
...
+                    __typename: 'OrderOption'
+                  }
+                }
+              }
+            }]
+          }
+        )
+
+        await waitFor(async () => {
+          await screen.findByRole('textbox', { name: 'Name' })
```
mandyparson, line 301: Why do we need the extra awaits? Seems to still work with just expect(screen.findByRole('textbox', { name: 'Name' }))

static/src/js/components/OrderOptionForm/__tests__/OrderOptionForm.test.jsx
```diff
@@ -255,6 +256,55 @@ describe('OrderOptionForm', () => {
...
+        })
+
+        expect(await screen.findByText('must NOT have more than 5 characters')).toBeInTheDocument()
```
mandyparson, line 304: Shouldn't it be 85?

static/src/js/components/OrderOptionForm/__tests__/OrderOptionForm.test.jsx
cgokey: The form data would not load without this await, Deep and I both looked at this, only way we could get this to work.

static/src/js/components/OrderOptionForm/__tests__/OrderOptionForm.test.jsx
cgokey: No, this is testing sortKey length, I'll add a comment above to make that clear though.

mandyparson (approved):

PR #1284, opened 2024-08-14
MMT-3860: There are several areas in which the drop-down items are sh…

cgokey:
# Overview

### What is the feature?

There are several areas in which the drop-down items are showing up as Option 1, Option 2, Option 3, these need to removed.

### What is the Solution?

UI layout needed `ui:fieldReplacesAnyOrOneOf': true` anywhere there was an anyOf and oneOf used as a constraint.

Collection Form:

- Collection Information 
- Temporal Information

Service Form:
- Service Operations

# Testing

Steps to repeat:
1) Create a new collection draft
2) Navigate over to Temporal Information
3) Find Temporal Extents
4) If you click it, you will see Options 1, 2, 3 and click it yields  "Null is not an object (evaluating ‘ar[Xr]’" and you can't navigate back, thus losing any unsaved changed.

The other place we are seeing Option 1, 2, 3 is in Spatial Information.
Steps to repeat:
Navigate to:
https://mmt.sit.earthdatacloud.nasa.gov/drafts/collections/CD1200483084-EDF_OPS/spatial-information
There is a error in the right hand navigation, see screenshot.
Also if you look down at:
Non Gridded Range Resolutions
Gridded Resolutions
Gridded Range Resolutions
Generic Resolutions
We will also see the Option 1, 2, 3 combo box.  

# Checklist

### DIFF
--- a/static/src/js/components/CustomTitleField/CustomTitleField.scss
+++ b/static/src/js/components/CustomTitleField/CustomTitleField.scss
@@ -9,6 +9,7 @@
 
   &__heading {
     font-weight: 700;
+    margin-top: 1em;
   }
 
   &__icon {
--- a/static/src/js/components/CustomTitleField/CustomTitleField.scss
+++ b/static/src/js/components/CustomTitleField/CustomTitleField.scss
@@ -8,8 +8,8 @@
   }
 
   &__heading {
-    font-weight: 700;
     margin-top: 1em;
+    font-weight: 700;
   }
 
   &__icon {
--- a/static/src/js/components/CustomTitleField/CustomTitleField.scss
+++ b/static/src/js/components/CustomTitleField/CustomTitleField.scss
@@ -8,7 +8,7 @@
   }
 
   &__heading {
-    margin-top: 1em;
+    margin-top: 1rem;
     font-weight: 700;
   }
 
--- a/static/src/js/components/MetadataForm/MetadataForm.scss
+++ b/static/src/js/components/MetadataForm/MetadataForm.scss
@@ -14,15 +14,13 @@
 .rjsf {
   .custom-array-field-template__description,
   .field-description {
-    margin-top: 1rem;
-    margin-bottom: 2rem;
+    margin-top: 0.5rem;
+    margin-bottom: 0.5rem;
     color: $gray-800;
     font-size: 0.925rem;
   }
 
   .row.row-children {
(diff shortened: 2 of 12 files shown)


static/src/js/components/MetadataForm/MetadataForm.scss
```diff
@@ -14,15 +14,13 @@
...
-    margin-top: 1rem;
-    margin-bottom: 2rem;
+    margin-top: 0.5rem;
+    margin-bottom: 0.5rem;
     color: $gray-800;
     font-size: 0.925rem;
   }
 
   .row.row-children {
-    margin-bottom: 3rem;
```
mandyparson, line 24: I see why you got rid of this, but I think it's necessary to at least keep some padding (perhaps a 1rem?) As it removes all padding from certain areas (ie. Collection >> Data Identification >> Collection Data Type)

static/src/js/components/CustomTitleField/CustomTitleField.scss
```diff
@@ -8,6 +8,7 @@
   }
 
   &__heading {
+    margin-top: 1em;
```
mandyparson, line 11: Keep with rem for consistency

static/src/js/schemas/uiSchemas/collections/collectionInformation.js
```diff
@@ -69,6 +69,10 @@ const collectionInformationUiSchema = {
               'ui:row': [
                 {
                   'ui:col': {
+                    style: {
+                      padding: '1.25rem 1.5rem',
```
mandyparson, line 73: Can we pull this styling out as a variable as we are using it a couple times here

static/src/js/components/MetadataForm/MetadataForm.scss
cgokey: No, it has some bad side affects in a lot of places.   Try adding it back in and you'll see what I mean.  I'm not saying the extra padding shouldn't be solved, just saying this way of doing it I things causes more issues than it helps.

static/src/js/components/CustomTitleField/CustomTitleField.scss
cgokey: Good catch.

mandyparson (approved):

dmistry1 (approved):

PR #1296, opened 2024-09-06
MMT-3877: Cloning bug fix

mandyparson:
# Overview

### What is the feature?

When I try to clone a collection and edit, I get the following error (see attached screen shot for details). 
Sorry! Cannot destructure property 'nativeId' of 'Tr' as it is null. See the collection record: 
https://mmt.uat.earthdatacloud.nasa.gov/drafts/collections/CD1269208926-MMT_2 

### What is the Solution?

There are three use cases I'm trying to solve with this:
1. Render the old banner when appropriate 
Screenshot 2024-09-06 at 2 05 39 PM

2. Render an error banner that says "Sorry, the record does not exist" instead of 'Cannot deconstruct 'granules' of 'Cr' as draft is null.' when a record doesn't exist (this is based on the currentURL)
Screenshot 2024-09-06 at 2 08 37 PM

3. Render a prompt that tells users to refresh the page if they've come from trying to clone or edit a published record (this is based on the previousURL)
Screenshot 2024-09-06 at 2 09 22 PM

ErrorBanner, ErrorBoundary, and PublishPreview

# Testing

### Reproduction steps

- **Environment for testing: you may run this locally toggle lines 23 and 24 in ErrorBanner to get all use cases

1. Make sure application runs as expected when spinning up locally. As in, you can click 'clone' on a published record and it works. 
2. Toggler line 27 in ErrorBoundary to if (!hasError) to see basic error.
3. Toggle line 23 in ErrorBanner to see what happens when the logic decides that cmr is lagging (***please let me know of any other use cases of cmr lagging we'd want to include here)
4. Toggle line 24 to see what happens if the record doesn't exist. 

### Attachments

# Checklist

- [ x] My changes generate no new warnings

### DIFF
--- a/static/src/js/components/ErrorBanner/ErrorBanner.jsx
+++ b/static/src/js/components/ErrorBanner/ErrorBanner.jsx
@@ -24,9 +24,9 @@ export const ErrorBanner = ({
   const currentURL = window.location.pathname
   const conceptKeywords = ['/collections/', '/variables/', '/services/', '/tools/']
 
-  // Checks to see that the endpoint is correct,
+  // Checks to see that the url is a concept URL,
   // indicating that graphQL made a call that returned null
-  const graphQLCallReturnsNull = (url) => {
+  const isConceptUrl = (url) => {
     if (conceptKeywords.some((keyword) => url.includes(keyword))) {
       return true
     }
@@ -35,8 +35,7 @@ export const ErrorBanner = ({
   }
 
   return (
-    // Checks to see if users came from a published record, indicating there is a CMR lag
-    (previousURL && graphQLCallReturnsNull(previousURL)) ? (
+    (previousURL && isConceptUrl(previousURL)) ? (
       <div>
         <Alert className="fst-italic fs-6" variant="warning">
           <i className="eui-icon eui-fa-info-circle" />
@@ -70,7 +69,7 @@ export const ErrorBanner = ({
 
             <span className="visually-hidden">{' '}</span>
 
-            <p data-testid={dataTestId}>{graphQLCallReturnsNull(currentURL) ? 'This record does not exist' : message}</p>
+            <p data-testid={dataTestId}>{isConceptUrl(currentURL) ? 'This record does not exist' : message}</p>
           </Alert>
         </Col>
       </Row>
--- a/static/src/js/components/ErrorBanner/ErrorBanner.jsx
+++ b/static/src/js/components/ErrorBanner/ErrorBanner.jsx
@@ -1,7 +1,8 @@
-import React from 'react'
-import PropTypes from 'prop-types'
+import { Button } from 'react-bootstrap'
 import Alert from 'react-bootstrap/Alert'
 import Col from 'react-bootstrap/Col'
+import PropTypes from 'prop-types'
+import React, { useEffect, useState } from 'react'
 import Row from 'react-bootstrap/Row'
 
 /**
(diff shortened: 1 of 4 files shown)


static/src/js/components/PublishPreview/PublishPreview.jsx
```diff
@@ -418,7 +418,7 @@ const PublishPreview = ({ isRevision }) => {
           </Row>
         )
       }
-      <ErrorBoundary>
+      <ErrorBoundary previousURL={window.location.pathname}>
```
cgokey, line 421: since this is a global variable already, do we need to include it as a prop?

mandyparson: yes because this capures the previousURL. If I were to call window.location.pathname in the file that this will be utilized, it would call the currentURL. I need this one so that the component that renders the error knows the user came from publishpreview

static/src/js/components/ErrorBanner/ErrorBanner.jsx
```diff
@@ -16,28 +17,77 @@ import Row from 'react-bootstrap/Row'
...
+  const conceptKeywords = ['/collections/', '/variables/', '/services/', '/tools/']
+
+  // Catches if we are coming from a published record, indicating this is a CMR lag error
+  useEffect(() => {
+    if (previousURL && conceptKeywords.some((keyword) => previousURL.includes(keyword))) {
+      setHasCMRLagError(true)
+    }
 
-        <span className="visually-hidden">{' '}</span>
+    if (currentURL && conceptKeywords.some((keyword) => currentURL.includes(keyword))) {
```
cgokey, line 34: do you need currentURL here?

cgokey (approved):

static/src/js/components/ErrorBanner/ErrorBanner.jsx
cgokey: Thinking about this some more.   I don't think you need a state variable for this second case.  I think maybe helper function, something like isConceptUrl() and then use that helper in the "else" part of your render.

static/src/js/components/ErrorBanner/ErrorBanner.jsx
```diff
@@ -16,28 +17,77 @@ import Row from 'react-bootstrap/Row'
...
+  previousURL
+}) => {
+  const [hasCMRLagError, setHasCMRLagError] = useState(false)
+  const [conceptIdDoesNotExist, setConceptIdDoesNotExist] = useState(false)
+  const currentURL = window.location.pathname
+  const conceptKeywords = ['/collections/', '/variables/', '/services/', '/tools/']
+
+  // Catches if we are coming from a published record, indicating this is a CMR lag error
+  useEffect(() => {
+    if (previousURL && conceptKeywords.some((keyword) => previousURL.includes(keyword))) {
```
cgokey, line 30: See comment below, but you can use the same helper here as well, just pass it the previousURL

static/src/js/components/ErrorBanner/ErrorBanner.jsx
```diff
@@ -16,28 +18,75 @@ import Row from 'react-bootstrap/Row'
...
+  message,
+  previousURL
+}) => {
+  const currentURL = window.location.pathname
+  const conceptKeywords = ['/collections/', '/variables/', '/services/', '/tools/']
 
-        <span className="visually-hidden">{' '}</span>
+  // Checks to see that the endpoint is correct,
+  // indicating that graphQL made a call that returned null
+  const graphQLCallReturnsNull = (url) => {
```
cgokey, line 29: maybe isConceptUrl(url) -- what is the significant or returns null in the method name?

mandyparson: I initially thought it sounded too vague and wanted to be more specific. However this is probably leaning towards being more misleading, I can't confirm that data is null. Though that is the most likely cause. Changed back!

htranho (approved):

PR #1297, opened 2024-09-11
Mmt 3877b: Error When Cloning and Editing a Draft Collection

mandyparson:
# Overview

### What is the bug?
Errors occur during two particular use cases:

Collections >> Collection >> Clone >> 'Cannot destructure property \'nativeId\' of \'Tr\' as it is null.' (this error comes from DraftPage.jsx line 90 when MMT tries to ask for the draft but the draft doesn't exist in CMR yet)

Collections >> Collection >> Clone >> Edit >> Save & Publish >> 'Cannot destructure property \'granules\' of \'Cr\' as it is null.' (this error comes from PublishPreview.jsx line 101 when MMT tries to ask for the published record but it doesn't exist in CMR yet) 

Please summarize the feature or fix.

Initially, the thought was that we should use useHistory for this bug. If a user is coming from publishpreview or is coming from draftpage, then render a refresh banner. But there are two issues with this:
1) useHistory is not supported for react-router v6 or above. https://stackoverflow.com/questions/63471931/using-history-with-react-router-dom-v6 it is recommended that you use useNavigation. If you looke up history npm, the use cases are only for actions and blocking certain transitions. There does not appear to be an array of pash urls saved for reference. 
2) Even if I could access history, I don't believe it is wise to use a url as a metric for what error to throw. What if a user does another type of action on PublishPreview or DraftPage that causes an error? It would throw a refresh banner no matter what. 

Rather than use useHistory, I have opted to render a refresh banner based on those repeatable errors I listed above. 

### What is the Solution?

Adding knownCMRLagErrors to ErrorBanner

ErrorBanner.jsx

# Testing

### Reproduction steps

- **Environment for testing: SIT or UAT
- **Collection to test with: Any

1. From a collection, click clone until you receive the refresh page error message. Let me know if you get 'record does not exist'
2. From a clone, click edit, add in a short name and entry title, then click Save & Publish until you receive the refresh page error message. Again, let me know if you receive any other error. 
3. Try to trigger error banners in any other areas around the app. I want to be sure this fix doesn't accidentally affect other areas that it shouldn't. 

### Attachments

# Checklist

### DIFF
--- a/static/src/js/components/ErrorBanner/ErrorBanner.jsx
+++ b/static/src/js/components/ErrorBanner/ErrorBanner.jsx
@@ -19,11 +19,7 @@ export const ErrorBanner = ({
   dataTestId,
   message
 }) => {
-  const knownCMRLagErrors = [
-    'Cannot destructure property \'granules\' of \'concept\' as it is null.',
-    'Cannot destructure property \'granules\' of \'Cr\' as it is null.',
-    'Cannot destructure property \'nativeId\' of \'Tr\' as it is null.',
-    'Cannot destructure property \'nativeId\' of \'draft\' as it is null.']
+  const knownCMRLagErrors = ['draft is null', 'concept is null']
 
   // Checks to see if error message is a known CMR Lag Error Message
   const isKnownCMRLagError = () => {
--- a/static/src/js/components/ErrorBanner/ErrorBanner.jsx
+++ b/static/src/js/components/ErrorBanner/ErrorBanner.jsx
@@ -9,7 +9,6 @@ import Row from 'react-bootstrap/Row'
  * @typedef {Object} ErrorBannerProps
  * @property {String} dataTestId A data-testid for the error message
  * @property {String} message A message displaying what error has occurred.
- * @property {String} previousURL A string that is sent down from PublishPreview only
  */
 
 /**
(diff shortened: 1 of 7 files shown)


static/src/js/pages/DraftPage/DraftPage.jsx
```diff
@@ -278,7 +278,7 @@ const DraftPage = () => (
     pageType="secondary"
     header={<DraftPageHeader />}
   >
-    <ErrorBoundary>
+    <ErrorBoundary previousURL={window.location.pathname}>
```
cgokey, line 281: you don't need previousURL here right, it is now `message=`

htranho (approved):

static/src/js/components/ErrorBanner/ErrorBanner.jsx
```diff
@@ -18,30 +17,31 @@ import Row from 'react-bootstrap/Row'
  */
 export const ErrorBanner = ({
   dataTestId,
-  message,
-  previousURL
+  message
 }) => {
-  const currentURL = window.location.pathname
-  const conceptKeywords = ['/collections/', '/variables/', '/services/', '/tools/']
+  const knownCMRLagErrors = [
```
cgokey, line 22: doesn't the error boundary have access to the stack trace?  maybe you can check for a specific kind of exception instead of a string comparison, which seems fragile.

cgokey (changes requested):
doesn't the error boundary have access to the stack trace? maybe you can check for a specific kind of exception instead of a string comparison, which seems fragile.

static/src/js/components/ErrorBanner/ErrorBanner.jsx
mandyparson: Here's the stack trace. You can see there are actually two different errors that populate here. I'm looking around for other methods aside from .message I could call on error to get different kinds of exceptions but am coming up blank. Did you have something in mind?
Screenshot 2024-09-13 at 12 16 53 PM

cgokey (approved):

PR #1298, opened 2024-09-16
SAP-129: Saving Collection Permissions ACLs is removing all but 20 selected collections

cgokey:
# Overview

### What is the feature?

There is a bug in the PermissionForm where it would only load the first 20 selected collections, even though the user may have had 100s.   As a result, when the user saved the permission form, it would only save the first 20 resulting in the rest of the collections be removed from the ACL.

### What is the Solution?

As a temporary solution, I've bumped up the limit parameter to graphql to pull up to 2000 collections.   I've also bumped up the available collections limit to 100, as users where stating that 20 was not large enough list size, due to some searches by short name having more than 20.

A permanent solution to this ticket will be to support more than 2000 selected collections but will require a bit more work.   Also the users would like the search within available collections to be more than just short name.

The collection permissions form.

# Testing

Find a collection permission with more than 20 selected collections and verify you can see them all in the selected collections pane.   Also verify the available collections pane shows 100 filtered collections.

# Checklist

### DIFF
--- a/static/src/js/components/CollectionSelector/__tests__/CollectionSelector.test.jsx
+++ b/static/src/js/components/CollectionSelector/__tests__/CollectionSelector.test.jsx
@@ -29,7 +29,11 @@ const setup = ({
       [{
         request: {
           query: GET_PERMISSION_COLLECTIONS,
-          variables: {}
+          variables: {
+            params: {
+              limit: 100
+            }
+          }
         },
         result: {
           data: {
@@ -229,7 +233,7 @@ describe('CollectionSelector', () => {
               params: {
                 options: { shortName: { pattern: true } },
                 shortName: 'C*',
-                limit: 20
+                limit: 100
               }
             }
           },
@@ -266,7 +270,7 @@ describe('CollectionSelector', () => {
 
       await user.type(searchField, 'C')
 
-      expect(await screen.findByText('Showing 1 items')).toBeInTheDocument()
+      expect(await screen.findByText('Showing 2 items')).toBeInTheDocument()
     })
   })
 
@@ -297,7 +301,7 @@ describe('CollectionSelector', () => {
             variables: {
               params: {
                 keyword: 'Col',
-                limit: 20
+                limit: 100
               }
             }
           },
--- a/static/src/js/components/PermissionForm/PermissionForm.jsx
+++ b/static/src/js/components/PermissionForm/PermissionForm.jsx
@@ -220,7 +220,10 @@ const PermissionForm = () => {
   const { data } = useSuspenseQuery(GET_COLLECTION_FOR_PERMISSION_FORM, {
     skip: conceptId === 'new',
     variables: {
-      conceptId
+      conceptId,
+      params: {
+        limit: 2000
+      }
     }
   })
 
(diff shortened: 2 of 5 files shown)


static/src/js/components/CollectionSelector/__tests__/CollectionSelector.test.jsx
```diff
@@ -229,7 +233,7 @@ describe('CollectionSelector', () => {
               params: {
                 options: { shortName: { pattern: true } },
                 shortName: 'C*',
-                limit: 20
+                limit: 100
```
jceaser, line 236: why 100 and not bigger?

static/src/js/components/PermissionForm/PermissionForm.jsx
```diff
@@ -220,7 +220,10 @@ const PermissionForm = () => {
   const { data } = useSuspenseQuery(GET_COLLECTION_FOR_PERMISSION_FORM, {
     skip: conceptId === 'new',
     variables: {
-      conceptId
+      conceptId,
+      params: {
+        limit: 2000
```
jceaser, line 225: so this is 2000, but the others are 100?

dmistry1 (approved):

jceaser (approved):
Approved, but I ask you consider why 100 is used.

static/src/js/components/CollectionSelector/__tests__/CollectionSelector.test.jsx
cgokey: Just makes a very large list, 100 was 5x more than what was previously used, so I figured I was pushing the UI/UX concept of that box as it was.

cgokey: The real solution is to allow better search criteria, right now it is just short name.

static/src/js/components/PermissionForm/PermissionForm.jsx
cgokey: Yeah this was the query that resulted in the corrupting ACLs.  before it only pulled in 20 and then if the user saved, it would only save 20 even though they had 100s in the database

static/src/js/components/CollectionSelector/__tests__/CollectionSelector.test.jsx
jceaser: what is the cost in supporting 2000?

cgokey: I think it is more of a usability issue vs cost.  smaller list, easier to find results, etc.

mandyparson (approved):

static/src/js/components/PermissionForm/PermissionForm.jsx
jceaser: answered in person

static/src/js/components/CollectionSelector/__tests__/CollectionSelector.test.jsx
jceaser: answered in person.

PR #1301, opened 2024-09-23
MMT-3895: Now pulls in ALL selected collections.

cgokey:
# Overview

### What is the feature?

A user reported that its primary access policy that exposes > 800 collections in ops was reduced to only 20 collections after a update the ACL.

### What is the Solution?

The selected collections pane was only retrieving 20 collections, even though the ACL could have many more.   Therefore, when the ACL was updated, it would only update with the 20 collections it retrieved and not the full list of collections in the ACL.

The solution is to use fetchMore provided by Apollo to request all the collections in the ACL.  

The collection permissions page.

# Testing

Try finding a ACL that has over 20 collections (or create a ACL that includes more than 20).   Verify you can add and remove collections.   Verify all collections you add show up in the right pane and the submit works properly with your additions and subtractions.

# Checklist

### DIFF
--- a/static/src/js/components/CollectionSelector/__tests__/CollectionSelector.test.jsx
+++ b/static/src/js/components/CollectionSelector/__tests__/CollectionSelector.test.jsx
@@ -270,7 +270,7 @@ describe('CollectionSelector', () => {
 
       await user.type(searchField, 'C')
 
-      expect(await screen.findByText('Showing 2 items')).toBeInTheDocument()
+      expect(await screen.findByText('Showing 1 items')).toBeInTheDocument()
     })
   })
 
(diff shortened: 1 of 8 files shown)


static/src/js/components/CollectionSelector/__tests__/CollectionSelector.test.jsx
```diff
@@ -270,7 +270,7 @@ describe('CollectionSelector', () => {
 
       await user.type(searchField, 'C')
 
-      expect(await screen.findByText('Showing 2 items')).toBeInTheDocument()
+      expect(await screen.findByText('Showing 1 items')).toBeInTheDocument()
```
mandyparson, line 273: Why did this change to 1 again?

mandyparson (approved):

htranho (approved):

static/src/js/components/CollectionSelector/__tests__/CollectionSelector.test.jsx
cgokey: I'm not sure, but the mock only has 1 collection as the result.   So it is correct.

PR #1303, opened 2024-09-25
MMT-3907: Fixed Attempting to Update a Collection Permission ACL With Collection From a Different Provider Silently Ignores the User Request

cgokey:
# Overview

### What is the feature?

As part of the process of testing saving ACLs with more than 20 selected collections ([MMT-3895](https://bugs.earthdata.nasa.gov/browse/MMT-3895)), there are some cases where trying to add collections to the selected collection list does not work properly.
In these cases, the collection gets added to the selected collection list and even when you click submit, you get a successful response back stating the ACL was saved properly.  But, when you look at the ACL permissions page, the added collections are not there.

The issue is the user is allowed to add any collection to the selected collection list, but the only ones that will work is those collections in the same provider as the ACL.  

### What is the Solution?

If the user is trying to update the ACL, we know the provider, so the solution is to just simply filter the available provider list by the ACL's provider so they can't choose an collection outside of their provider.

If the user is trying to create a new ACL, we do not know the provider until submission, so the solution is to run a check to verify that the selected collections match the selected provider associated with the ACL.

Collection Permissions

# Testing

Try editing an existing ACL permission, the available collections list should be filtered by the ACL's provider id.
Try creating a new ACL permission, the available collection list should show all collections (not filtered by any provider id).   But when you submit, the UI will ask you for the provider and then a check will occur to make sure all selected collections match that provider id.   If they do not, it will produce an error.

# Checklist


static/src/js/components/PermissionForm/PermissionForm.jsx
```diff
@@ -483,6 +491,16 @@ const PermissionForm = ({ selectedCollectionsPageSize }) => {
     })
   }
 
+  const providerIdCheckFails = (aclProviderId, conceptIds) => {
+    console.log('provider id=', aclProviderId)
```
cgokey, line 495: remove

htranho (approved):

static/src/js/components/PermissionForm/PermissionForm.jsx
```diff
@@ -546,6 +560,16 @@ const PermissionForm = ({ selectedCollectionsPageSize }) => {
       )
     }
 
+    if (providerIdCheckFails(providerId, conceptIds)) {
+      errorLogger('Error multiple providers in collections', 'PermissionForm: providerIdCheck')
+      addNotification({
+        message: 'Error multiple providers in collections',
```
abbottry, line 566: This error message isn't actionable, if we're going to display a message to the user, they need to be able to take an action to make the error go away. This message doesn't tell them what they need to do.

static/src/js/components/PermissionForm/PermissionForm.jsx
```diff
@@ -198,6 +198,14 @@ const PermissionForm = ({ selectedCollectionsPageSize }) => {
     }
   })
 
+  useEffect(() => {
+    if (providerId) {
+      const newUiSchema = cloneDeep(uiSchema)
+      newUiSchema.collectionSelection.selectedCollections['ui:providerId'] = providerId
```
abbottry, line 204: What is the purpose of modifying the UI schema, I'm not clear on the intentions.

static/src/js/components/Layout/Layout.jsx
```diff
@@ -236,7 +236,7 @@ const Layout = ({ className, displayNav }) => {
                             naked
                             Icon={FaUserAlt}
                           >
-                            {`${user.name} `}
+                            {`${user?.name} `}
```
abbottry, line 239: Were we seeing errors related to this?

abbottry (changes requested):
I'm confused by the solution here, I'm seeing some ui schema changes that I'm unsure the purpose of, but rather than preventing the user from doing something, we're just throwing an error expecting that they clean it up. We have a few pages in the app that use a provider dropdown, why not just use the dropdown to scope the collections that show up and then a vast majority of this PR can be undone.

static/src/js/components/Layout/Layout.jsx
cgokey: URS had some issues yesterday and it would not return the user name (MMT crashed in sit, uat, and prod).

static/src/js/components/PermissionForm/PermissionForm.jsx
cgokey: In order to pass the providerId to the collection selector field, it needs to go through the uischema.   This is so we can filter that left hand side (available collections) by the providerId.

static/src/js/components/PermissionForm/PermissionForm.jsx
cgokey: Suggestions?  This is the use case for creating a new ACL, where we don't have the provider id, until submission.   The old code was should just fail silently and showing a passing message even thought there were collections not in the same provider as the ACL.   We can make the message longer, if you have a better suggestion.

cgokey:
> I'm confused by the solution here, I'm seeing some ui schema changes that I'm unsure the purpose of, but rather than preventing the user from doing something, we're just throwing an error expecting that they clean it up. We have a few pages in the app that use a provider dropdown, why not just use the dropdown to scope the collections that show up and then a vast majority of this PR can be undone.

You want to add the provider id dropdown directly into the form?  It really only necessary for the new ACL use case.   We know the provider id already for the update use case.   It is just the "new ACL" that we don't know it until submission.   Most of the PR is actually just filtering the available collections by the provider id, the only code that will disappear is the provider id check for the new use case, but I agree, I think it makes sense to just add the combo box directly into the form rather than this check.

static/src/js/components/PermissionForm/PermissionForm.jsx
cgokey: I missed your comment, I'll add the provider id dropdown directly into the form.

cgokey: Actually, adding the provider id combo box doesn't really prevent this, as they can still select multiple collections with multiple provider ids.   So I'm going to keep the check, but if you have some suggestions as to how to word the error, that would be great.

static/src/js/components/PermissionForm/PermissionForm.jsx
```diff
@@ -37,6 +37,7 @@ import { cloneDeep } from '@apollo/client/utilities'
 import {
   GET_COLLECTION_FOR_PERMISSION_FORM
 } from '@/js/operations/queries/getCollectionForPermissionForm'
+import useAvailableProviders from '@/js/hooks/useAvailableProviders'
```
abbottry, line 40: When you come across files that are using a mix of `@` and `../` imports, please update them to use `@` -- this was introduced later in the effort and it didn't make sense to transition every file. So, convert the other imports to use `@` and alphabetize the imports please.

static/src/js/components/CollectionSelector/__tests__/CollectionSelector.test.jsx
```diff
@@ -31,7 +31,8 @@ const setup = ({
           query: GET_PERMISSION_COLLECTIONS,
           variables: {
             params: {
-              limit: 100
+              limit: 100,
+              provider: null
```
abbottry, line 35: When the provider dropdown is in use, I think @trevorlang was aiming to use it as the default, essentially IF someone had gone through the effort to select one, use the hook to pull the previously selected value when we present it again, rather than wiping it out. So, we should be using the providerId from the hook rather than wiping it out by default.

static/src/js/components/PermissionForm/PermissionForm.jsx
abbottry: I'd like to see a demo of this I'm not following. The ui schema is a ui facade for the data, the data should provide this type of information, but I could definitely be wrong.

static/src/js/components/PermissionForm/PermissionForm.jsx
abbottry: If CMR doesn't prevent this, we shouldn't be either. If there is a situation that we are allowing the user to do something that produces an error, then we need to _prevent it_ from happening. If CMR doesn't care about this, then we shouldn't be adding validation on top of CMR. 

IF CMR doesn't want this to happen, then CMR needs to make it return an error, and we can then file a ticket to prevent the user from creating data that will produce an error.

static/src/js/components/PermissionForm/PermissionForm.jsx
cgokey: Sure, there are multiple ways to get the provider id into the widget, but I can show you at least my ratio for why I choose this path.

cgokey: I removed this logic of passing providerId through uischema in favor or just grabbing it from `AppContext`.

static/src/js/components/PermissionForm/PermissionForm.jsx
cgokey: Ok removed.

static/src/js/components/CollectionSelector/CollectionSelector.jsx
```diff
@@ -28,6 +28,7 @@ import { useLazyQuery, useSuspenseQuery } from '@apollo/client'
 
 import { GET_PERMISSION_COLLECTIONS } from '@/js/operations/queries/getPermissionCollections'
 
+import useAppContext from '@/js/hooks/useAppContext'
```
abbottry, line 31: When we import files us `@` syntax into files that have imports not using it, we should be taking that opportunity to update the existing imports to use the newer `@` syntax.

static/src/js/components/CollectionSelector/__tests__/CollectionSelector.test.jsx
```diff
@@ -11,6 +11,7 @@ import userEvent from '@testing-library/user-event'
 
 import { GET_PERMISSION_COLLECTIONS } from '@/js/operations/queries/getPermissionCollections'
 
+import AppContext from '@/js/context/AppContext'
```
abbottry, line 14: When we import files us `@` syntax into files that have imports not using it, we should be taking that opportunity to update the existing imports to use the newer `@` syntax.

static/src/js/components/PermissionForm/PermissionForm.jsx
```diff
@@ -21,22 +21,23 @@ import saveTypesToHumanizedStringMap from '@/js/constants/saveTypesToHumanizedSt
...
 import { CREATE_ACL } from '@/js/operations/mutations/createAcl'
 import { UPDATE_ACL } from '@/js/operations/mutations/updateAcl'
-
-import CollectionSelectorPage from '@/js/pages/CollectionSelectorPage/CollectionSelectorPage'
-
-import { cloneDeep } from '@apollo/client/utilities'
 import {
   GET_COLLECTION_FOR_PERMISSION_FORM
 } from '@/js/operations/queries/getCollectionForPermissionForm'
+
```
abbottry, line 38: When we import files us `@` syntax into files that have imports not using it, we should be taking that opportunity to update the existing imports to use the newer `@` syntax.

static/src/js/components/PermissionForm/__tests__/PermissionForm.test.jsx
```diff
@@ -364,31 +364,29 @@ describe('PermissionForm', () => {
...
 
+        // Click group search
+        await user.click(combos[4])
         const option = screen.getByRole('option', { name: 'Mock group MMT_2' })
         await user.click(option)
+        await user.click(combos[4])
 
-        await user.click(groupSearch[3])
-
+        // Click Search order
```
abbottry, line 379: Wrong comment?

static/src/js/components/PermissionForm/__tests__/PermissionForm.test.jsx
```diff
@@ -1,8 +1,7 @@
 import {
   render,
   screen,
-  waitFor,
-  within
+  waitFor
```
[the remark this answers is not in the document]
cgokey: `waitFor` is used on line 488.

static/src/js/components/PermissionForm/__tests__/PermissionForm.test.jsx
cgokey: I removed it made no sense.

static/src/js/components/PermissionForm/PermissionForm.jsx
cgokey: done

static/src/js/components/CollectionSelector/__tests__/CollectionSelector.test.jsx
cgokey: done

static/src/js/components/CollectionSelector/CollectionSelector.jsx
cgokey: done

static/src/js/components/CollectionSelector/__tests__/CollectionSelector.test.jsx
cgokey: done

static/src/js/components/PermissionForm/PermissionForm.jsx
cgokey: done

abbottry (approved):

PR #1316, opened 2024-10-25
MMT-3918: Cleanup legacy group permissions when displaying and submitting ACLs

htranho:
# Overview

### What is the feature?

Legacy group permissions causes issues in graphql when submitting update to ACL

### What is the Solution?

Remove legacy group permissions when displaying and submitting ACLs update.

Editing permissions: Group permission search boxes

# Testing

See test session

### Attachments

See ticket MMT-3918.

# Checklist


dmistry1 (commented):
I see there are two utils where they do the exact same but one checks on `id` and the other checks for `groupId`. You can combine these two by passing the key ('id' or 'groupId') as a parameter into the util.

Also since `items` is an array you can use `.filter()` and avoid `forEach`

```
const removeLegacyPermissions = (groupPermissions, key) => {
  return groupPermissions?.filter(item => !item[key] || validUUID.test(item[key])) || [];
}
```
Then you can call in PermissionForm.jsx using :
`removeLegacyPermissions(permission, 'id')` (For the `id` you are going to have to pull items out of `groups` and pass `items` to the util. 
`removeLegacyPermissions(permission, 'groupId')`

This will reduce code duplication, and using .filter() is more efficient than using .forEach(). With .filter(), we can directly return the resulting array without needing to create and manually populate a new array.

dmistry1 (approved):

mandyparson (approved):

static/src/js/components/PermissionForm/PermissionForm.jsx
```diff
@@ -359,9 +359,15 @@ const PermissionForm = ({ selectedCollectionsPageSize }) => {
       const searchAndOrderGroupPermission = []
       const searchPermission = []
 
+      // Removes invalid group permission items, which has no id and no userType
+      // eslint-disable-next-line max-len
+      const removeInvalidGroupPermissionItems = (groupPermissionItems) => groupPermissionItems?.filter((item) => item.id || item.userType)
```
cgokey, line 364: Add a comment, stating it returns new array with invalid groups removed.

cgokey, line 364: I think you should add a test for this logic as well.

htranho: Done

htranho: Added

static/src/js/components/PermissionForm/__tests__/PermissionForm.test.jsx
```diff
@@ -1468,4 +1468,270 @@ describe('PermissionForm', () => {
       })
     })
   })
+
+  describe('form validation', () => {
+    describe('when invalid group permission', () => {
```
cgokey, line 1473: maybe "when invalid group permission occur" 
(reads a little better)

cgokey (changes requested):
Changes look good, I would just modify the test wording slightly.

static/src/js/components/PermissionForm/__tests__/PermissionForm.test.jsx
htranho: Done

PR #1327, opened 2024-11-26
Mmt 3936: URS Login/Token Refresh Results in Blank Page

mandyparson:
# Overview

### What is the feature?

This ticket does not have reproduction steps and the only thing that I was able to find upon inspect is that this behavior occurs when switching between SIT and either UAT or PROD. Needed a way to keep the page from hanging when a user toggles between the two

### What is the Solution?

It did appear to be an issue with cookies. Added the environment to the MMT_COOKIE label to overcome

List impacted areas.

Anywhere that used MMT_COOKIE

### Reproduction steps

- **Environment for testing: local pointing to SIT and mmt.SIT
- **Collection to test with:none

1. open your localhost in one tab and your SIT on the other on main branch. Notice that they will both start to load up again when you delete all your cookies. 
2.Now do the same when switched onto my branch. You'll notice that when you delete cookies on one side, the other does not automatically load up again. 

### Attachments

None. 
# Checklist


static/src/js/components/TemplateList/__tests__/TemplateList.test.jsx
```diff
@@ -9,7 +9,7 @@ import errorLogger from '@/js/utils/errorLogger'
 import getTemplates from '@/js/utils/getTemplates'
 import deleteTemplate from '@/js/utils/deleteTemplate'
 import NotificationsContext from '@/js/context/NotificationsContext'
-import MMT_COOKIE from '@/js/constants/mmtCookie'
+import MMT_COOKIE from '../../../../../../sharedContstants/mmtCookie'
```
cgokey, line 12: This should use the @/js/constants way of doing things (absolute path) rather than relative paths (../../../ etc)

static/src/js/hooks/useMMTCookie.js
```diff
@@ -1,6 +1,6 @@
 import { useCookies } from 'react-cookie'
 
-import MMT_COOKIE from '../constants/mmtCookie'
+import MMT_COOKIE from '../../../../sharedContstants/mmtCookie'
```
cgokey, line 3: Same here

static/src/js/providers/AuthContextProvider/AuthContextProvider.jsx
```diff
@@ -10,13 +10,13 @@ import { useIdle } from '@uidotdev/usehooks'
...
 import AuthContext from '@/js/context/AuthContext'
 
-import MMT_COOKIE from '@/js/constants/mmtCookie'
-
 import useMMTCookie from '@/js/hooks/useMMTCookie'
 
 import errorLogger from '@/js/utils/errorLogger'
 import refreshToken from '@/js/utils/refreshToken'
 
+import MMT_COOKIE from '../../../../../sharedContstants/mmtCookie'
```
cgokey, line 18: Same here

static/src/js/providers/AuthContextProvider/AuthContextProvider.jsx
```diff
@@ -10,13 +10,13 @@ import { useIdle } from '@uidotdev/usehooks'
...
+
```
cgokey, line 19: Remove extra line

static/src/js/providers/AuthContextProvider/__tests__/AuthContextProvider.test.jsx
```diff
@@ -14,7 +14,7 @@ import refreshToken from '@/js/utils/refreshToken'
 
 import NotificationsContextProvider from '@/js/providers/NotificationsContextProvider/NotificationsContextProvider'
 
-import MMT_COOKIE from '@/js/constants/mmtCookie'
+import MMT_COOKIE from '../../../../../../sharedContstants/mmtCookie'
```
cgokey, line 17: Same here

cgokey (changes requested):
PR looks good.  We should just not use relative paths in imports, so these should be changed back to absolute paths.

htranho (approved):

static/src/js/components/TemplateList/__tests__/TemplateList.test.jsx
```diff
@@ -9,7 +9,7 @@ import errorLogger from '@/js/utils/errorLogger'
 import getTemplates from '@/js/utils/getTemplates'
 import deleteTemplate from '@/js/utils/deleteTemplate'
 import NotificationsContext from '@/js/context/NotificationsContext'
-import MMT_COOKIE from '@/js/constants/mmtCookie'
+import MMT_COOKIE from 'sharedConstants/mmtCookie'
```
cgokey, line 12: does `@/js/constants/mmtCookie` not work ?

mandyparson: Correct. That's only for files within static/src. I had to create a new route.

cgokey (approved):

static/src/js/components/TemplateList/__tests__/TemplateList.test.jsx
mandyparson: This variable was needed for files within serverless/src as well and Matthew said that we need to keep the static/src and serverless/src separate (no shared contestants or utils). To solve this, I needed to cretae a new folder called sharedConstants which is similar to sharedUtils.

PR #1334, opened 2025-01-02
MMT-3960: Resetting millisecond part only when date time is selected.

htranho:
# Overview

### What is the feature?

Millisecond part of date time input should accept input values.

### What is the Solution?

Preserve input values of millisecond part and reset it only when date time value is selected.   

All date time fields.

# Testing

1. Existing temporal extends should be preserved when loaded into the form.
2. Copy and pasted values should be preserved.
3. Selected values from the widget should have the millisecond part set to 0.

### Attachments

N/A

# Checklist

### DIFF
--- a/static/src/js/components/CustomDateTimeWidget/CustomDateTimeWidget.jsx
+++ b/static/src/js/components/CustomDateTimeWidget/CustomDateTimeWidget.jsx
@@ -93,9 +93,10 @@ const CustomDateTimeWidget = ({
   }
 
   const handleChange = (newDate) => {
-    // The picket widget has a bug where it is not setting millis to 0 when selecting a time,
-    // so we need to reset the millisecond part. When the date time string is pasted in or a value
-    // exists, do not reset.
+    // The picker widget has a bug where it is not setting the ms. to 0 when a user selects a time.
+    // We are working around this bug by setting the ms to 0 only if the user selects/chooses
+    // a time from the widget. If they type in a date/time or paste in a date/time should ,
+    // we should not perform this reset as we should keep whatever ms they user enters.
     if (!value && dateInputMethod !== 'typed') newDate.setMilliseconds(0)
     const formattedDateTime = fromZonedTime(newDate, 'GMT').toISOString()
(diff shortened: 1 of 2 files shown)


static/src/js/components/CustomDateTimeWidget/CustomDateTimeWidget.jsx
```diff
@@ -89,14 +93,21 @@ const CustomDateTimeWidget = ({
   }
 
   const handleChange = (newDate) => {
-    // The picket widget has a bug where it is not setting millis to 0 when selecting a time.
-    newDate.setMilliseconds(0)
+    // The picket widget has a bug where it is not setting millis to 0 when selecting a time,
```
cgokey, line 96: typo

static/src/js/components/CustomDateTimeWidget/CustomDateTimeWidget.jsx
```diff
@@ -89,14 +93,21 @@ const CustomDateTimeWidget = ({
...
+    // so we need to reset the millisecond part. When the date time string is pasted in or a value
```
cgokey, line 97: maybe, "the picker widget has a bug where it is not setting the ms. to 0 when a user selects a time.   We are working around this bug by setting the ms to 0 only if the user selects/chooses a time from the widget.   If they type in a date/time or paste in a date/time should , we should not perform this reset as we should keep whatever ms they user enters."

static/src/js/components/CustomDateTimeWidget/CustomDateTimeWidget.jsx
htranho: corrected

static/src/js/components/CustomDateTimeWidget/CustomDateTimeWidget.jsx
htranho: Yes, updated.

mandyparson (approved):

cgokey (approved):

PR #1345, opened 2025-03-27
MMT-3976: BUG | Some fields using CustomSelectWidget are not utilizing enum list

mandyparson:
# Overview

### What is the feature?

Go here: https://mmt.sit.earthdata.nasa.gov/drafts/collections/CD1200484449-MMT_2/related-urls?revisionId=2 and under mime-type, try to find 'application/geo+json', you'll notice it isn't there. This bug is actually in multiple places. The enum list isn't what it's supposed to be according to the scheme. 

### What is the Solution?

The problem was using ui:controlled for this particular MimeType case:
Screenshot 2025-03-27 at 1 33 49 PM

While in most cases we use the 'ui:controlled' field for MimeType, we need to use the enum list specifically for mimetypes where URLContentType=DistributionURL; Type=USE SERVICE API from this you get the GET SERVICE section - the mime-type in there uses the enumerations. The enumerations exist because we never got to migrate them over to use KMS.

Anywhere those lines above were used. 

# Testing

### Reproduction steps

- **Environment for testing:local, sit, uat whatever you can spin up

1. Make sure the schema and the changes match up and that you can see the correct format (and enum list if applicable) in the forms. 

### Attachments
MimeType Before
Screenshot 2025-03-27 at 1 37 32 PM

MimeType After (If enum was needed)
Screenshot 2025-03-27 at 1 43 21 PM

# Checklist

- [] I have commented my code, particularly in hard-to-understand areas
- [] I have made corresponding changes to the documentation

### DIFF
--- a/static/src/js/schemas/uiSchemas/collections/collectionCitation.js
+++ b/static/src/js/schemas/uiSchemas/collections/collectionCitation.js
@@ -1,3 +1,4 @@
+import CustomSelectWidget from '../../../components/CustomSelectWidget/CustomSelectWidget'
 import CustomTextareaWidget from '../../../components/CustomTextareaWidget/CustomTextareaWidget'
 
 const collectionCitationUiSchema = {
@@ -244,6 +245,13 @@ const collectionCitationUiSchema = {
         },
         Description: {
           'ui:widget': CustomTextareaWidget
+        },
+        MimeType: {
+          'ui:widget': CustomSelectWidget,
+          'ui:controlled': {
+            name: 'mime-type',
+            controlName: 'mime_type'
+          }
         }
       }
     }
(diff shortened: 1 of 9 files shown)


cgokey:
Is format and mime type suppose to use enums or pull there values from CMR?   I think that is the reason for `controlName`, as it tells the UI to build the combo box from CMR keywords.

mandyparson:
@cgokey and @dmistry1 it looks like for the specific instance that Erich mentions in his comment below, we need to use the enum list in the schema. So this ticket now address just that.

From Erich:
"The URLMimeTypeEnum exists only for RelatedURLs where the URLContentType=DistributionURL; Type=USE SERVICE API from this you get the GET SERVICE section - the mime-type in there uses the enumerations. The enumerations exist because we never got to migrate them over to use KMS."

htranho (approved):

PR #1365, opened 2025-05-19
MMT-3996: As a user, I can edit the Spatial Extent

mandyparson:
# Overview

### What is the feature?

Users can edit Spatial Extent

### What is the Solution?

Add spatialExtent.jsx

VisualizationConfiguration and UiSchemas

# Testing

### Reproduction steps

- **Environment for testing:SIT
- **Collection to test with:New draft or "Created Using Form" draft

1. Go to Spatial Extent and play around with all the different checkboxes and dropdowns for fields and make sure they all look correct. 

### Attachments

Screenshot 2025-05-19 at 10 55 49 AM

# Checklist

### DIFF
--- a/static/src/js/schemas/uiSchemas/visualizations/spatialExtent.jsx
+++ b/static/src/js/schemas/uiSchemas/visualizations/spatialExtent.jsx
@@ -1,0 +1,1171 @@
+import React from 'react'
+
+import CustomSelectWidget from '@/js/components/CustomSelectWidget/CustomSelectWidget'
+import PreviewMapTemplate from '@/js/components/PreviewMapTemplate/PreviewMapTemplate'
+
+const spatialExtentUiSchema = {
+  'ui:heading-level': 'h3',
+  'ui:field': 'layout',
+  'ui:layout_grid': {
+    'ui:row': [
+      {
+        'ui:group': 'Spatial Extent',
+        'ui:group-description': true,
+        'ui:col': {
+          md: 12,
+          children: [
+            {
+              'ui:row': [
+                {
+                  'ui:col': {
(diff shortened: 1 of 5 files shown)


static/src/js/schemas/uiSchemas/visualizations/spatialExtent.jsx
```diff
@@ -0,0 +1,1171 @@
+import React from 'react'
+
+import CustomSelectWidget from '@/js/components/CustomSelectWidget/CustomSelectWidget'
+import PreviewMapTemplate from '@/js/components/PreviewMapTemplate/PreviewMapTemplate'
+
+const spatialExtentUiSchema = {
```
cgokey, line 6: I wonder if there should be a "common" dir, as I suspect this code is largely the same as what is used in collections?

mandyparson: It is, but I don't think it would be wise to put this in a common directory as it would make things more confusing. In my mind, all the files needed to create visualization forms should live under uiSchemas/visualizations. Not having some live under uiSchemas/common.

cgokey (approved):

static/src/js/schemas/uiSchemas/visualizations/spatialExtent.jsx
eudoroolivares2016: I kind of disagree. I think this is fine for now but, in the future refactoring to reduce code duplication is something we should do. Assuming it is **really** the same code but, if there are small differences maybe there is a way to inherit some of these properties

eudoroolivares2016 (approved):

PR #1385, opened 2025-07-01
MMT-4048: Concept Selection Widget needs to include version when retrieving concept full path.

htranho:
# Overview

### What is the feature?

Widget needs to use version to fetch concept full path. Without version, getting new concept full path in draft results in error because getKmsConceptFullPaths can't find it in published version.

### What is the Solution?

Pass version as parameter for kms call in getKmsConceptFullPaths.

Also fix 1 critical vulnerabilities pbkdf2 

Concept Selection Widget in keyword form.

# Testing

Add a new narrower keyword. Save it. Pick another keyword to edit, choose the new keyword as broader or narrower, hit save, see attachment for error shown. Fixed version (this branch) should not show that error.

### Attachments

Screenshot 2025-07-01 at 5 59 31 PM

# Checklist

### DIFF
--- a/static/src/js/utils/getKmsConceptFullPaths.js
+++ b/static/src/js/utils/getKmsConceptFullPaths.js
@@ -9,7 +9,7 @@ import { getApplicationConfig } from 'sharedUtils/getConfig'
  * @async
  * @function getKmsConceptFullPaths
  * @param {Object} params - The parameters for the function.
- * @param {string} params.conceptId - The UUID of the KMS concept whose full paths are to be fetched.
+ * @param {string} params.uuid - The UUID of the KMS concept whose full paths are to be fetched.
  * @param {string|null} [params.version] - Optional. The version of the concept to fetch. If not provided, the latest version is used.
  * @returns {Promise<string[]>} A promise that resolves to an array of full paths as strings.
  * @throws Will throw an error if the fetch operation or XML parsing fails.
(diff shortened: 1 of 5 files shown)


mandyparson:
Getting a lot of errors for "Encountered two children with the same key,"

mandyparson (approved):

static/src/js/utils/getKmsConceptFullPaths.js
```diff
@@ -8,15 +8,47 @@ import { getApplicationConfig } from 'sharedUtils/getConfig'
...
+ *   } catch (error) {
+ *     console.error('Error fetching concept paths:', error);
+ *   }
+ * }
+ *
+ * // Call the function
+ * fetchConceptPaths();
  */
-const getKmsConceptFullPaths = async (value) => {
+const getKmsConceptFullPaths = async ({ conceptId, version }) => {
```
eudoroolivares2016:
```suggestion
const getKmsConceptFullPaths = async ({ uuid, version }) => {
```
this is definitely a nitpick I see in other methods we use `uuid` for consistency I think we should use it here and pass it like that as the key where this gets called its a bit confusing since we do use conceptId as a unique identifier in unrelated parts of the MMT application for CMR

htranho: Done

PR #1387, opened 2025-07-14
MMT-4061: Use node.id to call delete keyword

htranho:
# Overview

### What is the feature?

User should be able to delete a leaf keyword found per filter from the keyword tree.

### What is the Solution?

Fetch keyword uuid from keyword tree node directly.

Filter with a pattern to find a leaf keyword and delete it per 'Delete' from it's context menu.

# Testing

Filter with a pattern to find a leaf keyword and delete it per 'Delete' from it's context menu. There should be no 404 error and keyword is deleted.

### Attachments

N/A

# Checklist

### DIFF
--- a/static/src/js/pages/KeywordManagerPage/KeywordManagerPage.jsx
+++ b/static/src/js/pages/KeywordManagerPage/KeywordManagerPage.jsx
@@ -112,10 +112,8 @@ const KeywordManagerPage = () => {
       setIsDeleting(true)
       setDeleteError(null)
       try {
-        // Check if nodeToDelete has a data property, if not use id directly
-        const conceptId = nodeToDelete.data ? nodeToDelete.data.id : nodeToDelete.id
         const params = {
-          conceptId,
+          conceptId: nodeToDelete.id,
           version: getVersionName(selectedVersion),
           token: tokenValue
         }


static/src/js/pages/KeywordManagerPage/KeywordManagerPage.jsx
```diff
@@ -112,10 +112,8 @@ const KeywordManagerPage = () => {
       setIsDeleting(true)
       setDeleteError(null)
       try {
-        // Check if nodeToDelete has a data property, if not use id directly
-        const conceptId = nodeToDelete.data ? nodeToDelete.data.id : nodeToDelete.id
         const params = {
-          conceptId,
+          conceptId: nodeToDelete.id,
```
eudoroolivares2016: I'm surprised this wouldn't trigger a test change I think we should have something it was kind of a subtle bug

htranho: Done: test added

mandyparson (approved):

package.json
```diff
@@ -2,7 +2,8 @@
   "name": "mmt",
   "version": "0.0.1",
   "scripts": {
-    "start": "vite",
+    "start": "node bin/start.js",
+    "start:app": "vite",
```
cgokey, line 6: ?

htranho: That is for starting everything in the same command windows.

eudoroolivares2016: @cgokey its the script I sent you guys that I was using so we can have 2 terminals instead of 5

eudoroolivares2016 (approved):

PR #1390, opened 2025-07-23
MMT-4050: BUG: When clicking publish, users may receive an intermittent error claiming previewMetadata returns null

mandyparson:
# Overview

### What is the bug?

User clicks Publish (not sure if it matters clicking publish from PublishPreview or MetadataForm) and should be redirected to /concept/conceptId
Instead, user is redirected to /drafts/concept/conceptId which calls getConceptDraft which is looking for previewMetadata. However, when a user clicks publish, that triggers the draft to be deleted, so users are trying to query for a draft that no longer exists. 

### What is the Solution?

Couple things:
In the old code, when a user clicked 'save and publish' it would trigger two navigations: one to drafts/{concept}/{conceptId} and one to {concept}/{conceptId}. This was happening to circumvent an issue with caching with was causing re-renders. There are a few things that needed to be done to clean this up:

1. Change how we cache data. Before, we were deleting the record from the cache to trigger a re-render, now we are updating the cache with the new data which prevents an unnecessary re-render. 
2. Take out the navigation to the new draft. Once the draft is ingested, the code now routes directly to the new record.

usePublishMutation, MetadataForm, DraftPage

# Testing

### Reproduction steps

- **Environment for testing: SIT
- **Collection to test with: any

1. Since the error was intermittent, it will be hard to reproduce the error on the main branch. Though you can try. On main, create a draft (I suggest making a visualizations draft as it has so few required fields) and clicking 'Save & Publish'. You may get the error 'Cannot preview null on field previewMetadata)'. 
2. On this branch, check that all save options work and that the publish button on DraftPage works for all concepts. Try and do things quickly to ensure that no other issues arise when we have multiple pings to cmr. 

### Attachments

# Checklist

### DIFF
--- a/static.config.json
+++ b/static.config.json
@@ -8,7 +8,7 @@
     "mmtHost": "http://localhost:5173",
     "edscHost": "https://search.sit.earthdata.nasa.gov",
     "kmsHost": "https://cmr.sit.earthdata.nasa.gov/kms",
-    "viewCitations": true,
+    "viewCitations": "true",
     "version": "development",
     "defaultResponseHeaders": {
       "Access-Control-Allow-Origin": "*",
--- a/static/src/js/components/Layout/Layout.jsx
+++ b/static/src/js/components/Layout/Layout.jsx
@@ -207,7 +207,7 @@ const Layout = ({ className, displayNav }) => {
                                 ]
                               },
                               // Remove in MMT-4059
-                              ...(viewCitations ? [{
+                              ...((viewCitations === 'true') ? [{
                                 title: 'Citations',
                                 version: `v${ummCit}`,
                                 children: [
--- a/static/src/js/constants/conceptIdTypes.js
+++ b/static/src/js/constants/conceptIdTypes.js
@@ -3,6 +3,7 @@
  */
 const conceptIdTypes = {
   C: 'Collection',
+  CIT: 'Citation',
   O: 'OrderOption',
   S: 'Service',
   T: 'Tool',
--- a/static/src/js/constants/conceptTypeDraftQueries.js
+++ b/static/src/js/constants/conceptTypeDraftQueries.js
@@ -1,10 +1,12 @@
+import { CITATION_DRAFT } from '../operations/queries/getCitationDraft'
+import { COLLECTION_DRAFT } from '../operations/queries/getCollectionDraft'
 import { SERVICE_DRAFT } from '../operations/queries/getServiceDraft'
 import { TOOL_DRAFT } from '../operations/queries/getToolDraft'
-import { COLLECTION_DRAFT } from '../operations/queries/getCollectionDraft'
 import { VARIABLE_DRAFT } from '../operations/queries/getVariableDraft'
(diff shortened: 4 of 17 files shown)


static/src/js/components/MetadataForm/MetadataForm.jsx
```diff
@@ -116,21 +115,21 @@ const MetadataForm = () => {
       cache.modify({
         fields: {
           // Remove the list of drafts from the cache. This ensures that if the user returns to the list page they will see the correct data.
-          drafts: () => {},
-          draft(existingDraftsRef, { DELETE }) {
```
cgokey: I'm a little confused on why you removed this part.  During a normal save operation, do you not want to clear our the cache for the specific draft, otherwise when it refreshes, won't it become out of date?

static/src/js/components/MetadataForm/MetadataForm.jsx
```diff
@@ -116,21 +115,21 @@ const MetadataForm = () => {
...
-
-            return existingDraftsRef
-          }
-
+          drafts: () => {}
         }
       })
-    }
+    },
+    refetchQueries: [
```
cgokey: I see, after the mutation happens, this tells it to re-fetch rather than manually manipulating the cache.

static/src/js/components/MetadataForm/MetadataForm.jsx
```diff
@@ -243,32 +253,38 @@ const MetadataForm = () => {
...
-        if (type === saveTypes.saveAndPublish) {
-          publishMutation(derivedConceptType, nativeId)
+          window.scroll(0, 0)
+        } else if (type === saveTypes.saveAndPublish) {
+          // All save options are set to ingestDraft and then refetch.
+          // Must first navigate to landing page for refetch of new draft before moving on to publishing record
+          navigate(`/drafts/${draftType}/${savedConceptId}/${currentSection}?revisionId=${savedRevisionId}`, { replace: true })
+
+          // Publishses the draft and navigates to published record
+          publishMutation(
```
cgokey: My first thought was on timing, but given you are in the onComplete, the save already finished, so you should be good to publish and receive the callback.   The one thing I didn't notice is that the callback probably should be called in a onComplete on the publish mutation.
```suggestion
const [publishDraftMutation] = useMutation(PUBLISH_DRAFT, {
    update: (cache) => {
      cache.modify({
        fields: {
          drafts: () => {},
          [queryName]: () => {},
          [getHumanizedNameFromTypeParam(queryName)]: () => {}
        }
      })
    },
    onCompleted: (data) => {
      const { publishDraft: publishedDraftResponse } = data
      setPublishDraft(publishedDraftResponse)
      setLoading(false)
    },
    onError: (publishError) => {
      setError(publishError)
      setLoading(false)
    }
  })

```

static/src/js/components/MetadataForm/MetadataForm.jsx
mandyparson: Correct. This may not be a necessary change at this point but it's the recommended way of updating the cache.

eudoroolivares2016:
We should update so that MMT deps are up to date

PR #1392, opened 2025-07-28
MMT-4050: BUG FIX: When clicking publish, users may receive an intermittent error claiming previewMetadata returns null

mandyparson:
# Overview

### What is the feature?

What is the bug?
User clicks Publish (not sure if it matters clicking publish from PublishPreview or MetadataForm) and should be redirected to /concept/conceptId
Instead, user is redirected to /drafts/concept/conceptId which calls getConceptDraft which is looking for previewMetadata. However, when a user clicks publish, that triggers the draft to be deleted, so users are trying to query for a draft that no longer exists.

### What is the Solution?

Take out the || statement from earlier save option and put an additional navigation into saveAndPublish

List impacted areas.

# Testing

### Reproduction steps

- **Environment for testing:SIT
- **Collection to test with:New

1. Create a new record and publish it using save&publish

### Attachments

# Checklist

### DIFF
--- a/static/src/js/components/MetadataForm/MetadataForm.jsx
+++ b/static/src/js/components/MetadataForm/MetadataForm.jsx
@@ -118,7 +118,8 @@ const MetadataForm = () => {
           // Remove the list of drafts from the cache. This ensures that if the user returns to the list page they will see the correct data.
           drafts: () => {},
           draft(existingDraftsRef, { DELETE }) {
-            const { __ref: draftRef = '' } = existingDraftsRef
+            // Examine need for || {} in MMT-4070
+            const { __ref: draftRef = '' } = existingDraftsRef || {}
 
             // If the ref includes this conceptId, delete it to force a refetch of the data
             if (draftRef.includes(conceptId)) {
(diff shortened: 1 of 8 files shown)


eudoroolivares2016 (approved):

static/src/js/components/MetadataForm/MetadataForm.jsx
```diff
@@ -268,6 +264,11 @@ const MetadataForm = () => {
         }
 
         if (type === saveTypes.saveAndPublish) {
+          // We still need to navigate to the new draft revision on save and publish
+          // even though we will end up navigating to the preview page.  The reason
+          // being is because the publish mutation causes the cache to be cleared and as a
+          // result the draft is refetched.
+          navigate(`/drafts/${draftType}/${savedConceptId}/${currentSection}?revisionId=${savedRevisionId}`, { replace: true })
```
cgokey: I don't see any difference in the logic.   In the previous implementation is just refactored to navigate on both "save" and "save and publish".   The new implementation just duplicates the navigation line to both places (unless I'm missing something).   I personally don't like that it is even navigating to draft page (not sure why it has to do that).   So I don't really see what this PR is gaining us if the fundamental logic is the same.

cgokey: Oh wait there is one difference in the previous implementation, it has `if (currentSection)`, is that what is causing the issue and why you moved the navigate to where publish is without that line?

mandyparson: Before, we were giving saveAndPublish two commands, to navigate and to publish, concurrently. This implementation now does it sequentially.

cgokey (changes requested):
Lets huddle this afternoon, still not seeing how this will fix things and wonder if we should just do it right where it only navigates the publish record.

PR #1413, opened 2025-09-30
MMT-4086: MMT collections permission form should surface errors properly

cgokey:
# Overview

### What is the feature?

Couple issues on this one:
1) The collections permissions form is not surfacing errors from cmr graphql.  \
2) If there are errors in the form, the submit button should be disabled.
3) If you go to an ACL that doesn't exist, it should provide a proper error message (not found or something), right now it gives a destructure error.

### What is the Solution?

For #1, wrote a helper method that will extract any errors it finds from the graphql response and surface them.
For #2, added a check if the form is valid, if not don't enable the submit button.
For #3, If you go to a ACL that doesn't exist, now shows a toast message and redirects you back to the permissions page.

Collection Permissions Page

# Testing

For #1, try adding a start/stop date but don't choose a mask.   The graphql error should properly surface now.
For #2, try not filling in a Name field, the submit button should not enable.
For #2, try hitting permission endpoint with /permissions/ACL-BAD or /permissions/ACL-BAD/edit should surface concept not found.

### Attachments

# Checklist

### DIFF
--- a/static/src/js/components/CollectionAssociationForm/CollectionAssociationForm.jsx
+++ b/static/src/js/components/CollectionAssociationForm/CollectionAssociationForm.jsx
@@ -306,7 +306,8 @@ const CollectionAssociationForm = ({ metadata }) => {
   // Creates an action cell based on the current concept type
   const buildActionsCell = useCallback((cellData, rowData) => {
     let disabled = false
-    let checked = false
+    // Use null to allow uncontrolled checkbox behavior, preserving toggle functionality
+    let checked = null
 
     const { conceptId: collectionConceptId } = rowData
     const { collections = {} } = fetchedDraft


htranho (approved):

static/src/js/components/PermissionForm/PermissionForm.jsx
mandyparson, line 370: Change to const { acl = {} } = data to remove errors when hitting endpoint with invalid concept id. IE. Go to http://localhost:5173/permissions/ACL1200410982-CMR and check the errors. Then check them again after this implementation.

static/src/js/components/PermissionForm/__tests__/PermissionForm.test.jsx
```diff
@@ -115,42 +120,45 @@ const setup = ({
...
+        <NotificationsContext.Provider value={notificationContext}>
+          <MemoryRouter initialEntries={[pageUrl]}>
+            <Routes>
+              <Route path="/permissions">
+                <Route
+                  element={
+                    (
+                      <ErrorBoundary>
+                        <Suspense>
+                          <PermissionForm selectedCollectionsPageSize={1} />
```
mandyparson, line 132: This is set to 1 but defaults to 2000?

static/src/js/pages/PermissionPage/PermissionPage.jsx
```diff
@@ -53,14 +53,14 @@ const PermissionPageHeader = () => {
...
-  const { data } = useSuspenseQuery(GET_COLLECTION_PERMISSION, {
+  const { data = {} } = useSuspenseQuery(GET_COLLECTION_PERMISSION, {
     variables: {
       conceptId
     }
   })
 
-  const { acl } = data
-  const { name } = acl
+  const { acl = {} } = data || {}
```
mandyparson: Can be const { acl } = data || {} since you're defaulting it below already

static/src/js/utils/__tests__/extractErrorsFromGraphQlResponse.test.js
```diff
@@ -0,0 +1,221 @@
...
+  test('should handle network errors', () => {
+    const fullError = {
+      networkError: {
+        name: 'ServerError',
+        response: {},
+        statusCode: 500,
+        result: {
+          errors: [
+            {
+              message: 'Context creation failed: Request failed with status code 400',
```
mandyparson, line 61: 500*

static/src/js/components/PermissionForm/__tests__/PermissionForm.test.jsx
cgokey: Yes, that is what the previous test did, only thing I've changed here is add a NotificationsContext.Provider.   You see an issue with the 1?

static/src/js/utils/__tests__/extractErrorsFromGraphQlResponse.test.js
cgokey: I pull this directly from the stack trace.   I think cmr graphql is returning a 500, but it includes the stack trace which told it was a 400 error, i.e., token, so it should be correct and accurately reflect what we are getting from cmr graphql.

static/src/js/components/PermissionForm/PermissionForm.jsx
cgokey: Now using this, as that is how other places are doing it: `const {
        name,
        catalogItemIdentity,
        groups,
        collections
      } = acl || {}`

mandyparson (approved):

PR #1500, opened 2026-05-22
MMT-4181: Get more variables, citations, services for a collection if needed and collections for a service

william-valencia:
# Overview

### What is the feature?

For a Collection Preview, check if we need to get more variables from CMR. Also adjust the limit for collections, services, and citations.

### What is the Solution?

Added extra calls to get more variables and adjusted some limits.

Metadata preview in MMT

# Testing

### Reproduction steps

- **Environment for testing:**

View the metadata preview in MMT for C1200462006-CMR_ONLY. Make sure you can see all the associations and page through them. Also view the service preview for S1200296179-EDF_DEV06. Make sure you can view the collection associations and page through them.

### Attachments

Screenshot 2026-05-21 at 10 39 07 PM
Screenshot 2026-05-21 at 10 39 54 PM

# Checklist

### DIFF
--- a/static/src/js/components/MetadataPreview/MetadataPreview.jsx
+++ b/static/src/js/components/MetadataPreview/MetadataPreview.jsx
@@ -100,21 +100,14 @@ const MetadataPreview = ({
   let { [conceptKey]: concept } = data || {}
 
   // The following variables are needed when we need to fetch extra variables from a collection
-  const { associationDetails = {} } = concept || {}
-  const { variables = [] } = associationDetails || {}
-  const variableCount = variables.length
-  const allVariableConceptIds = variables.map((variableTmp) => variableTmp.conceptId)
-  const excludeVariableConceptIds = variableItems.map((variableTmp) => variableTmp.conceptId)
-  const variableConceptIds = allVariableConceptIds
-    .filter((variableTmp) => !excludeVariableConceptIds.includes(variableTmp))
   const variables = concept?.associationDetails?.variables ?? []
-const variableCount = variables.length
-const excludeSet = new Set(variableItems.map((v) => v.conceptId))
+  const variableCount = variables.length
+  const excludeSet = new Set(variableItems.map((v) => v.conceptId))
 
-const variableConceptIds = variables
-  .map((v) => v.conceptId)
-  .filter((id) => !excludeSet.has(id))
-  .slice(0, conceptsResultLimitInt)
+  const variableConceptIds = variables
+    .map((v) => v.conceptId)
+    .filter((id) => !excludeSet.has(id))
+    .slice(0, conceptsResultLimitInt)
 
   // Returns a conceptsResultLimitInt-limited set of data until there are no newItems left
   const { loading: varLoading, error: varError } = useQuery(conceptTypeQueries.Variables, {
@@ -146,7 +139,7 @@ const variableConceptIds = variables
   // Display loading banner if we need to load additional variables and it hasn't completed yet
   if (varLoading
     || ((variableCount > conceptsResultLimitInt)
-    && (!varError && variableItems && (variableItems.length < variableCount)))) {
+    && (!varError && (variableItems.length < variableCount)))) {
     return (
       <>
         <div>
--- a/static/src/js/components/MetadataPreview/MetadataPreview.jsx
+++ b/static/src/js/components/MetadataPreview/MetadataPreview.jsx
@@ -45,7 +45,7 @@ const MetadataPreview = ({
 }) => {
   const { cmrHost, conceptsResultLimit } = getApplicationConfig()
 
-  const conceptsResultLimitInt = parseInt(conceptsResultLimit, 10)
+  const conceptsResultLimitInt = parseInt(conceptsResultLimit, 10) || 1000
 
   const { draftType } = useParams()
 
--- a/static/src/js/components/MetadataPreview/MetadataPreview.jsx
+++ b/static/src/js/components/MetadataPreview/MetadataPreview.jsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
 import {
   CitationPreview,
   CollectionPreview,
@@ -8,12 +8,14 @@ import {
   VisualizationPreview
 } from '@edsc/metadata-preview'
 import { useParams } from 'react-router'
-import { useSuspenseQuery } from '@apollo/client'
+import { useQuery } from '@apollo/client'
 import Col from 'react-bootstrap/Col'
 import PropTypes from 'prop-types'
 import Row from 'react-bootstrap/Row'
-
 import getConceptTypeByDraftConceptId from '@/js/utils/getConceptTypeByDraftConceptId'
+import parseError from '@/js/utils/parseError'
+import LoadingBanner from '../LoadingBanner/LoadingBanner'
+import ErrorBanner from '../ErrorBanner/ErrorBanner'
 
 import conceptTypeDraftQueries from '../../constants/conceptTypeDraftQueries'
 import conceptTypeQueries from '../../constants/conceptTypeQueries'
@@ -41,7 +43,9 @@ const MetadataPreview = ({
   conceptId,
   conceptType
 }) => {
-  const { cmrHost } = getApplicationConfig()
+  const { cmrHost, conceptsResultLimit } = getApplicationConfig()
+
+  const conceptsResultLimitInt = parseInt(conceptsResultLimit, 10)
 
   const { draftType } = useParams()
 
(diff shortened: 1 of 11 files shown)


static/src/js/components/MetadataPreview/__tests__/__mocks__/MatadataPreviewMocks.js
```diff
@@ -1658,7 +1663,14 @@ export const mockCollectionWithAssociatedVariables = {
     ]
   },
   associatedDois: null,
-  associationDetails: [],
+  associationDetails: {
+    variables: [{ conceptId: 'V100000_MMT' }]
```
mandyparson: There should be 21 associated variables here to match below

static/src/js/components/MetadataPreview/__tests__/__mocks__/MatadataPreviewMocks.js
```diff
@@ -3193,107 +3205,233 @@ export const mockCollectionWithAssociatedVariables = {
       {
         conceptId: 'V100000_MMT',
         name: 'Variable_00',
-        __typename: 'Variable'
+        type: 'SCIENCE_VARIABLE',
```
mandyparson: While this is true for getVariables, the only fields we're asking for on getCollection are conceptId, name, type, and typename

static/src/js/components/MetadataPreview/__tests__/__mocks__/MatadataPreviewMocks.js
```diff
@@ -3642,6 +3780,247 @@ export const mockVariableDraft = {
 
 }
 
+export const mockVariableList = {
```
mandyparson: can use this list above on line 3201 `variables: mockVariableList`

static/src/js/components/MetadataPreview/__tests__/MetadataPreview.test.jsx
```diff
@@ -163,7 +176,7 @@ describe('MetadataPreview', () => {
         }, {})
       })
 
-      expect(ToolPreview).toHaveBeenCalledTimes(1)
+      expect(ToolPreview).toHaveBeenCalledTimes(2)
```
mandyparson: Should be:
expect(ToolPreview).toHaveBeenCalledTimes(2)
expect(ToolPreview).toHaveBeenNthCalledWith(1, '<your assertion>')
expect(ToolPreview).toHaveBeenNthCalledWith(2, '<your assertion>')

That should help clear up why a second call is needed. Lines 132-176 can go in that first assertion and you should be able to remove the waitfor.

Same goes for other tests in this file

static/src/js/components/MetadataPreview/__tests__/MetadataPreview.test.jsx
```diff
@@ -622,14 +653,49 @@ describe('MetadataPreview', () => {
...
             }
           }
+        }, {
+          request: {
+            query: conceptTypeQueries.Variables,
+            variables: {
+              params: {
+                limit: conceptsResultsLimitInt,
+                cursor: null,
+                conceptId: ['V100000_MMT']
```
mandyparson: this should match the list of concept ids in your mock so we can see that this is supposed to be an array

static/src/js/components/MetadataPreview/__tests__/MetadataPreview.test.jsx
```diff
@@ -622,14 +653,49 @@ describe('MetadataPreview', () => {
...
+            }
+          }
+        }, {
+          request: {
+            query: conceptTypeQueries.Variables,
+            variables: {
+              params: {
+                limit: conceptsResultsLimitInt,
+                cursor: 'page-1-cursor',
+                conceptId: ['V100000_MMT']
```
mandyparson: should match the mock

static/src/js/operations/queries/getCollection.js
```diff
@@ -1,7 +1,7 @@
 import { gql } from '@apollo/client'
 
 export const GET_COLLECTION = gql`
-  query GetCollection ($params: CollectionInput, , $variableParams: VariablesInput) {
+  query GetCollection ($params: CollectionInput, $citationParams: CitationsInput, $serviceParams: ServicesInput, $variableParams: VariablesInput) {
```
mandyparson, line 4: @eudoroolivares2016 This wasn't part of the ticket, but why not add visualizations while we're here? I'm assuming at some point we're going to need to page through them now that they're available.

static/src/js/components/MetadataPreview/MetadataPreview.jsx
```diff
@@ -64,31 +68,104 @@ const MetadataPreview = ({
...
+    onCompleted: () => {
+      setVariableItems([])
+      setVariableCursor(null)
+    }
+  })
+
+  let { [conceptKey]: concept } = data || {}
+
+  // The following variables are needed when we need to fetch extra variables from a collection
+  const { associationDetails = {}, variables: metadata = {} } = concept || {}
```
mandyparson: It feels clunky to use both of the fields to achieve something that one of them can do on it's own. I'd vote for eliminating `variables: metadata = {}` and just using associationDetails.variables.length for the count instead.

static/src/js/components/MetadataPreview/MetadataPreview.jsx
```diff
@@ -64,31 +68,104 @@ const MetadataPreview = ({
...
+  // Returns a conceptsResultLimitInt-limited set of data and resets the variableCursor until there are no newItems left
+  const { loading: varLoading, error: varError } = useQuery(conceptTypeQueries.Variables, {
+    skip: (!variableConceptIds.length
+      || (variableItems && (variableItems.length >= variableCount)))
+      || variableCount <= conceptsResultLimitInt,
+    variables: {
+      params: {
+        limit: conceptsResultLimitInt,
+        cursor: variablesCursor,
+        conceptId: variableConceptIds
```
mandyparson, line 120: This is sending the full list of conceptIds. If you console.log variableConceptIds.length you'll notice it has all of the ids from the get go. We should only be asking for the set of ids that we need with this call. 

Additionally, because we already have all the ids, you shouldn't need a cursor. We can just slice out the ids that we need instead.

static/src/js/operations/queries/getCollection.js
eudoroolivares2016: Yea. I'd rather just not build the feature until they're actually needed, not just anticipated

static/src/js/components/MetadataPreview/__tests__/__mocks__/MatadataPreviewMocks.js
eudoroolivares2016: I'm not totally against doing it this way but, in general it seems like we could use less and just push page size etc to env vars mock them in the tests and then just do assertions for that so we don't have to change these or have the giant mocks

william-valencia: This isnt to important since it is just a mock but I can add it.

static/src/js/components/MetadataPreview/__tests__/__mocks__/MatadataPreviewMocks.js
william-valencia: Yes, but apollo is very verbose and prints a bunch of errors when running the tests if these are missing. So i decided to add them.

static/src/js/components/MetadataPreview/__tests__/__mocks__/MatadataPreviewMocks.js
william-valencia: Yes, But will need to make a 3rd list for the combination. I can do that.

static/src/js/components/MetadataPreview/MetadataPreview.jsx
william-valencia: Good catch, this code was actually from your original AssociatedVariables.jsx from preview. Will make the update.

static/src/js/components/MetadataPreview/MetadataPreview.jsx
william-valencia: I think I would rather use the cursor here, instead of saving state on the list of variable concept ids. Also this is how we are doing it in CMR preview.

static/src/js/components/MetadataPreview/__tests__/MetadataPreview.test.jsx
william-valencia: We would still need the await for. I actually tried figuring out why it is being rendered twice, but I couldn't figure it out. I think it was related to the changing from a suspend query to a regular query, but the CollectionPreview is only called once. I can spend more time looking into this if needed.

mandyparson: Every time we use .toHaveBeenCalledTimes, there should be accompanying .toHaveBeenNthCalledWith for each call. See lines 383-385 in kmsConceptSchemeSelector.test.jsx. Same goest for other tests in this file.

static/src/js/components/MetadataPreview/MetadataPreview.jsx
mandyparson: But you already have all the data you need when you create variableConceptIds on line 100. Every time you query the server you are querying it with the entire list of conceptIds. I'm not sure why associatedDetails.variables is not limited. But for your example collection (C1200462006-CMR_ONLY) if console.log out variablesConceptIds.length, you'll notice it comes back with all 3481. We should only send the array of variableIds we are asking for.

william-valencia: I can look into this

static/src/js/components/MetadataPreview/__tests__/MetadataPreview.test.jsx
```diff
@@ -474,97 +693,186 @@ describe('MetadataPreview', () => {
...
+          title: 'Title of Testing Collection Preview',
+          useConstraints: null,
+          version: 'v.1.0.0',
+          versionDescription: null,
+          versionId: null
+        },
+        token: null
+      }, {})
+
+      expect(CollectionPreview).toHaveBeenNthCalledWith(1, {
```
mandyparson: 2*

static/src/js/components/MetadataPreview/__tests__/MetadataPreview.test.jsx
```diff
@@ -595,18 +906,28 @@ describe('MetadataPreview', () => {
...
+        cmrHost: 'http://example.com',
+        conceptId: 'C1000000-MMT',
+        conceptType: 'collection',
+        conceptUrlTemplate: '/{conceptType}/{conceptId}',
+        isPlugin: true,
+        collection: mockCollection,
+        token: null
+      }, {})
+
+      expect(CollectionPreview).toHaveBeenNthCalledWith(1, {
```
mandyparson: 2*

static/src/js/components/MetadataPreview/__tests__/MetadataPreview.test.jsx
```diff
@@ -622,14 +943,49 @@ describe('MetadataPreview', () => {
...
+            }
+          }
+        }, {
+          request: {
+            query: conceptTypeQueries.Variables,
+            variables: {
+              params: {
+                limit: conceptsResultsLimitInt,
+                cursor: 'page-1-cursor',
+                conceptId: ['V100000_MMT', 'V100001_MMT', 'V100002_MMT', 'V100003_MMT', 'V100004_MMT', 'V100005_MMT', 'V100006_MMT', 'V100007_MMT', 'V100008_MMT', 'V100009_MMT', 'V100010_MMT', 'V100011_MMT', 'V100012_MMT', 'V100013_MMT', 'V100014_MMT', 'V100015_MMT', 'V100016_MMT', 'V100017_MMT', 'V100018_MMT', 'V100019_MMT', 'V100020_MMT']
```
mandyparson: there's only one on this page. When I said 'match the mock' I just mean the concept Id should align with the concept Id you're returning.

william-valencia: It does match, because we are using cursor and the full list of conceptId's here. This will change when we use a new algorithm for getting the extra variables without cursors.

mandyparson (approved):

static/src/js/components/MetadataPreview/MetadataPreview.jsx
```diff
@@ -64,31 +69,127 @@ const MetadataPreview = ({
...
+
+  // The following variables are needed when we need to fetch extra variables from a collection
+  const { associationDetails = {} } = concept || {}
+  const { variables = [] } = associationDetails || {}
+  const variableCount = variables.length
+  const allVariableConceptIds = variables.map((variableTmp) => variableTmp.conceptId)
+  const excludeVariableConceptIds = variableItems.map((variableTmp) => variableTmp.conceptId)
+  const variableConceptIds = allVariableConceptIds
+    .filter((variableTmp) => !excludeVariableConceptIds.includes(variableTmp))
+    .slice(0, conceptsResultLimitInt)
```
eudoroolivares2016:
```suggestion
  const variables = concept?.associationDetails?.variables ?? []
const variableCount = variables.length
const excludeSet = new Set(variableItems.map((v) => v.conceptId))

const variableConceptIds = variables
  .map((v) => v.conceptId)
  .filter((id) => !excludeSet.has(id))
  .slice(0, conceptsResultLimitInt)
```
Set is a bit faster but, at this scale shouldn't really matter then its just readability. Tried it and it passed the tests and looked right in the app

eudoroolivares2016 (approved):

PR #1509, opened 2026-09-02
MMT-4199: Create CRUD api for staged metadata

htranho:
# Overview

### What is the feature?
This PR adds a set of APIs for staging "concepts" (collections and other UMM concept types)
in MMT, backed by S3, plus a **cross-environment "Stage for Production" forwarding flow**
built on top of them. It adds Lambda handlers — `getStagedConcept`, `createStagedConcept`,
`deleteStagedConcept`, and `stageConceptForProduction` — plus a dedicated
`stagingApiKeyAuthorizer`. The S3-backed routes live under `/staged/{conceptType}` (PUT) and
`/staged/{conceptType}/{recordId}` (GET, DELETE); the forwarding route is
`POST /providers/{providerId}/{conceptType}/stage-for-production`.

`createStagedConcept` generates an opaque `recordId` (UUID) and stores the concept under it;
the get/delete routes address concepts by that `recordId`. (It's named `createStagedConcept`,
not `createOrUpdateStagedConcept` — every `PUT` generates a brand-new `recordId`, so it never
actually updates an existing record.)

### What is the Solution?

**Concept storage (S3)**
- New CDK-managed `s3.Bucket` `StagingConceptsBucket` (`mmt-<stage>-staging-concepts`): 30-day
  object-expiration lifecycle rule (`expire-staged-concepts`), `RemovalPolicy: RETAIN`, public
  access blocked, SSE-S3. Staged concepts are transient promotion artifacts.
- S3 key is `{conceptType}/{recordId}`.

**Handlers**
- **`createStagedConcept`** (`PUT /staged/{conceptType}`) — generates a `recordId` (UUID),
  writes the request body as-is to S3 at `{conceptType}/{recordId}` via `PutObjectCommand`,
  and returns `{ recordId }` (the caller already knows `conceptType` — it's in the request
  path). This is the only **machine-to-machine** route.
- **`getStagedConcept`** (`GET /staged/{conceptType}/{recordId}`) — retrieves one concept via
  `GetObjectCommand`, returned as `{ concept, conceptType, recordId }`.
- **`deleteStagedConcept`** (`DELETE /staged/{conceptType}/{recordId}`) — deletes via
  `DeleteObjectCommand`, no `HeadObject` preflight: S3's `DeleteObject` is already a
  successful no-op on a missing key, so deleting a nonexistent concept returns success, not
  a `404`.
- **`stageConceptForProduction`**
  (`POST /providers/{providerId}/{conceptType}/stage-for-production`) — forwarding Lambda,
  environment-agnostic (typical use is UAT → Production, but the target is whatever
  `STAGING_TARGET_*` points at). Validates `conceptType` and the request body, runs the
  per-user `fetchProviders` check (`providerId` comes from the path and is used **only** for
  this check — it is never forwarded), then `PUT`s the body to
  `${STAGING_TARGET_API_HOST}/staged/{conceptType}` with a
  `Staging-Api-Key: ${STAGING_TARGET_API_KEY}` header. Reads the generated `recordId` from
  the target's response and returns only
  `{ stagedConceptLink: ${STAGING_TARGET_MMT_HOST}/{conceptType}/staged/{recordId} }`; `502`
  if the target rejects or is unreachable, `500` if the staging target is not fully
  configured. The target's staging key never reaches the browser.

**Authentication & authorization**
- New API Gateway REQUEST authorizer **`stagingApiKeyAuthorizer`** — authenticates solely on
  the `Staging-Api-Key` header, compared against `process.env.STAGING_API_KEY`. Fails closed
  when the env var is unset; bypasses on `IS_OFFLINE`.
- `cdk/mmt/lib/mmt-authorizers.ts` refactored to a shared `makeRequestAuthorizer` helper;
  `stagingApiKeyAuthorizer` added alongside `edlAuthorizer` (EDL authorizer logical IDs
  unchanged). `STAGING_API_KEY` is injected only into the authorizer's Lambda.
- **Auth model:**
  - `createStagedConcept` (PUT) → `stagingApiKeyAuthorizer` only. The handler does no auth
    of its own. It is unauthenticated when run via `bin/api.mjs` locally (the local runner
    invokes no authorizers), the same as every other local route.
  - `getStagedConcept` / `deleteStagedConcept` → `edlAuthorizer` only. EDL-authenticated
    browser routes, but **not provider-scoped** — a staged concept is an opaque `recordId`
    with no provider/native identity, so there is nothing to authorize per-provider. Any
    authenticated MMT user may read/delete any staged concept.
  - `stageConceptForProduction` (POST) → `edlAuthorizer` + in-handler `fetchProviders`.
- Both S3 concept handlers still validate `conceptType` against the shared `s3ConceptTypes`
  allowlist.

**Infra / config wiring**
- `cdk/mmt/lib/mmt-shared-api-gateway-resources.ts` — `/staged` → `{conceptType}` →
  `{recordId}` resource tree, with CORS `OPTIONS` for `GET` / `DELETE` on `{recordId}`.
  `/staged/{conceptType}` carries only the machine-to-machine `PUT`, so it gets no `OPTIONS`.
  `/providers/{providerId}/{conceptType}` survives only as the parent of the
  `stage-for-production` action; there is no `{nativeId}` resource.
- `cdk/mmt/lib/mmt-functions.ts` — the concept Lambdas wired to the new resources;
  `stagingApiKeyAuthorizer` threaded through; `stagingTargetConfig` (the three
  `STAGING_TARGET_*` vars) injected only into `stageConceptForProduction`, not the shared
  Lambda environment. `StageConceptForProductionLambda` uses the default Lambda role (no S3).
- `cdk/mmt/lib/mmt-stack.ts` — creates the bucket; adds `Staging-Api-Key` to CORS
  `allowHeaders`; adds `STAGING_TARGET_API_HOST` / `_MMT_HOST` / `_API_KEY` (dev-safe
  defaults; only meaningfully set where a forward target is configured). **Synth guard:** on
  a deployed build (`NODE_ENV=production`), `cdk synth` throws if `STAGING_API_KEY` is
  missing/placeholder, or if `STAGING_TARGET_API_HOST` is set while `STAGING_TARGET_MMT_HOST`
  is missing **or** `STAGING_TARGET_API_KEY` is missing/placeholder — the three
  `STAGING_TARGET_*` vars are optional only as a group (all unset, or all set together), not
  individually.
- `bin/deploy-bamboo.sh` — passes `bamboo_STAGING_TARGET_API_HOST`,
  `bamboo_STAGING_TARGET_MMT_HOST`, `bamboo_STAGING_TARGET_API_KEY` (alongside
  `STAGING_API_KEY` and the concepts bucket name) through Bamboo → Docker → CDK → Lambda. The
  three `STAGING_TARGET_*` are defaulted to empty (`${bamboo_STAGING_TARGET_*:-}`), so they
  are **optional** Bamboo plan variables — only forwarding environments need to define them.

**Environment variables**

| Variable | Role | Required? |
|---|---|---|
| `STAGING_API_KEY` | inbound secret this environment accepts on the `Staging-Api-Key` header (verified by `stagingApiKeyAuthorizer`) | always |
| `STAGING_CONCEPTS_BUCKET_NAME` | this environment's concepts bucket (defaults to `mmt-${STAGE_NAME}-staging-concepts`) | always |
| `STAGING_TARGET_API_HOST` | outbound — API Gateway base URL `stageConceptForProduction` forwards to | optional — the on/off switch for forwarding |
| `STAGING_TARGET_MMT_HOST` | UI host used to build the `stagedConceptLink` deep link in the response | only if `STAGING_TARGET_API_HOST` is set |
| `STAGING_TARGET_API_KEY` | outbound secret sent to the target; must equal the target's `STAGING_API_KEY` | only if `STAGING_TARGET_API_HOST` is set |

See `docs/stage-for-production-env-vars.md` for the full per-environment breakdown.

**Tests**
- Vitest coverage for `getStagedConcept`, `createStagedConcept`, `deleteStagedConcept`,
  `stageConceptForProduction`, and `stagingApiKeyAuthorizer`: success paths,
  missing/invalid `conceptType`, missing/invalid/unset `Staging-Api-Key` (authorizer,
  fail-closed) and header-casing, S3-layer failures, the target-not-fully-configured cases
  for each of the three `STAGING_TARGET_*` vars individually, and the target-rejection /
  provider-unauthorized paths for the forwarder.
- Full run: `vitest run serverless/` → **30 files, 134 tests pass**. `cdk synth`
  (`STAGE_NAME=dev`) clean with template assertions verified (bucket lifecycle;
  `PUT /staged/{conceptType}` → `StagingApiKeyAuthorizer`; the staged `GET` / `DELETE`
  `{recordId}` routes and `POST /providers/{providerId}/{conceptType}/stage-for-production`
  → `EdlAuthorizer`; no `{nativeId}` resources; `templates` / `users` untouched). eslint /
  tsc clean.
- Verified end-to-end against a local API + S3: seed via `PUT /staged/{conceptType}`, get,
  delete, and the full stage-for-production loopback (returns a valid `stagedConceptLink`);
  also verified the synth guard throws when `STAGING_TARGET_API_HOST` is set but
  `STAGING_TARGET_MMT_HOST` is missing.

- `serverless/src/getStagedConcept/`, `createStagedConcept/`, `deleteStagedConcept/`,
  `stageConceptForProduction/`, `stagingApiKeyAuthorizer/` (all new)
- `serverless/src/utils/getConceptsBucketName.js` (new; `fetchProviders` is reused, unchanged)
- `sharedConstants/s3ConceptTypes.js` (new)
- `setup/startS3.js` — also creates the local concepts bucket
- CDK stack — `mmt-stack.ts`, `mmt-authorizers.ts`, `mmt-functions.ts`,
  `mmt-shared-api-gateway-resources.ts`
- `bin/deploy-bamboo.sh`
- `docs/stage-for-production-env-vars.md`, `docs/stage-for-production-changes.md`,
  `docs/stage-for-production-plan.md`

### Deploy prerequisites (not code)
- Bamboo plans that forward define `bamboo_STAGING_TARGET_API_HOST`,
  `bamboo_STAGING_TARGET_MMT_HOST`, `bamboo_STAGING_TARGET_API_KEY` (secret) — all three
  together, or none; non-forwarding plans leave them undefined (the script defaults them to
  empty). The forwarding environment's `STAGING_TARGET_API_KEY` must byte-for-byte equal the
  target environment's `STAGING_API_KEY` (same shared secret, two names).
- The forwarding Lambda is in-VPC and must be able to reach the target's private API Gateway
  (likely needs PrivateLink / VPC peering / a regional endpoint where the target is a
  different account).
- If `mmt-<stage>-staging-concepts` was pre-created manually, `cdk import` or delete it before
  the first deploy.

# Testing

`npm run start:fast` boots Vite, the local API (`http://localhost:4001` via `bin/api.mjs`,
which pre-builds + synths the CDK template), and local S3 (s3rver) concurrently. Offline, the
EDL authorizer is bypassed and `fetchProviders` accepts the test-mode token
`Authorization: Bearer ABC-1` (grants `MMT_1` / `MMT_2`); `bin/api.mjs` invokes no
authorizers, so the machine-to-machine PUT needs no `Staging-Api-Key` header locally.

- **Seed a concept:**
  `curl -X PUT localhost:4001/dev/staged/collections -H 'Content-Type: application/json' -d '{"ShortName":"Test","Version":"1"}'`
  → `{ "recordId": "<uuid>" }`
- **Get / delete:** `GET` and `DELETE localhost:4001/dev/staged/collections/<uuid>` with
  `Authorization: Bearer ABC-1`.
- **Stage-for-production loopback:** set `STAGING_TARGET_API_HOST=http://localhost:4001/dev`,
  `STAGING_TARGET_MMT_HOST=http://localhost:5173`,
  `STAGING_TARGET_API_KEY=local-staging-api-key` in the API process's environment, then
  `curl -X POST localhost:4001/dev/providers/MMT_1/collections/stage-for-production -H 'Authorization: Bearer ABC-1' -H 'Content-Type: application/json' -d '{"ShortName":"Test","Version":"1"}'`
  → `{ stagedConceptLink }` (e.g. `http://localhost:5173/collections/staged/<recordId>`).

### Attachments

N/A

# Checklist

## Summary by CodeRabbit

- **New Features**
  - Added staging concept management for creating, listing, viewing, and deleting concepts by record ID.
  - Added staging API-key authentication for protected operations.
  - Added support for promoting staged concepts to production with production links.
  - Added encrypted staging storage with automatic expiration.

- **Developer Experience**
  - Added local environment setup, S3 support, and testing scripts for staging workflows.

- **Documentation**
  - Added configuration guidance for staging and promotion environment settings.

- **Tests**
  - Expanded coverage for staging concepts, authentication, storage, and production promotion.


bin/deploy-bamboo.sh
```diff
@@ -76,6 +76,7 @@ dockerRun() {
         -e "AWS_SECRET_ACCESS_KEY=$bamboo_AWS_SECRET_ACCESS_KEY" \
         -e "AWS_SESSION_TOKEN=$bamboo_AWS_SESSION_TOKEN" \
         -e "COLLECTION_TEMPLATES_BUCKET_NAME=${bamboo_COLLECTION_TEMPLATES_BUCKET_NAME}" \
+        -e "STAGING_CONCEPTS_BUCKET_NAME=${bamboo_STAGING_CONCEPTS_BUCKET_NAME}" \
```
mandyparson, line 82: Need to add all of these to bamboo

serverless/src/stageConceptForProduction/handler.js
```diff
@@ -0,0 +1,122 @@
...
+
+  if (!productionApiHost || !productionStagingApiKey) {
+    console.error('Production promotion is not configured for this environment')
+
+    return {
+      statusCode: 500,
+      headers: defaultResponseHeaders
+    }
+  }
+
```
cgokey, line 81: We should URL-encode nativeId before forwarding; # can truncate the ID and ../ can change the destination provider after the permission check.

mandyparson:
Something I fear that was lost in translation: [Here](https://eosdis.slack.com/archives/C04HNK6AK55/p1788533219932299) Matthew and I decided that when a user stages a record they should receive a link that takes them straight to it in prod (something like /collections/staged/{unique-id}). The link is the connector between uat and prod. 

As it stands, this ticket still uses providerId as that connector. We want to move away from this rather than get into the weeds of what to do when a user stages a record under a provider that exists in UAT but not in prod. 

To fix this, the ticket will need to include a change to the createOrUpdateConcept so that it generates a unique id. The s3 key will have to change as well. It's currently {providerId}/{conceptId}/{nativeId}.json which prod won't be able to resolve from the uuid alone. Is it possible to key it just to the uuid? And finally, getConcept and deleteConcept should not call fetchProviders at all. There may be some other changes required here as well that I'm not thinking of.

serverless/src/getConcepts/handler.js
```diff
@@ -0,0 +1,89 @@
...
+  try {
+    const allowedProviderIds = await fetchProviders(event)
+
+    if (!allowedProviderIds.includes(providerId)) {
+      return {
+        statusCode: 404,
+        headers: defaultResponseHeaders
+      }
+    }
+
```
cgokey: The helper fetches the first S3 page, so providers with over 1000 staged records get an incomplete list.   Not sure if this is a problem if we clean up the staged records so the list remains small.

serverless/src/getConcepts/handler.js
```diff
@@ -0,0 +1,89 @@
...
+    const objectList = await s3ListObjects(s3Client, prefix, bucketName)
+
+    const body = objectList.map((object) => {
+      const [, , fileName] = object.Key.split('/')
```
cgokey: Splitting on / truncates native IDs like folder/record to folder. Removing the known prefix preserves the full ID and keeps the existing .json removal working.
```suggestion
const fileName = object.Key.slice(prefix.length)```

File: serverless/src/getConcepts/handler.js (hunk 3 above)
Comment (cgokey) [reply serverless/src/getConcepts/handler.js]: s3 should return `IsTruncated` and `NextContinuationToken` to page through the results

File: docs/stage-for-production-env-vars.md [hunk 5]
```diff
@@ -0,0 +1,106 @@
...
+
+## Per environment
+
+### PROD — final destination (receives, never forwards)
+
+| Variable | Value |
+|---|---|
+| `STAGING_API_KEY` | `<prod-secret>` — real, unique, non-placeholder. This is the key UAT uses to push in. |
+| `STAGING_CONCEPTS_BUCKET_NAME` | default (`mmt-prod-staging-concepts`) |
+| `PRODUCTION_API_HOST` | **unset / empty** |
```
cgokey: Please mark these Bamboo variables as required, since deploy-bamboo.sh fails if they’re unset. Alternatively, update the script to default them to empty strings so they remain optional.

cgokey:
Overall looks good, you may also want to tests for native IDs containing /, #, and .., plus an S3 response with IsTruncated: true? The current tests pass but don’t cover these cases.

cgokey (changes requested):

macrouch:
> Something I fear that was lost in translation: [Here](https://eosdis.slack.com/archives/C04HNK6AK55/p1788533219932299) Matthew and I decided that when a user stages a record they should receive a link that takes them straight to it in prod (something like /collections/staged/{unique-id}). The link is the connector between uat an dprod.
> 
> As it stands, this ticket still uses providerId as that connector. We want to move away from this rather than get into the weeds of what to do when a user stages a record under a provider that exists in UAT but not in prod.
> 
> To fix this, the ticket will need to include a change to the createOrUpdateConcept so that it generates a unique id. The s3 key will have to change as well. It's currently {providerId}/{conceptId}/{nativeId}.json which prod won't be able to resolve from the uuid alone. Is it possible to key it just to the uuid? And finally, getConcept and deleteConcept should not call fetchProviders at all. There may be some other changes required here as well that I'm not thinking of.


I mostly agree with this, but I could be persuaded that including providerId is fine. Provider IDs between CMR environments don't alway match, so are you asking the user to provide a new providerId before they stage the collection? Would it be easier to present them with a dropdown of acceptable options once they move the staged concept into a draft? In that case you just need the unique ID in the URL.

New questions from my review:

- I think the lambda names are confusing. In the context of MMT when I see "getConcept" I think that is a CMR concept. The same goes for every lambda name. Something like "getStagedConcept" is more clear what the lambda does.
- The naming of all the environment variables I find confusing, the main goal of this is UAT to PROD, but it needs to be functional on any environment. It does currently work from any environment to any other environment provided you get the values set correctly, but including production in the names of those variables could be confusing.
- You're duplicating code from your authorizer into your `createOrUpdateConcept`, just let the authorizer do its job. Side note, why is your local API not executing authorizers?
- I don't think `getConcepts` is necessary at all. If you provide the user a link directly to the staged concept there is no need to show a list of concepts to anyone. That is easier code and a better UX
- docs/stage-for-production-env-vars.md has some good information, but it also has ai slop that is wrong. 
- I can't figure out what is the purpose of `scripts/localStagingConceptsTesting/`? Every lambda is unit tested, doing curl against them is just testing your local API, which isn't production code. Those files aren't even called in CI
- `safeCompareSecret` seems like such overkill. You can utilize the lambda authorizer to verify the source IP is the know values of MMT UAT and you take any possibility of a user attacking the endpoint away

serverless/src/getConcepts/handler.js
htranho: getConcepts is no longer existing (Now removed, see Mathew's recommendation)

serverless/src/getConcepts/handler.js
htranho: handler no longer exists: getConcepts

htranho:
> Overall looks good, you may also want to tests for native IDs containing /, #, and .., plus an S3 response with IsTruncated: true? The current tests pass but don’t cover these cases.

Now only uuid as record ID, generated by the handler

scripts/localStagingConceptsTesting/getConcept.sh
```diff
@@ -0,0 +1,99 @@
...
+    echo "FAIL: $description (expected $expected_status, got $actual_status)"
+    echo "  Response body:"
+    sed 's/^/    /' /tmp/get_concept_response_body.json
+    FAIL_COUNT=$((FAIL_COUNT + 1))
+  fi
+}
+
+echo "== Seeding a concept to retrieve =="
+SEED_STATUS="$(
+  curl -s -o /tmp/get_concept_seed.json -w '%{http_code}' \
```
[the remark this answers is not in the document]
htranho: script is removed

scripts/localStagingConceptsTesting/deleteConcept.sh
```diff
@@ -0,0 +1,136 @@
...
+    sed 's/^/    /' /tmp/delete_concept_response_body.json
+    FAIL_COUNT=$((FAIL_COUNT + 1))
+  fi
+}
+
+echo "== Testing deleteConcept endpoint at ${BASE_URL}/${STAGE}/staged/... =="
+echo
+
+echo "-- seeding throwaway concept for delete tests --"
+RECORD_ID="$(seed_concept)"
```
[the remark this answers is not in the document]
htranho: Script removed

serverless/src/stageConceptForProduction/handler.js
htranho: now uuid as recordId created by this handler

scripts/localStagingConceptsTesting/postConcepts.sh
```diff
@@ -0,0 +1,141 @@
...
+# + a server-side call back to this same PUT route). Both end up writing to local
+# S3, but use this one when you just want fixture data without involving the
+# forwarding Lambda or the PRODUCTION_* config.
+#
+# Sources local-env.sh (if present) for shared local dev config
+# (STAGE_NAME, API_BASE_URL, STAGING_API_KEY, etc). Override any of these by
+# exporting them yourself before running this script.
+#
+# Usage:
+#   ./postConcepts.sh
```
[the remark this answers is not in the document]
htranho: script is removed

serverless/src/getStagedConcepts/handler.js
```diff
@@ -0,0 +1,105 @@
...
+    const allowedProviderIds = await fetchProviders(event)
+
+    if (!allowedProviderIds.includes(providerId)) {
+      return {
+        statusCode: 404,
+        headers: defaultResponseHeaders
+      }
+    }
+
+    const objectList = await s3ListObjects(s3Client, prefix, bucketName)
```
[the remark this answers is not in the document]
htranho: handler no longer exists

docs/stage-for-production-env-vars.md
htranho: Done

htranho:
> Something I fear that was lost in translation: [Here](https://eosdis.slack.com/archives/C04HNK6AK55/p1788533219932299) Matthew and I decided that when a user stages a record they should receive a link that takes them straight to it in prod (something like /collections/staged/{unique-id}). The link is the connector between uat and prod.
> 
> As it stands, this ticket still uses providerId as that connector. We want to move away from this rather than get into the weeds of what to do when a user stages a record under a provider that exists in UAT but not in prod.
> 
> To fix this, the ticket will need to include a change to the createOrUpdateConcept so that it generates a unique id. The s3 key will have to change as well. It's currently {providerId}/{conceptId}/{nativeId}.json which prod won't be able to resolve from the uuid alone. Is it possible to key it just to the uuid? And finally, getConcept and deleteConcept should not call fetchProviders at all. There may be some other changes required here as well that I'm not thinking of.

Implemented.

htranho:
> > Something I fear that was lost in translation: [Here](https://eosdis.slack.com/archives/C04HNK6AK55/p1788533219932299) Matthew and I decided that when a user stages a record they should receive a link that takes them straight to it in prod (something like /collections/staged/{unique-id}). The link is the connector between uat an dprod.
> > As it stands, this ticket still uses providerId as that connector. We want to move away from this rather than get into the weeds of what to do when a user stages a record under a provider that exists in UAT but not in prod.
> > To fix this, the ticket will need to include a change to the createOrUpdateConcept so that it generates a unique id. The s3 key will have to change as well. It's currently {providerId}/{conceptId}/{nativeId}.json which prod won't be able to resolve from the uuid alone. Is it possible to key it just to the uuid? And finally, getConcept and deleteConcept should not call fetchProviders at all. There may be some other changes required here as well that I'm not thinking of.
> 
> I mostly agree with this, but I could be persuaded that including providerId is fine. Provider IDs between CMR environments don't alway match, so are you asking the user to provide a new providerId before they stage the collection? Would it be easier to present them with a dropdown of acceptable options once they move the staged concept into a draft? In that case you just need the unique ID in the URL.
> 
> New questions from my review:
> 
> * I think the lambda names are confusing. In the context of MMT when I see "getConcept" I think that is a CMR concept. The same goes for every lambda name. Something like "getStagedConcept" is more clear what the lambda does.
> * The naming of all the environment variables I find confusing, the main goal of this is UAT to PROD, but it needs to be functional on any environment. It does currently work from any environment to any other environment provided you get the values set correctly, but including production in the names of those variables could be confusing.
> * You're duplicating code from your authorizer into your `createOrUpdateConcept`, just let the authorizer do its job. Side note, why is your local API not executing authorizers?
> * I don't think `getConcepts` is necessary at all. If you provide the user a link directly to the staged concept there is no need to show a list of concepts to anyone. That is easier code and a better UX
> * docs/stage-for-production-env-vars.md has some good information, but it also has ai slop that is wrong.
> * I can't figure out what is the purpose of `scripts/localStagingConceptsTesting/`? Every lambda is unit tested, doing curl against them is just testing your local API, which isn't production code. Those files aren't even called in CI
> * `safeCompareSecret` seems like such overkill. You can utilize the lambda authorizer to verify the source IP is the know values of MMT UAT and you take any possibility of a user attacking the endpoint away

Implemented your recommendations (bullet list)

serverless/src/createOrUpdateStagedConcept/handler.js
```diff
@@ -0,0 +1,90 @@
...
+
+import { getApplicationConfig } from '../../../sharedUtils/getConfig'
+import { getS3Client } from '../utils/getS3Client'
+import { getConceptsBucketName } from '../utils/getConceptsBucketName'
+import { s3ConceptTypes } from '../../../sharedConstants/s3ConceptTypes'
+
+let s3Client
+
+/**
+ * Create a concept in S3
```
mandyparson, line 12: Not update. So let's rename this so it reflects that. `createOrUpdateStagedConcept` should be `createStagedConcept`

serverless/src/stageConceptForProduction/handler.js
```diff
@@ -0,0 +1,128 @@
...
+      }
+    }
+
+    const { recordId } = await response.json()
+
+    return {
+      statusCode: 200,
+      headers: defaultResponseHeaders,
+      body: JSON.stringify({
+        stagedConceptLink: `${stagingTargetMmtHost}/staged/${conceptType}/${recordId}`
```
mandyparson, line 112: This will be the browser link and it needs to follow the convention we have set in mmt where we do collections/something-else. In this case, it should be `${stagingTargetMmtHost}/${conceptType}/staged/${recordId}`. We should also make this route a constant so you can refer back to it elsewhere if possible.

cdk/mmt/lib/mmt-stack.ts
```diff
@@ -33,13 +39,32 @@ const {
...
+const isMissingOrPlaceholder = (value: string) => !value || value === LOCAL_STAGING_API_KEY_PLACEHOLDER
+
+if (isDeployedEnvironment) {
+  if (isMissingOrPlaceholder(STAGING_API_KEY)) {
+    throw new Error('STAGING_API_KEY must be set to a non-placeholder value for deployed environments')
+  }
+
+  if (STAGING_TARGET_API_HOST && isMissingOrPlaceholder(STAGING_TARGET_API_KEY)) {
+    throw new Error('STAGING_TARGET_API_KEY must be set to a non-placeholder value when STAGING_TARGET_API_HOST is configured')
+  }
```
mandyparson, line 57: Should defend against missing or undefined STAGING_TARGET_MMT_HOST

serverless/src/stageConceptForProduction/handler.js
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+
+  const {
+    STAGING_TARGET_API_HOST: stagingTargetApiHost,
+    STAGING_TARGET_MMT_HOST: stagingTargetMmtHost,
+    STAGING_TARGET_API_KEY: stagingTargetApiKey
+  } = process.env
+
+  if (!stagingTargetApiHost || !stagingTargetApiKey) {
```
mandyparson, line 73: || !stagingTargetMmtHost

docs/stage-for-production-env-vars.md
```diff
@@ -0,0 +1,74 @@
...
+bed and UAT → PROD is the real promotion path.
+
+## The variables
+
+| Variable | Role | Bamboo plan variable | Required? |
+|---|---|---|---|
+| `STAGING_API_KEY` | **Inbound** secret this environment accepts on the `Staging-Api-Key` header | `bamboo_STAGING_API_KEY` (secret) | **required** (deployed) |
+| `STAGING_CONCEPTS_BUCKET_NAME` | This environment's concepts bucket — leave at the `mmt-${STAGE_NAME}-staging-concepts` default | `bamboo_STAGING_CONCEPTS_BUCKET_NAME` | required |
+| `STAGING_TARGET_API_HOST` | **Outbound** — API Gateway base URL the `stageConceptForProduction` Lambda `PUT`s to. Empty ⇒ the handler returns `500` (forwarding disabled) | `bamboo_STAGING_TARGET_API_HOST` | optional |
+| `STAGING_TARGET_MMT_HOST` | UI host used to build the deep link returned to the browser (`stagedConceptLink` in the response) | `bamboo_STAGING_TARGET_MMT_HOST` | optional |
```
mandyparson, line 15: this is not optional

serverless/src/stageConceptForProduction/handler.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }
+
+  const {
+    PRODUCTION_API_HOST: productionApiHost,
+    PRODUCTION_MMT_HOST: productionMmtHost,
+    PRODUCTION_STAGING_API_KEY: productionStagingApiKey
+  } = process.env
+
+  if (!productionApiHost || !productionStagingApiKey) {
```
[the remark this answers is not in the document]
htranho: outdated

serverless/src/stageConceptForProduction/handler.js
htranho: Done

cdk/mmt/lib/mmt-stack.ts
htranho: good catch, done.

serverless/src/stageConceptForProduction/handler.js
htranho: condition added

docs/stage-for-production-env-vars.md
htranho: document updated

serverless/src/createStagedConcept/handler.js
```diff
@@ -0,0 +1,90 @@
...
+
+import { getApplicationConfig } from '../../../sharedUtils/getConfig'
+import { getS3Client } from '../utils/getS3Client'
+import { getConceptsBucketName } from '../utils/getConceptsBucketName'
+import { s3ConceptTypes } from '../../../sharedConstants/s3ConceptTypes'
+
+let s3Client
+
+/**
+ * Create a concept in S3
```
htranho: Done

bin/deploy-bamboo.sh
htranho: Will add before deploying to SIT

mandyparson (approved):

cgokey (approved):

